Issue #14, opened 2016-03-03
offsetLeft/offsetTop's behavior is unexpected in bottom and right hint alignment


bulyonov:
Using `offsetLeft` actually offsets the hint in the opposite direction (pushing it more towards left as it's increased) when the tip is positioned with `topright` because it's added to the css `right` property.

We can either also accept `offsetRight` or `offsetBottom` (which would be less ambiguous and offer more control, or subtract instead of add `offsetLeft` and `offsetTop` in the case of right and bottom orientations.

*/cc @chrisirhc*

bulyonov:
The issue was fixed in #13 . The changes will be published at 0.2.

PR #16, opened 2016-03-03
Remove offsetLeft/offsetTop entirely.

bulyonov:
Removing offsets for hints entirely: we don't need them on the chart. The changes can be made with CSS.

This change will solve #14.

@uber-common/data-visualization, please review.

### DIFF
--- a/src/example/plot/static-hints.js
+++ b/src/example/plot/static-hints.js
@@ -47,7 +47,7 @@ export default class Example extends React.Component {
             {x: 3, y: 15}
           ]}/>
         <Hint value={{x: 1, y: 10}}/>
-        <Hint value={{x: 0.4, y: 14}} orientation="topleft">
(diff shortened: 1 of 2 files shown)


chrisirhc:
Hm, how will these be done using CSS? I don't think you can mimic this behavior through margins because they're absolutely positioned.

bulyonov:
Instead of using `offsetLeft` and `offsetTop` (adds some ambiguity), I switched to using CSS classes. Now the hint gets the class `rv-hint--orientation-[value]`, where `value` is either `topleft`, `topright`, `bottomleft` or `bottomright`.

**Please note, that this pull request contains a breaking fix.** Previously values of orientation were corresponding the *opposite* position. Now this ambiguity is fixed.

chrisirhc:
lgtm . Might need an FAQ or documentation example to show how to set up offsets

PR #17, opened 2016-03-04
Fix custom values for scales.

bulyonov:
Custom values (e.g. color={someColor}) for scales did not work well. The problem was hidden in the linear scale that could not return a function for string values. Now the issue is fixed by using `category` scale instead.

@uber-common/data-visualization , please review.


chrisirhc:
lgtm

Issue #13, opened 2016-03-03
Bars are 0-height when y-values are the same


bulyonov:
When data of the series has same y values, the height of bars are calculated as 0. It also happens when there is only one bar in data.

bulyonov:
Fixed in #19 and merged into master. Will be published in v0.2 of the library.

bulyonov:
Fixed in #19 and merged into master. Will be published in v0.2 of the library.

PR #19, opened 2016-03-08
Add zero to bar chart; align bar items to 0

bulyonov:
This update does following things:
- Changes the way how the series' flags are passed to the parent (`isDomainAdjustmentNeeded` is moved to `getParentConfig` function). 
- Automatically adds zero to the domain if the series requires that (through zeroBaseValue parent config; see #13 for more detail).
- Replaces existing `BarSeries` with `VerticalBarSeries` and `HorizontalBarSeries` in order to simplify the code.

@uber-common/data-visualization please take a look.

### DIFF
--- a/src/lib/plot/xy-plot.js
+++ b/src/lib/plot/xy-plot.js
@@ -183,6 +183,19 @@ export default class XYPlot extends React.Component {
       attrProps[key] = props[key];
     });
 
+    const zeroBaseProps = {};
+    ATTRIBUTES.forEach(attr => {
+      getSeriesChildren(props.children).forEach((child, index) => {
+        if (!child) {
+          return;
+        }
+        const {zeroBaseValue} = child.type.getParentConfig(attr, child.props);
+        if (zeroBaseValue) {
+          zeroBaseProps[`${attr}BaseValue`] = 0;
+        }
+      });
+    });
+
     const adjustBy = new Set();
     const adjustWhat = new Set();
     getSeriesChildren(props.children).forEach((child, index) => {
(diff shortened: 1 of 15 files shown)


baileyid:
FYI. This is an example data that breaks on my side now.
[{x: 0, y: 945994}, {x: 1, y: 134463}]

screen shot 2016-03-08 at 10 59 54 am
screen shot 2016-03-08 at 11 00 09 am

Hopefully this could fix it?

bulyonov:
@baileyid, it should.

Could you clarify more on your problem? Do you have any errors out there? What type of chart do you have?

src/lib/utils/data-utils.js
```diff
@@ -51,15 +51,16 @@ export function isObjectPropertyInUse(arr, propertyName) {
 /**
  * Add zero to the domain.
```
chrisirhc: This needs to be updated

bulyonov: LGTM for the rest of things? :)

src/lib/utils/scales-utils.js
```diff
@@ -198,25 +202,36 @@ function _createScaleObjectForFunction(domain, range, type, distance, attr) {
...
-      attr,
-      type
-    );
-  return _createScaleObjectForFunction(domain, range, type, distance, attr);
+  let domain = initialDomain || _getDomainByAttr(allData, attr, type);
+  if (typeof baseValue !== 'undefined') {
+    domain = addValueToArray(domain, baseValue);
+  }
+
+  return _createScaleObjectForFunction(
```
chrisirhc: Might want to refactor away the need for this function later on

src/lib/plot/xy-plot.js
```diff
@@ -183,6 +183,19 @@ export default class XYPlot extends React.Component {
...
+        if (!child) {
+          return;
+        }
+        const {zeroBaseValue} = child.type.getParentConfig(attr, child.props);
+        if (zeroBaseValue) {
+          zeroBaseProps[`${attr}BaseValue`] = 0;
+        }
+      });
+    });
+
```
chrisirhc: Is this duplicate logic from below?

chrisirhc:
Looks good. This is a breaking change so we'll need to take note of that on the version bump.

PR #20, opened 2016-03-08
Avoid using static getters for displayName.

bulyonov:
`static get displayName` produces breaking interference with livereactload. Currently `displayName` property is attached to the class after the class is created.
**Note: This should be changed in future, when static properties will come out of the Babel's stage-0**

@uber-common/data-visualization , please take a look.


ibgreen:
@bulyonov - Looks perfect, thanks!

PR #24, opened 2016-03-11
Changed default animation duration from 100ms to 1000ms

adhbh:
Fix for https://github.com/uber-common/react-vis/issues/23

The default animation duration was very short (100ms). Therefore, the animated treemap didnt seem to animate. It is good to keep default animation to 1000ms.


chrisirhc:
This is likely too long as a default for most visualizations. Perhaps we should be looking into configuration and change the duration for the example.

adhbh:
I thought 100ms is too short for default animation duration, hence changed it. Have fixed it for example here: https://github.com/uber-common/react-vis/pull/26

PR #26, opened 2016-03-12
Fixed animation duration in example

adhbh:
Fix for https://github.com/uber-common/react-vis/issues/23


bulyonov:
@adhbh , I see several possible issues in this pull request:
- currently you are overriding defaults without providing any alternative for the user;
- you still rely on the numeric value, that is good for large treemaps, but it will be too long for a tiny visualization.

The problem of the issue is *not* in the delay, but in the way how the animation in the treemap component works. Currently the animation is *not* clearly visible: tree map cells jump from one position to another. It's a CSS issue. We are working on the fix.

PR #27, opened 2016-03-15
Fixed broken links in RadialChart documentation

samhogg:
The links within the RadialChart docs were 404ing, fixed to point to the correct doc.


chrisirhc:
Great catch. Thank you!

PR #28, opened 2016-03-16
Adding stacking to the charts

bulyonov:
Stacking was added to the orthogonal chart. It can be enabled using `stackBy` parameter on `XYPlot`: 

```jsx
<XYPlot stackBy="y">
```

Some of the series (area series and column series) have to be updated to support the new behavior.

### DIFF
--- a/src/lib/plot/xy-plot.js
+++ b/src/lib/plot/xy-plot.js
@@ -26,9 +26,7 @@ import {
   getSeriesChildren,
   getSeriesPropsFromChildren} from '../utils/series-utils';
 
-import {
-  getInnerDimensions,
-  getDataFromChildren} from '../utils/chart-utils';
+import {getInnerDimensions} from '../utils/chart-utils';
 
 import {
   CONTINUOUS_COLOR_RANGE,
@@ -230,7 +228,7 @@ class XYPlot extends React.Component {
   }
 
   render() {
-    const {width, height, animation, stackBy} = this.props;
+    const {width, height, animation} = this.props;
     const {scaleMixins, data} = this.state;
 
     if (!data || !data.length || ![].concat(...data).filter(d => d).length) {
--- a/src/lib/plot/xy-plot.js
+++ b/src/lib/plot/xy-plot.js
@@ -22,6 +22,7 @@ import React from 'react';
 import equal from 'deep-equal';
 
 import {
+  getStackedData,
   getSeriesChildren,
   getSeriesPropsFromChildren} from '../utils/series-utils';
 
@@ -61,7 +62,8 @@ class XYPlot extends React.Component {
       onMouseLeave: React.PropTypes.func,
       onMouseMove: React.PropTypes.func,
       onMouseEnter: React.PropTypes.func,
-      animation: AnimationPropType
+      animation: AnimationPropType,
+      stackBy: React.PropTypes.oneOf(ATTRIBUTES)
     };
   }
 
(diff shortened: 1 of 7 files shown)


src/lib/plot/xy-plot.js
```diff
@@ -225,11 +216,45 @@ class XYPlot extends React.Component {
...
-    const {width, height, animation} = this.props;
+  /**
+   * Checks if the plot is empty or not.
+   * Currently checks the data only.
+   * @returns {boolean} True for empty.
+   * @private
+   */
+  _isPlotEmpty() {
+    const {data} = this.state;
+    return !data || !data.length || ![].concat(...data).filter(d => d).length
```
chrisirhc: I don't think the `[].concat(...data)` is needed, unless `data` has a `length` attribute but doesn't have `.filter` which should break everything anyway.

src/lib/utils/series-utils.js
```diff
@@ -65,6 +65,42 @@ function collectSeriesTypesInfo(children) {
...
+    }
+    childData.push(data.map((d, dIndex) => {
+      let prevValue = 0;
+      let prevIndex = childIndex - 1;
+      while (prevIndex >= 0) {
+        if (childData[prevIndex]) {
+          break;
+        }
+        prevIndex--;
+      }
```
chrisirhc: This seems quite confusing, you're looping through the children, then looping to find the previous child that is truthy?

This logic to find the `prevIndex` seems like it's something that could be outside of the `data.map` loop. You're trying to find the previous child's data that is truthy right?

src/lib/utils/series-utils.js
```diff
@@ -65,6 +65,42 @@ function collectSeriesTypesInfo(children) {
   return result;
 }
 
+export function getStackedData(children, attr) {
+  const childData = [];
+  const attr0 = `${attr}0`;
+  children.forEach((child, childIndex) => {
```
chrisirhc: Shouldn't this method do nothing if `attr` is falsey?

src/test/series-utils.js
```diff
@@ -22,7 +22,6 @@ import test from 'tape';
 import 'babel-polyfill';
 
 import React from 'react';
-import equal from 'deep-equal';
 
 import {
```
chrisirhc: I think it's a good time to add a test for the `getStackedData` since it's quite complex looking.

src/lib/utils/series-utils.js
bulyonov: @chrisirhc, it cannot be falsey.

src/lib/plot/xy-plot.js
bulyonov: @chrisirhc , `[].concat(...data)` is needed: it's an array of arrays (e.g. array of data for each series) and I have to know if *any* data is passed to it.

src/test/series-utils.js
bulyonov: @chrisirhc , should be in a separate pull request. Much more changes are pending and blocked by this diff.

src/lib/plot/xy-plot.js
chrisirhc: How about `data.some(d => d.some(d => d))` ?

src/lib/utils/series-utils.js
chrisirhc: line 85 of xy-plot calls this with `stackBy` which could be empty/null .

bulyonov: Yes. Even it is empty the correct result is returned (please see `!attr` check on line 78). However, I moved it in the inner function to avoid misunderstanding.

src/lib/utils/series-utils.js
bulyonov: Done, thank you!

PR #37, opened 2016-03-21
Enable stacking for area and bar charts

bulyonov:
- Add the getAttr0Value to all series: it tries to retrieve the data from [attr]0 property if given property is defined and fall back to baseValue if no property is available.
- Add the proper support of [attr]0 for the area and bar charts (not required for the rest of the series).
- Add the functionality to turn off grouping for bar charts if stacking is enabled.

Tests will follow in the next diff.

@uber-common/data-visualization , please take a look

### DIFF
--- a/src/lib/plot/series/bar-series.js
+++ b/src/lib/plot/series/bar-series.js
@@ -83,33 +83,36 @@ class BarSeries extends AbstractSeries {
   _updateSeries() {
     const container = getDOMNode(this.refs.container);
     const {
-      sameTypeTotal = 1,
-      sameTypeIndex = 0,
-      data} = this.props;
-
-    const {
+      _stackBy,
+      data,
       lineSizeAttr,
       valuePosAttr,
       linePosAttr,
       valueSizeAttr} = this.props;
+
+    let {
+      sameTypeTotal = 1,
+      sameTypeIndex = 0} = this.props;
 
     if (!data || !data.length) {
       return;
     }
 
     const distance = this._getScaleDistance(linePosAttr);
-    const baseValue = this._getScaleBaseValue(valuePosAttr);
     const lineFunctor = this._getAttributeFunctor(linePosAttr);
     const valueFunctor = this._getAttributeFunctor(valuePosAttr);
+    const value0Functor = this._getAttr0Functor(valuePosAttr);
+
+    if (_stackBy === valuePosAttr) {
+      sameTypeTotal = 1;
+      sameTypeIndex = 0;
+    }
 
     const rects = d3.select(container).selectAll('rect')
       .data(data)
       .on('mouseover', this._mouseOver)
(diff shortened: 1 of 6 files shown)


src/lib/utils/series-utils.js
```diff
@@ -65,10 +65,21 @@ function collectSeriesTypesInfo(children) {
...
+ * Each next value of attr is equal to the previous value plus the difference
+ * between attr0 and attr.
+ * @param {Array} children Array of children.
+ * @param {string} attr Attribute.
+ * @returns {Array} New array of children for the series.
+ */
 export function getStackedData(children, attr) {
   const childData = [];
   let prevIndex = -1;
   children.forEach((child, childIndex) => {
```
chrisirhc: This looks like it can be re-written as a `const childData = children.map(…` to improve clarity and readability.

src/lib/utils/series-utils.js
```diff
@@ -80,14 +91,17 @@ export function getStackedData(children, attr) {
     }
     const attr0 = `${attr}0`;
     childData.push(data.map((d, dIndex) => {
-      let prevValue = 0;
-      if (prevIndex >= 0) {
-        const prevD = childData[prevIndex][dIndex];
-        prevValue = prevD[attr0] + prevD[attr];
+      // In case if it's the first series don't try to override any values.
+      if (prevIndex < 0) {
```
chrisirhc: Might want to consider checking this before the line 93 (`childData.push`) so you can push without looping doing the `.map`.

src/lib/plot/series/bar-series.js
```diff
@@ -83,33 +83,36 @@ class BarSeries extends AbstractSeries {
...
 
     const distance = this._getScaleDistance(linePosAttr);
-    const baseValue = this._getScaleBaseValue(valuePosAttr);
     const lineFunctor = this._getAttributeFunctor(linePosAttr);
     const valueFunctor = this._getAttributeFunctor(valuePosAttr);
+    const value0Functor = this._getAttr0Functor(valuePosAttr);
+
+    if (_stackBy === valuePosAttr) {
+      sameTypeTotal = 1;
+      sameTypeIndex = 0;
```
chrisirhc: Hmm.. how come this line is needed? Does this mean that these are reset whenever stacking by the value attribute? Perhaps a comment?

chrisirhc: I'm guessing this means that you can't have multiple stacked column series like: 
image (http://excelhelphq.com/tutorial-clustered-stacked-column-bar-charts-free/)

chrisirhc:
Looks good, only some minor comments for improvement.

src/lib/plot/series/bar-series.js
chrisirhc: Perhaps we can add a comment here to state what this is for?

PR #38, opened 2016-03-21
Add time series as an example

bulyonov:
Adding time chart as an example of the library usage.

### DIFF
--- a/src/example/plot/time-chart.js
+++ b/src/example/plot/time-chart.js
@@ -1,0 +1,62 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


src/example/plot/time-chart.js
```diff
@@ -0,0 +1,62 @@
...
+        xType="time"
+        width={300}
+        height={300}>
+        <HorizontalGridLines />
+        <VerticalGridLines />
+        <XAxis title="X Axis" />
+        <YAxis title="Y Axis" />
+        <LineSeries
+          data={[
+            {x: timestamp + 86400000, y: 3},
```
chrisirhc: 8640000 looks like it could be a constant like DAY_S

chrisirhc:
Look good

PR #44, opened 2016-04-05
Fix 0 height bars

### LINKED ISSUE #40: Bars have 0 height
The bars have the 0 height in following case

```jsx
<XYPlot, {
	  width: 900,
	  height: 250,
	  yDomain: [0, 9367]
}>
  <VerticalBarSeries, {
    data: [{x: 0, y: 9367}, {x: 1, y: 184}]
  } />
  <XAxis />
  <YAxis />
</XYPlot>
```

bulyonov:
0 height bars were caused not by the incorrect calculation of height, but by the wrong calculation of the bar width, when distance NaN instead of a real number. This small change fixes the calculation of distance (e.g. bar width) for the case when the first distance is the shortest.

This pull request fixes #40 .

@uber-common/data-visualization , please review.

### DIFF
--- a/src/lib/utils/scales-utils.js
+++ b/src/lib/utils/scales-utils.js
@@ -293,7 +293,8 @@ function _computeRightDomainAdjustment(values) {
  */
 function _computeScaleDistance(values, domain, index, scaleFn) {
   if (values.length > 1) {
-    return Math.abs(scaleFn(values[index]) - scaleFn(values[index - 1]));
+    const i = index > 0 ? index : 1;
+    return Math.abs(scaleFn(values[i]) - scaleFn(values[i - 1]));
   }
   if (values.length === 1) {
     return Math.abs(scaleFn(domain[1]) - scaleFn(domain[0]));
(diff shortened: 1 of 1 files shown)


src/lib/utils/scales-utils.js
```diff
@@ -293,7 +293,8 @@ function _computeRightDomainAdjustment(values) {
  */
 function _computeScaleDistance(values, domain, index, scaleFn) {
```
chrisirhc: I suggest renaming index to `shortestDistIndex` or `bestDistIndex` since as it reads right now, it looks as if this scale distance is called for all indexes.

src/lib/utils/scales-utils.js
```diff
@@ -293,7 +293,8 @@ function _computeRightDomainAdjustment(values) {
...
   if (values.length > 1) {
-    return Math.abs(scaleFn(values[index]) - scaleFn(values[index - 1]));
+    const i = index > 0 ? index : 1;
```
chrisirhc: Math.max would also do? Otherwise, I think this is clearer if you did `index >= 1`

chrisirhc:
Looks good to me. Minor nitpicks.
Also, we should fix the tests in a separate PR.

PR #48, opened 2016-04-08
Adjust version detection to support react@15

samhogg:
Hey there!

I've been testing `react@15` (which came out today) with my project, seems like everything works fine apart from the version detection in `react-utils.js` which doesn't cater to the fact that `react` is now using major versions.

This PR fixes the above issue - have smoke tested and seems to be running fine with `react@15` and `react@0.14.x`

Thanks!

### DIFF
--- a/src/lib/utils/react-utils.js
+++ b/src/lib/utils/react-utils.js
@@ -20,11 +20,11 @@
 
 import React from 'react';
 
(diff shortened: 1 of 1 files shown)


chrisirhc:
Looks great to me.  :+1: Thank you @samhogg

samhogg:
Thanks!

PR #51, opened 2016-04-13
Make props control coloring of tree map

mcnuttandrew:
Also added example of usage to the documentation.

### DIFF
--- a/src/example/treemap/dynamic-treemap.js
+++ b/src/example/treemap/dynamic-treemap.js
@@ -67,6 +67,7 @@ export default class DynamicTreemapExample extends React.Component {
     return (
       <Treemap
         animation={true}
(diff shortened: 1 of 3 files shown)


bulyonov:
@mcnuttandrew , thank you for your change. I clearly see what you want to achieve, but unfortunately I cannot approve this pull request: we are using a slightly different pattern on passing scale parts in XYPlot and we should follow them at `Treemap` as well.

Both current code and your pull request in `Treemap` are not using colorRange, colorType, colorDomain properties. Please take a look at how `scale-utils` work: they are doing what you potentially need.

chrisirhc:
Closing in favor of #52 . Thanks for looking into this @mcnuttandrew .

PR #52, opened 2016-04-16
Ability to use colorRange and opacityRange for the treemap

bulyonov:
- Supported `color` and `opacity` scale for the treemap.
- Performed some code cleanup: removed useless `_c` function, `defaultProps` were added, etc.

Also solves the problem of #51

### DIFF
--- a/src/lib/treemap/treemap.js
+++ b/src/lib/treemap/treemap.js
@@ -22,28 +22,19 @@ import React from 'react';
 import d3 from 'd3';
 
 import {getCSSAnimation, AnimationPropType} from '../utils/animation-utils';
+import {getAttributeFunctor} from '../utils/scales-utils';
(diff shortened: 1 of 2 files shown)


src/lib/treemap/treemap.js
```diff
@@ -22,29 +22,20 @@ import React from 'react';
...
-    return prefix;
+function getFontColorFromBackground(background) {
+  if (background) {
+    return d3.hsl(background).l > 0.57 ? '#222' : '#fff';
   }
-  return `${prefix}__${className}`;
+  return null;
 }
 
-const DEFAULT_SCALES = {
```
bulyonov: Old scales, not needed anymore

chrisirhc:
:+1:

PR #61, opened 2016-04-20
Fix opacity for radial chart

### LINKED ISSUE #58: Pie chart gives fails to render opacity correctly
When I set opacity to any value for the pie chart I get a console error:

Cannot use data defined opacity for this seriestype. Using fallback value instead.

bulyonov:
This change fixes the opacity for the radial chart: previously the scale was not generated properly.

This pull request fixes #58.


chrisirhc:
👍  Looks good

PR #66, opened 2016-04-27
Fix scale-utils mutating passed domain

### LINKED ISSUE #65: scale-utils mutates the domain array passed to react-vis
image

This causes weird artifacts such as the chart resizing itself on re-render if the same config object is used on each render.

chrisirhc:
This causes weird artifacts such as the chart resizing itself on re-render if the same config object is used on each render.

Fixes #65


bulyonov:
Very good, thank you, @chrisirhc !

PR #69, opened 2016-04-28
Fix bar series distance calculation

### LINKED ISSUE #68: Bar widths incorrectly calculated for bar charts
The bar widths seem to be incorrectly calculated

image

The algorithm shouldn't be producing widths where there would be bars that overlap.

chrisirhc:
The bar series distance calculation should calculate the minimum distance between the points, and use that to create bars that don't overlap.

- Add failing test cases

Fixes #68

### DIFF
--- a/src/lib/utils/scales-utils.js
+++ b/src/lib/utils/scales-utils.js
@@ -83,7 +83,7 @@ const SCALE_FUNCTIONS = {
  * @returns {number} Index of an element where the smallest distance was found.
  * @private
  */
-function _getSmallestDistanceIndex(values, scaleObject) {
+export function _getSmallestDistanceIndex(values, scaleObject) {
   const scaleFn = _getScaleFnFromScaleObject(scaleObject);
   let result = 0;
   if (scaleFn) {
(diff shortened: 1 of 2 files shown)


src/lib/utils/scales-utils.js
```diff
@@ -78,24 +78,26 @@ const SCALE_FUNCTIONS = {
...
  * the index of the element, where the smallest distance was found.
+ * It returns the first occurrence of i where
+ * `scale(value[i]) - scale(value[i - 1])` is minimal
  * @param {Array} values Array of values.
  * @param {Object} scaleObject Scale object.
  * @returns {number} Index of an element where the smallest distance was found.
  * @private
  */
-function _getSmallestDistanceIndex(values, scaleObject) {
+export function _getSmallestDistanceIndex(values, scaleObject) {
```
bulyonov: Private functions cannot be exported.

chrisirhc: I think we should strive for correctness and stability. I've seen private APIs in the wild, they're usually marked with underscore. If users choose to rely on these methods, we make no guarantees, and it's indicated here that it's private API.

bulyonov:
LGTM, @chrisirhc , I'm merging that.

SimpleCookie:
This issue appear to still be relevant in the latest version, including the main example at https://uber.github.io/react-vis/examples/showcases/plots

@chrisirhc @bulyonov

PR #71, opened 2016-05-01
Add 'dataKey' prop to MarkSeries

haydn:
The `dataKey` prop is passed to D3's `data()` function to override the default by-index behavior:

https://github.com/mbostock/d3/wiki/Selections#data

This is particularly useful when animating a scatterplot.

### DIFF
--- a/src/lib/plot/series/mark-series.js
+++ b/src/lib/plot/series/mark-series.js
@@ -28,6 +28,13 @@ import {DEFAULT_SIZE, DEFAULT_OPACITY} from '../../theme';
 
 class MarkSeries extends AbstractSeries {
 
+  static get propTypes() {
+    return {
+      ... AbstractSeries.propTypes,
+      dataKey: React.PropTypes.func
+    };
+  }
+
   constructor(props) {
     super(props);
     this._mouseOver = this._mouseOver.bind(this);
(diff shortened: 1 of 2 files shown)


src/lib/plot/series/mark-series.js
```diff
@@ -64,12 +71,14 @@ class MarkSeries extends AbstractSeries {
...
-    const {data} = this.props;
+    const {
+      data,
+      dataKey} = this.props;
     if (!data) {
       return;
     }
     const circles = d3.select(container).selectAll('circle')
-      .data(data)
+      .data(data, dataKey)
```
haydn: Actually, is it better to just assume that if there's an ID property it should be used as the key?

```js
const keyFunc = (data[0] && data[0].id) ? ((d, i) => (d && d.id) || i) : undefined;
const circles = d3.select(container).selectAll('circle')
  .data(data, keyFunc)
```

chrisirhc:
@bulyonov could you comment on this?
The key is used for animations and this should be useful for LineSeries and other series as well. We also need documentation and think about testing.

Perhaps we should start discussion in an issue first?

Apercu:
Gonna close this since we changed quite drastically our api and we're not using `select`s anymore.

Issue #62, opened 2016-04-21
XYPlot appears to require a consistant number of children


mcnuttandrew:
For both area and line charts it appears that there is a render issue such that react-vis requries a consitant number of elements. For instance, say we were render three line series:

``
<XYPlot>
<series 1>
<series 2>
<series 3>
</XYPlot>
``
and then add grid lines

``
<XYPlot>
<grid>
<grid>
<series 1>
<series 2>
<series 3>
</XYPlot>
``

More likely than not series 3 would not appear

bulyonov:
Fixed in #75

PR #75, opened 2016-05-06
Dont wrap data prop

### LINKED ISSUE #72: Greater than five data series wraps data
When trying to plot more than five line series i find that each additional series greater than five doesn't render. This appears to be because react-vis is wrapping the data so that given six line series, what you would expect would be:

```
<lineseries data=data_1>
<lineseries data=data_2>
<lineseries data=data_3>
<lineseries data=data_4>
<lineseries data=data_5>
<lineseries data=data_6>
```
But instead i find that last one gets the firsts data, that is:

```
<lineseries data=data_1>
<lineseries data=data_2>
<lineseries data=data_3>
<lineseries data=data_4>
<lineseries data=data_5>
<lineseries data=data_1>
```

I think that this is connected to https://github.com/uber/react-vis/issues/62.

bulyonov:
Further investigation showed that the precalculated data was picked up from the wrong index in the data array.

This PR fixes #72 and #62

### DIFF
--- a/src/lib/plot/xy-plot.js
+++ b/src/lib/plot/xy-plot.js
@@ -243,7 +243,13 @@ class XYPlot extends React.Component {
     const children = React.Children.toArray(this.props.children);
     const seriesProps = getSeriesPropsFromChildren(children);
     return children.map((child, index) => {
-      const dataProps = data[index] ? {data: data[index]} : null;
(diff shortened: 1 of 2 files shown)


chrisirhc:
Looks good to me

PR #87, opened 2016-05-19
Eliminate the use of external fonts for the library

### LINKED ISSUE #82: Consider removing custom fonts from the library
Custom fonts are quite heavy, they are loaded from the external place (this doesn't seem right). We should remove this from main.scss and place them into the examples.

bulyonov:
Previous decision of having external custom fonts in the library was questionable. External fonts are removed from the library and applied in examples only. The user may define his own fonts in the body of the document and the fonts will be picked up well.

This PR fixes #82

### DIFF
--- a/src/example/main.scss
+++ b/src/example/main.scss
@@ -1,3 +1,5 @@
+@import url(https://fonts.googleapis.com/css?family=Sintony:400,700);
+
 @import "../main";
 
 body {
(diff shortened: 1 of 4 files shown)


src/example/main.scss
```diff
@@ -1,3 +1,5 @@
+@import url(https://fonts.googleapis.com/css?family=Sintony:400,700);
+
```
chrisirhc: It doesn't seem that the font is used anywhere in the example

bulyonov: It's used on line 8 of this file:
```sass
font-family: Sintony, Helvetica, sans-serif;
```

chrisirhc: Looks good then. 👍

PR #89, opened 2016-05-20
Switch to enzyme for testing react components

bulyonov:
- Switched to enzyme for testing: the library provides much more functionality and helps to write simpler tests.
- Made big cleanup/refactoring in tests, fixed some incomplete props.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,10 +53,9 @@
     "react-addons-shallow-compare": ">=0.14.0",
     "react-addons-test-utils": "^15.0.2",
     "react-dom": ">=0.14.0",
-    "run-browser-babel": "^4.0.1",
+    "run-browser-babel": "^5.2.0",
     "tape": "^4.0.0",
-    "uglify-js": "^2.6.1",
-    "unitest": "^0.10.0"
+    "uglify-js": "^2.6.1"
   },
   "peerDependencies": {
     "react": "0.14.x - 15.x",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -71,9 +72,9 @@
     "build": "npm run build-js && npm run build-css",
     "watch": "npm run build-example && npm run watch-example && (budo dist/example/main.js --live --css dist/example/main.css)",
     "lint": "eslint src",
-    "execute-tests": "run-browser-babel dist/test/index.js --phantom",
+    "execute-tests": "run-browser-babel dist/test/index.js --bx react/addons --bx react/lib/ReactContext --bx react/lib/ExecutionEnvironment --phantom",
```
chrisirhc: What's the `--bx` option for? To exclude from babel?

bulyonov: @chrisirhc , excluding from browserify. Otherwise, enzyme will be failing on compilation. The reason for it is simple: [this](https://github.com/airbnb/enzyme/blob/bd5249c8590874056ca63c5ba908de067c5b5e90/src/react-compat.js#L23) and other requires do not exist for react >=0.14.x, **but** are required during the browserify's compilation

chrisirhc: There's an `if (REACT013)`, at the beginning of the block. Perhaps it's a bug in enzyme?

bulyonov: It's not. You can refer to their docs.

chrisirhc: You mean http://airbnb.io/enzyme/docs/guides/browserify.html . I see,

bulyonov: Yes, thank you. Not sure if this can be fixed.

PR #90, opened 2016-05-20
Performance: only add one resize listener with makeVisFlexible

btford:
In a somewhat un-scientific test on my local machine, I saw a 50% reduction in latency when you have 4-5 flexible components.

I think we should discuss this a bit before merging, but here's the basic idea:

Rather than have each instance add an event listener to the window, we consolidate the subscripting into a global queue. This also means that we only have one debounce-er for the event as well.

The same idea should be applicable to anywhere that a component uses `componentDidMount` and `componentWillUnmount` to add event listeners.

The current implementation introduces a couple extra helpers. I think we can simplify this since `react-vis` only needs this pattern in one place.

**Another note:** this change removes the possibility for setting the debounce for resize on a per-component basis. I personally think that this is silly– reacting to the resize multiple times due to a difference in debounce rate would likely degrade performance.

The current PR makes the debounce fixed at 100ms. I'm not sure what the best way to expose a global debounce setting would be. Are there any other such global settings in react-vis?

### DIFF
--- a/src/lib/make-vis-flexible.js
+++ b/src/lib/make-vis-flexible.js
@@ -25,6 +25,65 @@ import window from 'global/window';
 
 const CONTAINER_REF = 'container';
 
+// As a performance enhancement, we want to only listen once
+const resizeSubscribers = [];
+const DEBOUNCE_DURATION = 100;
+let timeoutId = null;
+
+/**
+ * Calls each subscriber, debounced to the
+ */
+function debounceEmitResize() {
+  window.clearTimeout(timeoutId);
+  timeoutId = window.setTimeout(emitResize, DEBOUNCE_DURATION);
+}
+
+/**
+ * Calls each subscriber once syncronously.
+ */
+function emitResize() {
+  resizeSubscribers.forEach(cb => cb());
+}
+
+/**
+ * Add the given callback to the list of subscribers to be caled when the
+ * window resizes. Returns a function that, when called, removes the given
+ * callback from the list of subscribers. This function is also resposible for
+ * adding and removing the resize listener on `window`.
+ *
+ * @param {Function} cb - Subscriber callback function
+ * @returns {Function} Unsubscribe function
+ */
+function subscribeToDebouncedResize(cb) {
+  resizeSubscribers.push(cb);
+
+  // if we go from zero to one Flexible components instances, add the listener
+  if (resizeSubscribers.length === 1) {
+    window.addEventListener('resize', debounceEmitResize);
+  }
+  return function unsubscribe() {
+    removeSubscriber(resizeSubscribers, cb);
+
+    // if we have no Flexible components, remove the listener
+    if (resizeSubscribers.length === 0) {
+      window.clearTimeout(timeoutId);
+      window.removeEventListener('resize', debounceEmitResize);
+    }
+  };
+}
+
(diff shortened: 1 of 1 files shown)


src/lib/make-vis-flexible.js
```diff
@@ -25,6 +25,48 @@ import window from 'global/window';
...
+    window.clearTimeout(timeoutId);
+    timeoutId = window.setTimeout(emit, 100);
+  }
+
+  function emit() {
+    subscribers.forEach((cb) => cb());
+  }
+}
+
+function remove(list, item) {
```
btford: Is there a helper/pattern already in the codebase for removing from an array?

src/lib/make-vis-flexible.js
```diff
@@ -25,6 +25,45 @@ import window from 'global/window';
...
+      // if we have no Flexible components, remove the listener
+      if (resizeSubscribers.length == 0) {
+        window.clearTimeout(timeoutId);
+        window.removeEventListener('resize', emit)
+      }
+    }
+  }
+}
+
+function remove(list, item) {
```
bulyonov: Could you rename `remove` to something more specific? e.g. `removeListener`?

src/lib/make-vis-flexible.js
```diff
@@ -25,6 +25,44 @@ import window from 'global/window';
...
+  if (resizeSubscribers.length == 1) {
+    window.addEventListener('resize', debounceEmitResize);
+  }
+  return function unsubscribe() {
+    removeSubscriber(resizeSubscribers, cb);
+
+    // if we have no Flexible components, remove the listener
+    if (resizeSubscribers.length == 0) {
+      window.clearTimeout(timeoutId);
+      window.removeEventListener('resize', emit);
```
chrisirhc: This should be `debounceEmitResize`

src/lib/make-vis-flexible.js
```diff
@@ -25,6 +25,44 @@ import window from 'global/window';
 
 const CONTAINER_REF = 'container';
 
+// As a performance enhancement, we want to only listen
+let resizeSubscribers = [];
+let timeoutId = null;
+
+function debounceEmitResize() {
+  window.clearTimeout(timeoutId);
+  timeoutId = window.setTimeout(emitResize, 100);
```
chrisirhc: 100 should be a constant

src/lib/make-vis-flexible.js
```diff
@@ -25,6 +25,44 @@ import window from 'global/window';
 
 const CONTAINER_REF = 'container';
 
+// As a performance enhancement, we want to only listen
+let resizeSubscribers = [];
```
chrisirhc: This is really a `const` since it's never intended to change

bulyonov:
@btford , I highly suggest to add descriptions for the functions that you have. JSDoc is not enforced in this project, but can be a good option.

src/lib/make-vis-flexible.js
btford: 👍

src/lib/make-vis-flexible.js
btford: 👍

btford:
@bulyonov Done.

PR #96, opened 2016-05-25
Add time-utc scale, similar to d3.time.scale.utc

alphastorm:
Tested locally by modifying `src/example/plot/time-chart.js` to use `time-utc`.


chrisirhc:
Looks good to me.

Issue #93, opened 2016-05-23
Add the checks for most frequent user's errors


bulyonov:
There are several most frequent errors made by users that should be checked:
- `null`s or `undefined`s in the data properties;
- mismatching `x` properties for stacked charts;
- negative or zero values for `log` scales.

mcnuttandrew:
I don't know if these are really the users most frequent errors? I'm gonna close this because I don't think it necessary any more

PR #97, opened 2016-05-27
if a mark x or y is NaN the associated path causes and error.

jckr:


chrisirhc:
This seems like it's due to the data having errors. Perhaps we should be feeding these errors back to the user? Hiding the error seems unexpected.

bulyonov:
We should report about the errors in data instead of hiding them. #93 already has a list of things that should be checked. I'd suggest to add it here as well

jckr:
dropping it

PR #98, opened 2016-05-28
Fix stacked chart rendering

chrisirhc:
Previously, stacked charts weren't rendering correctly as the
stacked data wasn't set correctly on series children.

Regression from #75 .

- Add testing for stacked bar data

### DIFF
--- a/src/test/xy-plot.js
+++ b/src/test/xy-plot.js
@@ -1,0 +1,115 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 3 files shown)


src/test/xy-plot.js
```diff
@@ -0,0 +1,115 @@
...
+// THE SOFTWARE.
+
+import test from 'tape';
+import React from 'react';
+import {shallow} from 'enzyme';
+import VerticalBarSeries from '../lib/plot/series/vertical-bar-series';
+import XAxis from '../lib/plot/x-axis';
+import XYPlot from '../lib/plot/xy-plot';
+
+test('Render a stacked bar chart', assert => {
```
bulyonov: @chrisirhc , I clearly see the technical reasons for this test file, but I'm not sure if separating only **these two** tests from the rest of the tests for the `XYPlot` component is a good idea. I'd suggest to move the rest of the related tests from `components.js` to this file.

bulyonov:
LGTM

PR #103, opened 2016-06-09
Added simple legend for plot charts

macrigiuseppe:
The legend is being displayed on top of the chart. the legend item list is generated by reading all series along with their titles.
Title is new attribute for AbstractrSeries that is only used by Legend to diplay the name of serires along with its color.
If no title is provided the legend will display 'Series [index]' as label.

I added an example to the example page.

screen shot 2016-06-09 at 9 59 42 am

### DIFF
--- a/src/lib/plot/legend-pill.js
+++ b/src/lib/plot/legend-pill.js
@@ -1,0 +1,52 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import React, {Component} from 'react';
+
+class LegendPill extends Component {
+
+  static get propTypes() {
+    return {
+      color: React.PropTypes.string
+    };
+  }
+
+  static get defaultProps() {
+    return {
+      color: 'black'
+    };
+  }
+
+  render() {
+    const {props: {color}} = this;
+    return (
+      <span
+        className="rv-xy-plot__legend__pill"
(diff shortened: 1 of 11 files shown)


src/lib/plot/legend-pill.js
```diff
@@ -0,0 +1,52 @@
...
+    return {
+      color: 'black'
+    };
+  }
+
+  render() {
+    const {props: {color}} = this;
+    return (
+      <span
+        className="rv-xy-plot__series--legend--pill"
```
chrisirhc: Following BEM conventions, this should be something like `rv-xy-plot__series__legend-pill` or `rv-xy-plot__series__legend__pill`

src/lib/plot/legend.js
```diff
@@ -0,0 +1,52 @@
...
+            color={color(index)}
+            index={index}
+            key={index}
+            label={seriesLabels[index]}/>
+        );
+      });
+    }
+
+    return (
+      <div className="rv-xy-plot__series--legend">
```
chrisirhc: `rv-xy-plot__legend` might be more appropriate since `legend` isn't a modifier.

src/lib/plot/series/abstract-series.js
```diff
@@ -50,7 +50,8 @@ export default class AbstractSeries extends PureRenderComponent {
       onSeriesMouseOut: React.PropTypes.func,
       onSeriesClick: React.PropTypes.func,
       onNearestX: React.PropTypes.func,
-      animation: AnimationPropType
+      animation: AnimationPropType,
+      title: React.PropTypes.string
```
chrisirhc: How about calling this `label` since they're referred to as `seriesLabels` in other parts of the code? Alternatively, change `seriesLabels` to seriesTitles`, as long as there's consistent terminology.

src/lib/utils/series-utils.js
```diff
@@ -140,3 +139,37 @@ export function getSeriesPropsFromChildren(children) {
...
+
+/**
+ * Returns a list of labels for all series.
+ * A label is defined by passing the prop "title" to a series element
+ * @param {Array} children
+ * @returns {Array} List of strings|null is title is not defined
+ */
+export function getSeriesLabels(children) {
+  const labels = [];
+  children.forEach((child, childIndex) => {
```
chrisirhc: Looks like you can use `children.map`

chrisirhc:
Looks good, took a pass at this.

bulyonov:
Thank you for your PR @macrigiuseppe. It looks great, however it doesn't cover all the use cases.  Here are the question that remain open:

 - Is there any way to detach the legend from the chart?
 - Is it possible to align the legend to the top? Right? Bottom?
 - What will happen if the legend will be too long?
 - Is it possible to visualize continuous colors/opacity/stroke size on the legend?

I'd suggest to have a legend as a separate component that can be wired with the chart instead of having legend as a part of the chart.

macrigiuseppe:
@bulyonov those are all interesting questions. 

I think we can definitely work on functionalities incrementally? For instance the alignment can be easily achieved by using css rules.
* Is it possible to visualize continuous colors/opacity/stroke size on the legend? I need more info on this.

Follow up question on having the legend as an independent component, how would you envision passing   data to the legend? Should the developer duplicate the effort of passing the same information to both chart and legend?

This is a really simply version that it will satisfy requirements and once this is completed we can review all the necessary requirements.

macrigiuseppe:
@chrisirhc, @bulyonov i updated my pull request. Let me know what you think.
Thanks

chrisirhc:
I spoke to @bulyonov about this PR.
Yes, I think we should work on these incrementally. In the meantime, we can release this as an experimental feature under a private API (`_hasLegend` perhaps).

A few things with regards to Legend as an indepedent component:
- If we use Legend as a child component, it'll look like:
```
<XYPlot>
  <Legend />
</XYPlot>
```
- seriesLabels can be passed to every child component in getClonedChildComponents(), that way, the user doesn't need to add those properties themselves.
- need to add a `beforeSvg` static variable on the class, and add child components that have that attribute be added before the svg (refer to `c.type.requiresSVG` in the code)

PR #104, opened 2016-06-10
Fix ordinal scales for d3 4.0

bulyonov:
Ordinal scales don't support `.rangePoints()` method anymore, this functionality was moved to the new `scalePoint()` scale. This fix returns the previous behavior.

We should consider renaming scale type `ordinal` to something else: currently `ordinal` scale type is not an ordinal scale anymore.


chrisirhc:
Which version did this move happen? We might need to increase the version requirements in package.json

PR #106, opened 2016-06-11
fix rendering of bar charts with ordinal scales

### LINKED ISSUE #105: Bar charts aren't rendering correctly when using ordinal scale
image broken from #104 .

chrisirhc:
- Update example to include example of bar chart with ordinal scale

Fixes #105


jckr:
lgtm.

PR #111, opened 2016-07-07
Add interpolation property

terezka:
This way we can have smooth curves! :sparkles: (And I really need that, please accept)


chrisirhc:
@terezka , it looks like you've created this PR to point to your master branch. That's why it includes all your latest commits even though they're not related to the interpolation PR. You can create a branch that only contains the feature you would like to add then change this PR to point to that branch.

With regards to adding the interpolation attribute, I think it sounds good.

terezka:
..Right, will fix asap.

Issue #114, opened 2016-07-21
React-vis crashes if animation is defined for all chart types


mcnuttandrew:
I'm seeing a bug wherein if animation duration is set via a prop like:

`animation: {duration: 500},`

Then I get a crash error from react-vis like

![screen shot 2016-07-21 at 10 27 33 am](https://cloud.githubusercontent.com/assets/6854312/17032372/c533aa7a-4f2d-11e6-84d7-784e867dc99b.png)

This is completely avoidable by not declaring animation. This appears to affect all chart types

bulyonov:
Fixed in #115 and #126

PR #115, opened 2016-07-22
Upgrade all d3 modules to 1.0

bulyonov:
This PR upgrades all d3 modules to 1.0 or later. 

*Note*: currently d3 modules don't support `import` so currently `require` is used.


chrisirhc:
Does this fix #114 ?

chrisirhc:
How come they don't support `import`? Isn't `import` backward compatible? What error are you seeing when you use it?

ibgreen:
@chrisirhc 
> How come they don't support import?

Looks like d3 v4 is not compatible with babel import at the moment.
https://github.com/d3/d3/issues/2733

bulyonov:
Since there's no possibility import defaults in d3 modules, I'm importing all of the functions from the module. We should make a second pass to clean up the imports later.

PR #116, opened 2016-08-10
Remove d3-axis from axes components

bulyonov:
Current PR does following:
- removes the use of d3 axis from `Axis`, `XAxis` and `YAxis` components, renders the axis using React;
- changes the axis API:
  + `labelFormat`, `labelValues`, `ticksTotal` are now removed;
  + `tickSizeInner`, `tickSizeOuter`, `tickPadding` and `tickFormat` are added instead (compatible to d3);
  + `title` and `classSet` are not supported for now;
  + `animation` is not working;
  + `tickTotal` works now (previously the user could not set them).
- Slightly changes the CSS styling.


chrisirhc:
Tests seem to be failing.

src/lib/plot/axis/x-axis.js
```diff
@@ -20,39 +20,67 @@
...
 
-  render() {
+const defaultProps = {
+  orientation: BOTTOM,
+  attr: 'x'
+};
+
+class XAxis extends PureRenderComponent {
+
+  _getDefaultAxisProps() {
```
chrisirhc: It seems like these computation steps should be performed at the very last minute in the render of the `Axis` as they're required computation for each `orientation` to work. That way, users can also use `Axis`, and `XAxis`, `YAxis` are just simple wrappers to apply a few defaults.

chrisirhc:
Looks good, one comment on the computation of default axis props.

PR #119, opened 2016-08-13
refine treemap doc

nauhc:
Specify Treemap width in the example. Solved "Warning: Failed propType: Required prop `width` was not specified in `Treemap`. Check the render method of `App`".


bulyonov:
Thank you, @nauhc

Issue #118, opened 2016-08-11
Cannot read property 'offsetWidth' of undefined


nklijia:
I am using the makeWidthFlexible() funciton to make the plot auto updating the size.
But many times, I am getting this error
"Cannot read property 'offsetWidth' of undefined"

This error is from this line 
https://github.com/uber/react-vis/blob/master/src/lib/make-vis-flexible.js#L115

salincakli:
getDOMNode() was replaced by reactDOM.findDOMNode with React v 0.15

chrisirhc:
Looks like this can happen when the `containerElement` isn't found in the DOM so there's an additional check needed there.

bulyonov:
Resolving this issue, since #120 was merged in. The fix will be on the next release (0.5)

PR #120, opened 2016-08-18
fix make-vis-flexible

nklijia:
remove redundant param from removeSubscriber() function, it breaks the removeSubscriber function.

So when the componentWillUnmount, the subscriber actually not remove the callback from the subscriber.

Fixing issue: https://github.com/uber/react-vis/issues/118


bulyonov:
@nklijia , awesome, thank you!

PR #121, opened 2016-08-19
Allow the user to apply margin as a single value or as a partial object

bulyonov:
- Now you no longer need to pass an object with all sides (e.g. `left`, `right`, `top` and `bottom`): partial objects such as `{left: 1000}` are accepted.
- The defaults are applied on top of the margins.
- The user is allowed to pass a number for all margins instead of an object (will be applied to all sides).


chrisirhc:
Lgtm

PR #124, opened 2016-08-22
Do not recreate the domain each time when the scale is passed

bulyonov:
The parent picks up the domain from all values that are being passed **and** passes this domain for the children. It helps to reduce the excessive computation of the values, eliminates the use of `_allData` array and helps to run the components separately from the parent.

### DIFF
--- a/src/lib/plot/xy-plot.js
+++ b/src/lib/plot/xy-plot.js
@@ -182,9 +182,6 @@ class XYPlot extends React.Component {
 
     // Make sure that the domain is set and pass the domain as well.
     ATTRIBUTES.forEach(attr => {
-      if (!attrProps[`${attr}Type`]) {
-        attrProps[`${attr}Type`] = 'linear';
-      }
       if (!attrProps[`${attr}Domain`]) {
         attrProps[`${attr}Domain`] = getDomainByAttr(
           allData,
--- a/src/lib/plot/xy-plot.js
+++ b/src/lib/plot/xy-plot.js
@@ -20,22 +20,18 @@
 
 import React from 'react';
 import equal from 'deep-equal';
-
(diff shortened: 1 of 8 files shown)


chrisirhc:
How about adding a simple test to test the creation of this domain prop?

src/lib/plot/xy-plot.js
```diff
@@ -157,18 +159,12 @@ class XYPlot extends React.Component {
...
-   * them further.
-   * @param {Object} data Array of all data.
-   * @param {Object} props Props of the component.
-   * @returns {Object} Map of scale-related props.
-   * @private
-   */
-  _getScaleMixins(data, props) {
+  _getUserScaleProps(data, props) {
+    const filteredData = data.filter(d => d);
+    const allData = [].concat(...filteredData);
```
chrisirhc: You can also do `[...filterData]`

src/test/components.js
```diff
@@ -92,13 +92,21 @@ const XYPLOT_SERIES_PROPS = {
 const XYPLOT_XAXIS_PROPS = {
   xRange: [0, 1],
   xDomain: [0, 1],
-  xType: 'linear'
+  xType: 'linear',
+  width: 100,
+  height: 100,
+  top: 0,
+  left: 0
```
chrisirhc: What are top and left for? Are these the margins?

bulyonov: We do not allow the user to work with the margins anymore. If the user wants to align the elements inside the XYPlot, he should use top, left, width and height. Otherwise, those values will be picked up by the margins.

`marginLeft`, `marginRight`, `marginTop`, `marginBottom`, `innerWidth` and `innerHeight` properties are still passed by XYPlot to set the default position

PR #125, opened 2016-08-22
Change the API of grid lines, update the API of axes

bulyonov:
- Updated the API of grid lines: support the same one, that is used in the new Axis (`tickValues` and `tickTotal`);
- Eliminate the use of d3-axis: all grids are now rendered using React.
- Renamed the `orientation` property of the axes to `position` (makes more sense, doesn't overlap with the grid's orientation anymore).
- Performed minor updates.

### DIFF
--- a/src/lib/plot/axis/axis-ticks.js
+++ b/src/lib/plot/axis/axis-ticks.js
@@ -19,26 +19,26 @@
 // THE SOFTWARE.
 
 import React from 'react';
-import {POSITION, getTickValues} from '../../utils/axis-utils';
+import {ORIENTATION, getTickValues} from '../../utils/axis-utils';
 import {getAttributeScale} from '../../utils/scales-utils';
 
-const {LEFT, RIGHT, TOP, BOTTOM} = POSITION;
+const {LEFT, RIGHT, TOP, BOTTOM} = ORIENTATION;
 
 const propTypes = {
   width: React.PropTypes.number.isRequired,
   height: React.PropTypes.number.isRequired,
-  position: React.PropTypes.oneOf([
+  orientation: React.PropTypes.oneOf([
     LEFT, RIGHT, TOP, BOTTOM
   ]).isRequired
 };
 
-function _getTickTextAttributes(position) {
-  const textAnchor = position === LEFT ?
+function _getTickTextAttributes(orientation) {
+  const textAnchor = orientation === LEFT ?
     'end' :
-    (position === RIGHT ? 'start' : 'middle');
-  const dy = position === TOP ?
+    (orientation === RIGHT ? 'start' : 'middle');
+  const dy = orientation === TOP ?
     '0' :
-    (position === BOTTOM ? '0.72em' : '0.32em');
+    (orientation === BOTTOM ? '0.72em' : '0.32em');
   return {
     textAnchor,
     dy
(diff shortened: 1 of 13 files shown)


src/lib/plot/axis/axis-ticks.js
```diff
@@ -51,16 +51,10 @@ function _getTickFormatFn(scale, tickTotal, tickFormat) {
...
-  return !tickValues ?
-    (scale.ticks ? scale.ticks(tickTotal) : scale.domain()) :
-    tickValues;
-}
-
 function AxisTicks(props) {
   const {
     attr,
-    orientation,
+    position,
```
chrisirhc: The d3 API refers to this as `orientation`. Any strong reason to change this to `position`?

Ref:
https://github.com/d3/d3-axis/blob/master/README.md#api-reference

src/lib/plot/vertical-grid-lines.js
```diff
@@ -19,43 +19,30 @@
...
-      marginTop: React.PropTypes.number,
-      marginLeft: React.PropTypes.number,
-      innerWidth: React.PropTypes.number,
-      innerHeight: React.PropTypes.number,
-      ...getScalePropTypesByAttribute('x')
-    };
-  }
+const propTypes = {
+  ...GridLines.propTypes,
+  orientation: React.PropTypes.oneOf([VERTICAL])
```
chrisirhc: Might want to call this `gridOrientation` to disambiguate.

src/lib/plot/axis/axis-ticks.js
bulyonov: @chrisirhc , agree on that, but we have an ambiguity between `vertical`|`horizontal` orientations and `top`|`left`|`right`|`bottom` orientations as well.

I doubt that it makes any sense to reuse the sides for the grid, thus I decided to rename `orientation` to `position`.

One more thing: in 0.4.x the `orientation` of Axis is overriden by the XAxis/YAxis, so there should be no confusion

chrisirhc:
Looked through. Only concerns here are on the API props naming. `orientation` and `position` seem quite confusing especially since we're adding a new meaning to orientation and changing all references of orientation to position.

src/lib/utils/axis-utils.js
```diff
@@ -54,3 +44,16 @@ export function getTicksTotalFromSize(size) {
...
 }
+
+/**
+ * Get the tick values from a given d3 scale.
+ * @param {d3.scale} scale Scale function.
+ * @param {number} tickTotal Total number of ticks
+ * @param {Array} tickValues Array of tick values if they exist.
+ * @returns {Array} Array of tick values.
+ */
+export function getTickValues(scale, tickTotal, tickValues) {
```
mcnuttandrew: can you add some tests for this? Now that it's pulled out as a util it seems like it'd be a lot easier to unit test it

src/styles/plot.scss
```diff
@@ -48,13 +48,10 @@ $rv-xy-plot-tooltip-padding: 7px 10px;
 }
 
 .rv-xy-plot__grid-lines {
-  line {
```
mcnuttandrew: +1 for not using node type selectors

bulyonov:
@chrisirhc , done, thanks!

Issue #123, opened 2016-08-19
`animation` default property not set in plots


alphastorm:
The [animation docs](https://github.com/uber/react-vis/blob/master/docs/animation.md) say that child components will be animated by default.  However, the property is not set in [defaultProps](https://github.com/uber/react-vis/blob/master/src/lib/plot/xy-plot.js#L68).  This appears to affect all chart types.

bulyonov:
Fixed in #126. Please refer to the updated docs when 0.6 will be releaed.

Issue #91, opened 2016-05-23
Assign the series event listeners using react (e.g. not using d3)


bulyonov:
Currently the components' event listeners are assigned via d3 either at `componentDidMount` or at `componentDidUpdate`. The usage of d3 definitely helps: the event handler automatically gets the values associated with the DOM node, but these event listeners are not testable.

Thus, we should move the components' event listeners to the `render` function.

bulyonov:
Resolved in #126: d3-selections are no longer used

Issue #55, opened 2016-04-18
Pie chart gives console error when changing number of segments is animated


mcnuttandrew:
The pie chart yields a bunch of console errors like this

`d3.js:8756 Error: Invalid value for <path> attribute d="M109.35191368826634,412.2430898706445A426.5,426.5 0 0.000006912,1 -225.26700899816143,362.14965805466653L-144.9842766002235,233.08532321974673Z"`

when animation is on and the number of segments is increasing or decreasing. This doesn't appear to affect the functionality of the chart, but is a little worrisome.

chrisirhc:
Marking this for a later release since it doesn't affect functionality.

bulyonov:
Apparently d3 is a source of problems here: it is not able to animate arcs from the initial state without the `d` property.

bulyonov:
Further investigation showed that d3 generates a `d` property for a path, which contains not `x,y` pairs, but a single value only (`0`, which means that both `x` and `y` are zero). Surprisingly this `d` cannot be animated using d3.

bulyonov:
Fixed in #126

PR #126, opened 2016-08-23
Add the new animations

bulyonov:
This PR changes the way how the components are animated:

- removes d3-transition;
- does animation using react;
- adds a possibility for minor tweaks of the animation

These changes are made for `XYPlot`, `RadialChart` and `Treemap` (and all their subcomponents)

This issue addresses #55, #91, #114 and #123


src/lib/plot/axis/axis.js
```diff
@@ -61,7 +70,7 @@ const defaultProps = {
   orientation: BOTTOM
 };
 
-class Axis extends PureRenderComponent {
+class Axis extends React.Component {
```
chrisirhc: How come this needs to be changed from PureRenderComponent?

src/lib/animation.js
```diff
@@ -0,0 +1,61 @@
...
+      interpolator: null
+    };
+    this._renderChildren = this._renderChildren.bind(this);
+  }
+
+  componentWillReceiveProps(props) {
+    const {data} = this.state;
+    const newData = this._extractPropsFromChild(props);
+    this.setState({
+      interpolator: interpolate(data, newData)
```
chrisirhc: Is this state to cache the `interpolator`?

src/lib/plot/axis/axis.js
bulyonov: @chrisirhc You are right. It is still in progress, testing this out.

src/lib/animation.js
bulyonov: Didn't know that, thanks!

src/lib/animation.js
```diff
@@ -0,0 +1,87 @@
...
+        _animation: Math.random()
+      }
+    );
+  }
+
+  render() {
+    const defaultStyle = {i: 0};
+    const style = {i: spring(1)};
+    return (
+      <Motion {...{defaultStyle, style, key: Math.random()}}>
```
chrisirhc: What's `key: Math.random()` for? If it's a required hack, there should be a comment here as well.

bulyonov: The spring is animated only at the time when motion renders. This means that it requires to be reinitialized somehow to make the next animation.

src/lib/plot/series/abstract-series.js
```diff
@@ -104,47 +100,50 @@ export default class AbstractSeries extends PureRenderComponent {
    * Mouse over handler for all series.
    * @param {boolean} useValue Use value handler if true.
    * @param {Object} d Value object
+   * @param {Object} event Event.
    * @private
    */
-  _mouseOverHandler(useValue, d) {
+  _mouseOverHandler(useValue, d, event) {
```
chrisirhc: How is `event` passed to this?

bulyonov: @chrisirhc Event is passed from inside of each of the series. I made the change to make it more transparent.

PR #127, opened 2016-08-25
First version of legends

bulyonov:
This PR adds:

### Stateless color legends (with and without search) for discrete colors

It should probably be wrapped into the bigger legend component later.
Following legends support custom colors/titles and `onItemClick` handlers.

legends

### Stateless continuous color legends
screen shot 2016-08-29 at 7 24 38 pm

### Stateless continuous size legends
screen shot 2016-08-29 at 7 25 12 pm


chrisirhc:
@bulyonov how about an example of how a legend can be used beside a line chart?

src/styles/legends.scss
```diff
@@ -0,0 +1,63 @@
...
+
+.rv-discrete-color-legend-item.clickable {
+  cursor: pointer;
+  &:hover {
+    background: #f9f9f9;
+  }
+}
+
+.rv-search-wrapper {
+  display: flex;
```
chrisirhc: Use of flex might affect adoption

src/lib/legends/discrete-color-legend.js
```diff
@@ -0,0 +1,50 @@
...
+  width: React.PropTypes.number
+};
+
+function DiscreteColorLegend({items, width, height, onItemClick}) {
+  return (
+    <div className="rv-discrete-color-legend" style={{width, height}}>
+      {items.map((item, i) =>
+        <DiscreteColorLegendItem key={i}
+          {...item}
+          onClick={onItemClick ? () => onItemClick(item, i) : null}
```
jckr: can't we handle the default click of an item? we already decide pretty much what happens when a user clicks on an item (the underlying datapoint gets a truthy disabled property, the element gets a disabled class). I'm fine with allowing the user to override that behavior but we shouldn't have them to implement a click handler imo.

src/lib/legends/search-wrapper.js
```diff
@@ -0,0 +1,61 @@
...
+    null;
+  return (
+    <div className="rv-search-wrapper" style={{width, height}}>
+      <form className="rv-search-wrapper__form">
+        <input
+          type="search"
+          placeholder={searchPlaceholder}
+          className="rv-search-wrapper__form__input"
+          value={searchText}
+          onChange={onChange}/>
```
jckr: likewise, can't we manage the onChange? we're always going to filter items as in the example. If the user wants to override this (ie ....startWith instead of ....indexOf() >-1) fine but I don't think we should have them implement it.

src/lib/legends/discrete-color-legend.js
bulyonov: Yes, this definitely should be handled, but only at the situations when the legend is a part of some chart. Currently legends don't provide this functionality, so they should be wired to the chart manually

src/lib/legends/search-wrapper.js
bulyonov: Yes, now the items are searched by default.

PR #129, opened 2016-08-26
Fix typos and link in xy-plot.md

dmitry-zaets:
- Fix link to `Hint` component
- Fix value of style attribute


bulyonov:
Thank you, @dmitry-zaets !

PR #133, opened 2016-08-31
Updates for DiscreteColorLegend: default colors, slight API changes, etc

### LINKED ISSUE #32: Legends for charts with multiple series
I've been pondering this for the last couple days, working out a means of building a legend with interactivity hooks into the existing XY chart series.

Perhaps something like this could work as a starter - wrapping a higher order component around the series components. This component would essentially be an interface rendering a 'default legend' and include event handlers to do things like filtering on selection of a specific series.

```jsx
<XYPlot
  width={300}
  height={300}>
  <HorizontalGridLines />
  <Legend onSelection={this.onSelection} style={style}>
    <LineSeries
      color='red'
      data={[
        {x: 1, y: 10},
        {x: 2, y: 5},
        {x: 3, y: 15}
      ]}/>
    <LineSeries
      color='blue'
      data={[
        {x: 1, y: 10},
        {x: 2, y: 5},
        {x: 3, y: 15}
      ]}/>
  </Legend>
  <XAxis />
  <YAxis />
</XYPlot>
```

An alternative would be just to have a standalone `<Legend/>` component that is fed information about series the chart from the `XYPlot` - potentially less complexity there, I guess.

Would appreciate thoughts (or alternative approaches!) on this, and happy to help with PR to get it landed.

bulyonov:
This PR makes following changes:

- Simplified the way how elements are passed into `DiscreteColorLegend`: now either objects (`title`, `color` and `disabled`) or strings (treated as titles) can be used.
- The default colors of the library are assigned as to the legends in case if no color is passed.
- The complex chart example is updated: the legend is added, the `onNearestX` is very simplified.
- A new `index` property was added into `info` object of `onNearestX(value, info)` callback (easier for the use).

Docs were updated according to these changes.

This pull request fixes #32

### DIFF
--- a/docs/xy-plot.md
+++ b/docs/xy-plot.md
@@ -281,7 +281,10 @@ Exact opacity for all series points in pixels or a series object.
 #### onNearestX (optional)
 Type: `function(value, info)`  
 A callback function which is triggered each time when the mouse pointer gets close to some X value.
-Callback is triggered with two arguments. `value` is the data point, `info` consists of `innerX` (the left position of the value) and `event` (the React event).
+Callback is triggered with two arguments. `value` is the data point, `info` object has following properties:
+- `innerX` is the left position of the value;
+- `index` is the index of the data point in the array of data;
+- `event` is the event object.
 
 #### onValueMouseOver (optional)
 Type: `function(d, info)`  
--- a/src/example/main.scss
+++ b/src/example/main.scss
@@ -89,11 +89,22 @@ article {
 }
 
 .example-with-click-me {
+  position: relative;
   text-align: center;
   &:hover {
     .click-me {
       animation: none;
     }
+  }
+
+  .chart {
+    margin-right: 200px;
+  }
+
+  .legend {
+    position: absolute;
+    text-align: left;
+    right: 0;
   }
 }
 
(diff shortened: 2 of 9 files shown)


src/example/main.scss
```diff
@@ -89,12 +89,23 @@ article {
...
+  position: relative;
   text-align: center;
   &:hover {
     .click-me {
       animation: none;
     }
   }
+
+  .chart {
+    margin-right: 200px;
```
chrisirhc: Need these styles to get it looking right?
We'll need to figure out how to make these components integrate better without additional styling/CSS in the future.

docs/xy-plot.md
```diff
@@ -281,7 +281,10 @@ Exact opacity for all series points in pixels or a series object.
 #### onNearestX (optional)
 Type: `function(value, info)`  
 A callback function which is triggered each time when the mouse pointer gets close to some X value.
-Callback is triggered with two arguments. `value` is the data point, `info` consists of `innerX` (the left position of the value) and `event` (the React event).
+Callback is triggered with two arguments. `value` is the data point, `info` object has following properties:
+- `innerX` is the left position of the value;
+- `index` is the index of the data point in the array of data;
+- `event` is the event object.
```
chrisirhc: Nice

PR #140, opened 2016-09-02
Add react-vis-graph-components to react-vis

episodeyang:
Hey @bulyonov and @chrisirhc,

Here is my work in progress on merging the `graph-components` with the react-vis code base. 

Right now the plan is to 
- [ ] add example components
- [ ] add test using those example components
- [ ] add example components to example page

I will make sure the linter and the test are happy.

Ge


src/lib/graphs/blocks/block-anchor.js
```diff
@@ -0,0 +1,84 @@
...
+const propTypes = {
+  component: any.isRequired,
+  alignHorizontal: oneOf(Object.keys(HORIZONTAL_ALIGNS).map(key => HORIZONTAL_ALIGNS[key])),
+  alignVertical: oneOf(Object.keys(VERTICAL_ALIGNS).map(key => VERTICAL_ALIGNS[key]))
+};
+
+const defaultProps = {
+  alignHorizontal: HORIZONTAL_ALIGNS.RIGHT,
+  alignVertical: VERTICAL_ALIGNS.MIDDLE,
+  margin: 0
```
bulyonov: Could you comment on what is margin here?

src/lib/graphs/blocks/circle-node.js
```diff
@@ -0,0 +1,56 @@
+import React, {PropTypes, cloneElement, Children} from 'react';
+import NODE_TYPES from '../node-types';
+
+const {string, number} = PropTypes;
+const propTypes = {
+  /** name of the node, used for links to look the node up. */
```
bulyonov: `/**` is used for jsdoc, I'd suggest not to use it here.

episodeyang: These are the syntax defined by react-docgen, borrowing the jsdoc's syntax :D

npm-debug.log.295131701
```diff
@@ -0,0 +1,34 @@
+0 info it worked if it ends with ok
```
chrisirhc: npm-debug* files need to be removed.

src/index.js
```diff
@@ -57,3 +57,6 @@ export {default as ContinuousSizeLegend}
   from './lib/legends/continuous-size-legend';
 
 export {default as makeWidthFlexible} from './lib/make-vis-flexible';
+
+/* graph components */
+export * from './lib/graphs';
```
chrisirhc: Let's be explicit here on the exports from the graphs. This makes it clear which exports are from which file. Imagine when this file has more than one `export *` statement, it'd be difficult to find the source of an export without looking through all the `export *` files.

src/lib/graphs/utils/pad-line.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Created on 8/15/16.
+ */
```
chrisirhc: `Created` comments should be cleaned up.

src/lib/graphs/utils/find-columns.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Created on 8/25/16.
+ */
+import splitHeadsFromRest from './split-heads-from-rest';
```
chrisirhc: If this function is only used by this file, I think it makes sense to include it as part of this file. It can be split out later when it becomes heavier or has external uses.

mcnuttandrew:
I suggest we close this PR, as we have demonstrated that it is entirely possible to make graphs using the XYPlot

mcnuttandrew:
Alright it's been a few days! I'm gonna go ahead and close this. If anyone feels different feel free to re-open

PR #145, opened 2016-09-09
Add the support tickAngle property

bulyonov:
- The new `tickLabelAngle` property was added: it's responsible for the angle of the label for the tick mark.
- Fixed the misaligned axis title (#146).


chrisirhc:
lgtm

PR #152, opened 2016-09-22
Allow the user to pass undefined and nulls as scale props

### LINKED ISSUE #149: Do not treat undefined values as passed attributes for domains
When `undefined` is passed to such props as `xDomain`, `yRange` or other scale related props, it is treated as the real value. This behavior breaks the compatibility with the old source code. This should be fixed.

bulyonov:
This change fixes the problem introduced in 0.5, when the empty scale props were treated as existing values (worked well with undefined/null before).

This PR fixes #149


chrisirhc:
Lgtm

Issue #162, opened 2016-11-10
Allow animation={false}


jameskraus:
Currently, the animation property can only be set to an object or true. The [docs specify](https://github.com/uber/react-vis/blob/master/docs/animation.md) that it can be set to `false`, but this throws errors: 

![image](https://cloud.githubusercontent.com/assets/415057/20185250/576e08be-a739-11e6-8344-79b757b0b93c.png)

PR #163, opened 2016-11-10
Adding false to allowed values for `animation` property

jameskraus:
Related issue: https://github.com/uber/react-vis/issues/162


chrisirhc:
Thanks for your contribution, @jameskraus !
Before I merge in your changes, I need you to sign our Uber Contributor License Agreement: http://t.uber.com/cla
Could you please do that and then post a comment here when that's done?

jameskraus:
I can't sign any CLAs. If you wanted to make a similar, yet unique change to the code, then I guess that would work. It just so happens that there is a built in proptype that encapsulates `oneOf([true, false])`, so anyone else could make that switch.

PR #166, opened 2016-11-11
Add literal scale

mcnuttandrew:
Introduce a literal scale so that properties can be set specified precisely in the data

### DIFF
--- a/src/lib/utils/scales-utils.js
+++ b/src/lib/utils/scales-utils.js
@@ -51,6 +51,13 @@ const ORDINAL_SCALE_TYPE = 'ordinal';
 const CATEGORY_SCALE_TYPE = 'category';
 
 /**
+ * Literal scale.
+ * @type {string}
+ * @const
+ */
+const LITERAL_SCALE_TYPE = 'literal';
+
+/**
  * Log scale name.
  * @type {string}
  * @const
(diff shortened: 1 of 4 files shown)


src/lib/utils/scales-utils.js
```diff
@@ -51,6 +51,13 @@ const ORDINAL_SCALE_TYPE = 'ordinal';
 const CATEGORY_SCALE_TYPE = 'category';
 
 /**
+ * Literal scale.
```
chrisirhc: Might be worthwhile explaining how this differs from d3's identity scale

https://github.com/d3/d3-scale#scaleIdentity

src/lib/utils/scales-utils.js
```diff
@@ -615,3 +622,28 @@ export function getMissingScaleProps(props, data, attributes) {
...
+/**
+ * Return a d3 scale that returns the literal value that was given to it
+ * @returns {function} literal scale.
+ */
+export function literalScale() {
+  function scale(d) {
+    return d;
+  }
+
+  scale.domain = function domain() {
```
chrisirhc: Might want to shorten all these to `() => scale;`, or DRY this with one function, since all the functions are performing exactly the same operation.

src/test/scales-utils.js
```diff
@@ -203,3 +204,17 @@ test('scales-utils/getMissingScaleProps', function t(assert) {
...
+test('scale-utils/literalScale', function t(assert) {
+  const s = literalScale();
+
+  assert.equal(s(0.5), 0.5, 'acts as the identity');
+  assert.equal(s(1), 1, 'acts as the identity');
+  assert.equal(s(1.5), 1.5, 'acts as the identity');
+  assert.equal(s(2), 2, 'acts as the identity');
+  assert.equal(s(2.5), 2.5, 'acts as the identity');
+
+  assert.equal(s('2'), '2', 'does NOT coerce input to a number');
```
chrisirhc: 👍 Here lies the value of this scale

src/lib/utils/scales-utils.js
```diff
@@ -615,3 +624,24 @@ export function getMissingScaleProps(props, data, attributes) {
...
+  function response() {
+    return scale;
+  }
+
+  scale.domain = response;
+  scale.range = response;
+  scale.unknown = response;
+  scale.copy = response;
+
+  return scale;
```
bulyonov: Since scale is immutable, I'd suggest to move it outside of the `literalScale` function

mcnuttandrew:
this PR is described in #167

Issue #153, opened 2016-09-27
Treemap should be able to switch layout strategy


mcnuttandrew:
The current tree map only supports one layer of children, which basically makes it a rectangular pie chart. While this is cool, it would be a lot more useful if it could be support a full tree of children. This could also give us the opportunity to embrace other d3 hierarchal layouts such as partition (which can be used to make sunburst diagrams), packing, and trees.

mcnuttandrew:
In a profoundly shocking turn of events, treemap actually already supports multiple levels of children. Without too much difficulty it appears that this chart could also be warped into packing and tree, so i've updated the issue name accordingly

mcnuttandrew:
Scaling back my commentary, i think this feature should be constrained to allowing treemaps to also use the packing layout. The two structures are basically the same, and the changeover would be pretty small.

PR #173, opened 2016-11-17
Expand tree example

mcnuttandrew:
It turns out the the current tree map does support nested children! However this is unfortunately unclear from the examples so I expanded the base one to be based on bostock's default data set. #153

### DIFF
--- a/src/example/main.scss
+++ b/src/example/main.scss
@@ -178,7 +178,3 @@ article {
     position: absolute;
   }
 }
-
-.nested-tree-example {
-  border: thin solid #fff;
-}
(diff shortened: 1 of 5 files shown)


src/styles/treemap.scss
```diff
@@ -4,13 +4,13 @@
 }
 
 .rv-treemap__leaf {
-  position: absolute;
+  border: thin solid #fff;
```
chrisirhc: Does this style change have any implication downstream? I'm concerned that a library downstream now sees a border where there wasn't one before.

src/example/main.scss
```diff
@@ -178,3 +178,7 @@ article {
     position: absolute;
   }
 }
+
+.nested-tree-example {
```
chrisirhc: Isn't it `.nested-tree-example .rv-treemap__leaf` ?

Issue #181, opened 2016-11-19
Export the components that underly the build of Axes, GridLines, Plots, etc.


alee8:
Seems like a great way to build on the existing components for custom purposes.  I certainly have a desire to do so.  For example, custom XYPlot, X/YAxis, Vertical/HorizontalGridLines.

mcnuttandrew:
@alee8 this is done after #183 right?

alee8:
As much as was permitted.  However, I tried to include the files in
CityFlow after pulling the latest and could not import them.
I am unable to work on this until tomorrow the earliest.
Any chance you can try to import these 3 new files and in particular, the
utils functions and see if you see the same issue?

Alison

PR #183, opened 2016-11-22
Expose GridLines, AxisUtils, ScaleUtils

alee8:
This addresses part of #181. 
To reduce exposing API surface, picked only a couple of base component and utils to export.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -39,6 +39,7 @@ export {default as Crosshair} from './lib/plot/crosshair';
 export {default as XAxis} from './lib/plot/axis/x-axis';
 export {default as YAxis} from './lib/plot/axis/y-axis';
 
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -57,3 +58,6 @@ export {default as ContinuousSizeLegend}
   from './lib/legends/continuous-size-legend';
 
 export {default as makeWidthFlexible} from './lib/make-vis-flexible';
+
+export {default as AxisUtils} from './lib/utils/axis-utils';
+export {default as ScalesUtils} from './lib/utils/scales-utils';
```
chrisirhc: For consistency, how about exporting this as `ScaleUtils`? Since `Axis` is singular

PR #185, opened 2016-11-22
Adding ability to specify d3 curve function name as a line series prop

skastel:
In order to make smoothed line series, a user should be able to specify the name of one of the many smoothing functions that d3-shape provides: https://github.com/d3/d3-shape#curves

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,10 @@ npm test
 
 ## Change log
 
+### (UNRELEASED)
+
+- Improvement: added line smoothing via d3-shape curve functions ([#185](https://github.com/uber/react-vis/pull/185)).
+
 ### 0.6.4
 
 - Bugfix: Fixed the issue with numeric titles in legends ([#154](https://github.com/uber/react-vis/pull/154)).
--- a/docs/xy-plot.md
+++ b/docs/xy-plot.md
@@ -364,9 +364,18 @@ The way to align the hint inside the chart. When `auto` is set the hint is tryin
 Set the hint to `topleft` if you want to see a "conventional" hint alignment.
 
 #### curve (optional)
-Type: `string`
+Type: `string|function`
 Default: `null`  
-Apply the named curve function from the D3 shape library to smooth the line series plot, see [the D3 documentation](https://github.com/d3/d3-shape#curves) for function names.
+Apply the provided or named curve function from the D3 shape library to smooth the line series plot, see [the D3 documentation](https://github.com/d3/d3-shape#curves) for function names and instructions. Providing the function, not the name, will require importing the d3-shape package in order to configure it:
+
+```javascript
+// Setting up with only a name
+const stringCurveProp = <LineSeries data={data} curve={'curveMonotoneX'} .../>;
+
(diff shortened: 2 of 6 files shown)


mcnuttandrew:
Hey @skastel 
This is super cool! Can you add an example to the examples page of this in use?

skastel:
@mcnuttandrew Done added some smoothed lines in the examples page including on the main complex chart.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-vis",
-  "version": "0.6.5",
+  "version": "0.6.6",
```
chrisirhc: We try to keep version bumps outside of PRs. That way, we can keep commit history readable/clean, and the version bump/publishing process tied closely together.

docs/xy-plot.md
```diff
@@ -363,6 +363,11 @@ Default: `auto`
 The way to align the hint inside the chart. When `auto` is set the hint is trying to stay inside the bounding box of the chart.
 Set the hint to `topleft` if you want to see a "conventional" hint alignment.
 
+#### curve (optional)
+Type: `string`
+Default: `null`__
```
chrisirhc: trailing `__` ?

README.md
```diff
@@ -83,6 +83,10 @@ npm test
 
 ## Change log
 
+### 0.6.6
```
chrisirhc: Let's add `(unreleased)` here, so we can remove it when we release, and we can append to the change log later.

docs/xy-plot.md
skastel: Oops, fixed.

package.json
skastel: Reverted.

README.md
skastel: Done.

src/lib/plot/series/line-series.js
```diff
@@ -36,15 +36,27 @@ const STROKE_STYLES = {
...
+  strokeStyle: React.PropTypes.oneOf(Object.keys(STROKE_STYLES)),
+  curve: React.PropTypes.string
 };
 
 class LineSeries extends AbstractSeries {
+
+  _renderLine(data, x, y, curve) {
+    let line = d3Shape.line();
+    if (curve !== null && d3Shape[curve]) {
+      line = line.curve(d3Shape[curve]);
```
chrisirhc: Was looking through the curve documentation. Any thoughts on how to specify `var line = d3.line().curve(d3.curveCatmullRom.alpha(0.5));` with the curve prop?

skastel: How about allowing users to import d3-shape and configure and submit their own curve function:
```javascript
if (curve !== null) {
  if (typeof curve === 'string' && d3Shape[curve]) {
    line = line.curve(d3Shape[curve]);
  } else if (typeof curve === 'function') {
    line = line.curve(curve);
  }
}
```

chrisirhc: Sounds good 👍

skastel: Ok, updated to support functions as well as strings, also updated the docs accordingly.

PR #188, opened 2016-11-23
0.6.6

mcnuttandrew:


chrisirhc:
This shouldn't be a PR. Version bumps go directly to master to avoid git tag mess.

PR #190, opened 2016-11-29
More exports181

mcnuttandrew:


alee8:
First effort did not account for the fact that neither axis-utils.js and scales-utils.js had default export. Rectifying by providing an export default for each.

1. All previously marked export const and function in axis-utils are in an object marked as export default.
2. All marked export function except functions with underscore in scales-utils are exported in an object marked as export default.

Issue #180, opened 2016-11-18
Different CSS class names to differentiate XAxis from YAxis so they can be separately styled


alee8:
On an XYPlot, the same class name is used for the XAxis as well as the YAxis. If you want to use CSS to change the styling, you would have to use nth-of-type to have any chance of changing styling on one of the axes.  However, this is brittle and is dependent on which Axis is specified first and second.

mcnuttandrew:
Hey @alee8 I'm gonna close this as the solution is prod. Feel free to reopen if i've missed something

alee8:
Thanks for the heads up and the fix.  Will definitely use it now.

Alison

PR #191, opened 2016-11-29
Add specific class names to x and y axes

mcnuttandrew:
Also allow users to specify custom classnames! Addresses #180

### DIFF
--- a/src/lib/plot/axis/axis.js
+++ b/src/lib/plot/axis/axis.js
@@ -73,8 +73,8 @@ const defaultProps = {
 };
 
 const predefinedClassName = 'rv-xy-plot__axis';
-const VERTICAL_CLASS_NAME = 'rv-xy-plot__axis-vertical';
-const HORIZONTAL_CLASS_NAME = 'rv-xy-plot__axis-horizontal';
+const VERTICAL_CLASS_NAME = 'rv-xy-plot__axis--vertical';
+const HORIZONTAL_CLASS_NAME = 'rv-xy-plot__axis--horizontal';
 
 class Axis extends PureRenderComponent {
 
--- a/src/lib/plot/axis/axis.js
+++ b/src/lib/plot/axis/axis.js
@@ -153,8 +153,9 @@ class Axis extends PureRenderComponent {
       top,
       width
     } = props;
-    const isY = [LEFT, RIGHT].indexOf(orientation) > -1;
-    const axisClassName = isY ? VERTICAL_CLASS_NAME : HORIZONTAL_CLASS_NAME;
+    const isVertical = [LEFT, RIGHT].indexOf(orientation) > -1;
+    const axisClassName = isVertical ? VERTICAL_CLASS_NAME :
+      HORIZONTAL_CLASS_NAME;
     return (
       <g
         transform={`translate(${left},${top})`}
--- a/src/lib/plot/axis/axis.js
+++ b/src/lib/plot/axis/axis.js
@@ -22,11 +22,7 @@ import React from 'react';
 
 import Animation from '../../animation';
 import PureRenderComponent from '../../pure-render-component';
-import {
-  DIRECTION,
-  getTicksTotalFromSize,
-  ORIENTATION
-} from '../../utils/axis-utils';
+import {ORIENTATION, getTicksTotalFromSize} from '../../utils/axis-utils';
 import AxisLine from './axis-line';
 import AxisTicks from './axis-ticks';
 import AxisTitle from './axis-title';
@@ -150,12 +146,12 @@ class Axis extends PureRenderComponent {
 
     const {
       className,
-      left,
-      top,
-      width,
       height,
+      left,
       orientation,
-      title
+      title,
+      top,
+      width
     } = props;
     const isY = [LEFT, RIGHT].indexOf(orientation) > -1;
     const axisClassName = isY ? VERTICAL_CLASS_NAME : HORIZONTAL_CLASS_NAME;
--- a/src/lib/plot/axis/axis.js
+++ b/src/lib/plot/axis/axis.js
@@ -43,9 +43,6 @@ const propTypes = {
   orientation: React.PropTypes.oneOf([
     LEFT, RIGHT, TOP, BOTTOM
   ]),
-  axisType: React.PropTypes.oneOf([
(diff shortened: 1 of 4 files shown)


src/lib/plot/axis/x-axis.js
```diff
@@ -31,14 +32,21 @@ const propTypes = {
   ])
 };
 
+delete propTypes.axisType;
```
chrisirhc: Using `delete` prevents optimizations and it's a mutating operation. How about alternatives such as the rest operator or using lodash's `omit`?

src/lib/plot/axis/y-axis.js
```diff
@@ -32,14 +32,21 @@ const propTypes = {
...
 
 function YAxis(props) {
+  const axisProps = {
+    ...props,
+    axisType: DIRECTION.VERTICAL
+  };
+
   return (
-    <Axis {...props} />
+    <Axis {...axisProps} />
```
chrisirhc: You can supply additional props here by: `<Axis {...props} axisType={DIRECTION.VERTICAL} />`.
This might be cleaner as it doesn't require the introduction of a new variable (`axisProps` vs `props`).

src/lib/plot/axis/x-axis.js
mcnuttandrew: I'm not sure I understand what you are suggesting with rest parameters? I also haven't heard of delete preventing optimizations? (Though I would be excited to see source saying otherwise)

chrisirhc: Here's a related discussion: https://groups.google.com/forum/#!topic/v8-users/zE4cOHBkAnY
and https://github.com/google/google-api-nodejs-client/issues/375#issuecomment-76538779 .

For example with rest: `const {axisType, ...propsWithoutAxisType} = Axis.propTypes;`

src/lib/plot/axis/x-axis.js
```diff
@@ -19,13 +19,15 @@
...
+
 import Axis from './axis';
-import {ORIENTATION} from '../../utils/axis-utils';
+import {DIRECTION, ORIENTATION} from '../../utils/axis-utils';
 
 const {TOP, BOTTOM} = ORIENTATION;
 
+const {axisType, ...axisPropsWithoutType} = Axis.propTypes;
 const propTypes = {
-  ...Axis.propTypes,
```
chrisirhc: You can do custom validators, an even better option here might be to return an Error if axisType is specified when using x-axis:

https://facebook.github.io/react/docs/typechecking-with-proptypes.html

src/lib/plot/axis/axis.js
```diff
@@ -39,13 +43,17 @@ const propTypes = {
   orientation: React.PropTypes.oneOf([
     LEFT, RIGHT, TOP, BOTTOM
   ]),
+  axisType: React.PropTypes.oneOf([
+    DIRECTION.VERTICAL, DIRECTION.HORIZONTAL
+  ]).isRequired,
```
chrisirhc: Reading this PR in detail, could the `axisType` a derived from the `orientation`? For example, would there be a case where a `LEFT` or `RIGHT` orientation axis, would have a axisType of `HORIZONTAL`?

mcnuttandrew: i guess because they are required that's true

src/lib/plot/axis/axis.js
```diff
@@ -140,18 +149,20 @@ class Axis extends PureRenderComponent {
...
+      className,
       left,
       top,
       width,
       height,
       orientation,
       title
     } = props;
-
+    const isY = [LEFT, RIGHT].indexOf(orientation) > -1;
```
chrisirhc: I typically recommend `.includes` for bit of added readability, if there are existing usages of `.includes`, we should use that instead since that would mean existing users have it already set up.

src/lib/plot/axis/x-axis.js
```diff
@@ -19,6 +19,7 @@
 // THE SOFTWARE.
 
 import React from 'react';
+
```
chrisirhc: whitespace change unnecessary

src/lib/plot/axis/x-axis.js
```diff
@@ -38,7 +39,7 @@ const defaultProps = {
 
 function XAxis(props) {
   return (
-    <Axis {...props} />
+    <Axis {...props}/>
```
chrisirhc: Same here

src/lib/plot/axis/axis.js
```diff
@@ -140,18 +145,20 @@ class Axis extends PureRenderComponent {
...
       height,
+      left,
       orientation,
-      title
+      title,
+      top,
+      width
     } = props;
-
+    const isY = [LEFT, RIGHT].indexOf(orientation) > -1;
```
chrisirhc: I typically recommend `.includes` for bit of added readability, if there are existing usages of `.includes`, we should use that instead since that would mean existing users have it already set up.

mcnuttandrew: I'm totally with you, but there aren't any examples of includes in react-vis, while there are a number of instances of indexOf

src/lib/plot/axis/axis.js
```diff
@@ -71,6 +72,10 @@ const defaultProps = {
   orientation: BOTTOM
 };
 
+const predefinedClassName = 'rv-xy-plot__axis';
+const VERTICAL_CLASS_NAME = 'rv-xy-plot__axis-vertical';
```
chrisirhc: `--vertical` is the correct BEM convention. Do you think we'll get requests for separate classNames for left vs right?

src/lib/plot/axis/axis.js
chrisirhc: cool, no change here then, 👍  .

src/lib/plot/axis/axis.js
mcnuttandrew: i doubt it, i've also subverted that potential issue by allowing users to supply their own

src/lib/plot/axis/axis.js
chrisirhc: `isY` as a variable name might be a little confusing since there's an attribute and it may not be `y`

Issue #67, opened 2016-04-27
Add ability to customize bar width on bar series


chrisirhc:
Right now, the bar widths are automatically calculated, but there's no way to override it.

This is particularly useful for histograms.

![image](https://cloud.githubusercontent.com/assets/132584/14840593/5c887040-0bea-11e6-8a66-613b036b74d1.png)

bkrall:
+1

ranjithnori:
+1

greatermeans:
+1

k010:
+1

mcnuttandrew:
@chrisirhc This is addressed by rect-series, so i'm going to close this issue. Feel free to re-open if you feel different

sprzedwojski:
How would I use the rect-series if I need to set a custom bar width, but my x axis labels are strings?

My data:
```
[
  {
    "x": "Label 1",
    "y": 0.3
  },
  {
    "x": "Label 2",
    "y": 0.2
  },
  {
    "x": "Label 3",
    "y": 0.7
  }
]
```

It seems to be impossible to define an `x0` if my x axis labels are not numbers. 
Therefore I think setting a custom width on the vertical bar chart would be a useful feature.

treetopvt:
I also would find it useful to limit the width of a bar.  When working with ordinal data, I don't see how we can use a rect-series

sprzedwojski:
I managed to "hack" the rect-series to draw a vertical bar chart with strings as x-axis labels.
Below is my example code, maybe it helps somebody:

`const values = [
  {"y": 0.9},
  {"y": 0.3},
  {"y": 0.5}
]

const labels = [
  'Label 1',
  'Label 2',
  'Label 3'
]

class BarChartCustomBarWidth extends React.Component {
  
  // by tweaking this value you can resize the columns
  // 1 - columns touch each other
  // 0 - columns are invisible
  columnWidth = 0.5
  
  transformDataToRectSeries(raw) {
    const transformed = raw.map((el, i) => {
      el.x0 = (i + 1) - this.columnWidth / 2
      el.x = (i + 1) + this.columnWidth / 2
      return el
    })
    transformed.unshift({x: 0.5, y: 0}) // fake data point, enables margin between y-axis and first column
    transformed.push({x: 3.5, y: 0}) // fake data point, enables margin between last column and chart end
    return transformed
  }

  render() {
    return (
      <XYPlot width={500} height={400}>
        <XAxis
[...]

PR #197, opened 2016-12-10
Add rect series

mcnuttandrew:
In the current state of affairs bar chart centers data on the provide x value. Sometime this is great (eg when you have categorical data), but when you have data on a continuous variable this makes no sense. In such an enviroment you'd want to exactly specify where your boxes will sit. 

So to fill this gap, I propose adding a "histogramMode" to the bar series. The central change being a boolean flag on both of the bar series, that when activated sets the left edge of a bar to be x0, and the right edge to be x
screen shot 2016-12-09 at 9 15 43 pm

So rather than providing data for a bar chart like this:

[{x: 0, y: 3},{x: 1, y: 1}] 

You'd add a little more information

[{x0: 0, x: 1, y: 3},{x0: 1, x: 3, y: 1}] 

And get a lot more precision. This is a frequently requested feature (#67).

### DIFF
--- a/src/example/plot/stacked-bar-chart.js
+++ b/src/example/plot/stacked-bar-chart.js
@@ -26,33 +26,85 @@ import {
   YAxis,
   VerticalGridLines,
   HorizontalGridLines,
+  HorizontalBarSeries,
   VerticalBarSeries} from '../../';
 
 export default class Example extends React.Component {
   render() {
     return (
-      <XYPlot
-        width={300}
-        height={300}
-        stackBy="y">
-        <VerticalGridLines />
-        <HorizontalGridLines />
-        <XAxis />
-        <YAxis />
-        <VerticalBarSeries
-          data={[
-            {x: 1, y: 10},
-            {x: 2, y: 5},
-            {x: 3, y: 15}
-          ]}
-        />
-        <VerticalBarSeries
-          data={[
-            {x: 1, y: 12},
-            {x: 2, y: 2},
-            {x: 3, y: 11}
-          ]}/>
-      </XYPlot>
+      <div>
+        <XYPlot
+          width={300}
+          height={300}
+          stackBy="x">
+          <VerticalGridLines />
+          <HorizontalGridLines />
+          <XAxis />
+          <YAxis />
+          <HorizontalBarSeries
+            data={[
+              {y: 2, x: 10},
+              {y: 4, x: 5},
+              {y: 5, x: 15}
+            ]}
+          />
+          <HorizontalBarSeries
+            data={[
+              {y: 2, x: 12},
+              {y: 4, x: 2},
+              {y: 5, x: 11}
+            ]}/>
+        </XYPlot>
+
+        <XYPlot
+          width={300}
+          height={300}
+          stackBy="y">
+          <VerticalGridLines />
+          <HorizontalGridLines />
+          <XAxis />
+          <YAxis />
+          <VerticalBarSeries
+            data={[
+              {x: 2, y: 10},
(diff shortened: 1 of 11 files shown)


src/example/plot/stacked-bar-chart.js
```diff
@@ -26,33 +26,85 @@ import {
...
+          />
+          <HorizontalBarSeries
+            data={[
+              {y: 2, x: 12},
+              {y: 4, x: 2},
+              {y: 5, x: 11}
+            ]}/>
+        </XYPlot>
+
+        <XYPlot
```
mcnuttandrew: There weren't any examples of doing a stack horizontal chart, so i added one to be sure i wasn't breaking anything

src/lib/plot/series/area-series.js
```diff
@@ -44,7 +44,9 @@ class AreaSeries extends AbstractSeries {
   }
 
   render() {
-    const {animation, className, curve, data, marginLeft, marginTop} = this.props;
```
mcnuttandrew: lint was failing?

chrisirhc:
A few things here:
* This seems to be an API that allows for overlapping rectangles to be drawn. While that granularity of control might be desirable, I'm not sure we want to expose that right on the BarSeries, since it's now really more like a RectSeries where one needs to track x and x0.
* I find adding a mode to the same series and that changes the shape of the data required (requiring an x0) to be surprising.
* There's quite a bit of duplicate code for handling the two modes. This doesn't address #67 directly as it doesn't touch on changing either the width or spacing between the bars, it's instead putting it on the user to calculate all the positions of x0, x.

Have you considered modifying the logic of the bar series to use custom widths / spacing between bars?

mcnuttandrew:
> This seems to be an API that allows for overlapping rectangles to be drawn. While that granularity of control might be desirable, I'm not sure we want to expose that right on the BarSeries, since it's now really more like a RectSeries where one needs to track x and x0.

So if we were to move this logic into a RectSeries would that mean we would have  VerticalRectSeries and HorizontalRectSeries in addition?

mcnuttandrew:
> Have you considered modifying the logic of the bar series to use custom widths / spacing between bars? 

- I was considering using width initially, but all of the accessors are pretty hardcoded, eg if i wanted to express width in x coordinates, there's not a clean way to do that. Eventually it would be cool to expose a bunch of accessors so that anything could be manipulated, but i was trying to make a kinda low impact diff. 
- I'm not sure i get what you mean by spacing?

chrisirhc:
> So if we were to move this logic into a RectSeries would that mean we would have VerticalRectSeries and HorizontalRectSeries in addition?

The question becomes: are you blocked by the need for fine grained x, x0 control over the chart? Or do you need a bar series with no spacing in between the bars (see image in #67)?

src/example/plot/stacked-bar-chart.js
```diff
@@ -26,33 +26,59 @@ import {
...
-          ]}
-        />
-        <VerticalBarSeries
-          data={[
-            {x: 1, y: 12},
-            {x: 2, y: 2},
-            {x: 3, y: 11}
-          ]}/>
-      </XYPlot>
+      <div>
```
chrisirhc: Does this change break the layout of the docs page?

chrisirhc:
Also, need to update the title of this PR before merge

src/example/plot/stacked-bar-chart.js
mcnuttandrew: I don't mind the way this looks, but i'd change it if you take issue with it

PR #198, opened 2016-12-11
Remove non-existent file import in example/main.js

alee8:
Prevents example from running locally.


mcnuttandrew:
You beat me to it. i don't know how the example is compiling locally for me with this in place.

PR #200, opened 2016-12-12
Add additional orientation options in hint.js and provide 3 examples using these orientation options

alee8:
Add 8 more orientation options all related to edges:
-   'edgetopLeft',
-   'edgebottomLeft',
-   'edgetopRight',
-   'edgebottomRight',
-   'topEdgeleft',
-   'bottomEdgeleft',
-   'topEdgeright',
-   'bottomEdgeright'

Add 3 examples (dynamic-simple-edge-hints, dynamic-simple-topedge-hints, dynamic-complex-edge-hints) illustrating these new orientation options. 

Export ORIENTATION object with all the orientation constants and use them instead of the strings.
Convert examples to use these constants rather than the strings (for backwards compatibility, the original orientation strings are unchanged).

dynamic-complex-hints.js shows the power of using these edge orientation which also uses:
- divs to draw the hint-text box and the hint-pole; they are provided as children to the Hint component.  The hint-pole cannot be done with LineSeries as shown in (dynamic-simple-topedge-hints.js) because it needs to extend from bottom to top edge of plot and not from bottom of innerplot to top of plot (see image).
- flex to draw the text box beyond the top of the plot (using align-items: flex-end) and a few other flex tricks.  This demonstrates a pure div approach that could not be done with subclassing GridLines or using LineSeries for the pole because of the need to go from edge to edge of the plot. 
- exploits existing support by Hint to set margins and to use them to create short hint-pole-stem that connects with LineSeries (if you want to see what I mean, comment out the LineSeries in dynamic-complex-hints.js.

### DIFF
--- a/src/lib/plot/hint.js
+++ b/src/lib/plot/hint.js
@@ -23,6 +23,25 @@ import React from 'react';
 import PureRenderComponent from '../pure-render-component';
 import {getAttributeFunctor} from '../utils/scales-utils';
 
+/*
+ * Hint is placed in one of four quadrants around a data point (imagine the
+ * point bisected by two axes -vertical, horizontal- creating 4 quadrants around
+ * a data point).
+ *
+ * A new extension to Hint introduced the edges of chart/plot area as a valid
+ * place to pin one dimension (edge) of a data point where a hint is placed.
+ * The data point's other dimension determines the location along the pinned
+ * edge as well as which side of the edge. Thus, the data point's
+ * x (if top/bottom edge is pinned) or y (if left/right edge is pinned) value
+ * determines the location of the edge plus descriptor of which side (top/bottom
+ * if left/right edge is pinned or left/righ if top/bottom edge is pinned).
+ *
+ * The constants identify positioning the Hint in either a quadrant or
+ * combination of an edge and an axis.
+ *
+ * The constants use the following naming format:
+ *   <edgetop | edgebottom | top | bottom>_<edgeleft | edgeright | left | right>
+ */
 const ORIENTATION = {
   AUTO: 'auto',
   TOPLEFT: 'topleft',
@@ -94,12 +113,18 @@ class Hint extends PureRenderComponent {
    * @private
    */
   _getCSSRight(x) {
+    if (x === undefined || x === null) {
+      return {
+        right: 0
+      };
+    }
+
     const {
       innerWidth,
       marginRight
     } = this.props;
     return {
-      right: x === undefined || x === null ? 0 : marginRight + innerWidth - x
+      right: marginRight + innerWidth - x
     };
   }
 
@@ -111,9 +136,15 @@ class Hint extends PureRenderComponent {
    * @private
    */
   _getCSSLeft(x) {
+    if (x === undefined || x === null) {
+      return {
+        left:  0
+      };
+    }
+
     const {marginLeft} = this.props;
     return {
-      left: x === undefined || x === null ? 0 : marginLeft + x
+      left:  marginLeft + x
     };
   }
 
@@ -125,12 +156,18 @@ class Hint extends PureRenderComponent {
    * @private
    */
   _getCSSBottom(y) {
+    if (y === undefined || y === null) {
+      return {
+        bottom: 0
+      };
+    }
+
     const {
       innerHeight,
       marginBottom
     } = this.props;
     return {
-      bottom: y === undefined || y === null ? 0 : marginBottom + innerHeight - y
+      bottom: marginBottom + innerHeight - y
     };
   }
 
@@ -142,9 +179,15 @@ class Hint extends PureRenderComponent {
    * @private
    */
   _getCSSTop(y) {
+    if (y === undefined || y === null) {
+      return {
+        top: 0
+      };
+    }
+
     const {marginTop} = this.props;
     return {
-      top: y === undefined || y === null ? 0 : marginTop + y
+      top: marginTop + y
     };
   }
 
--- a/src/lib/plot/hint.js
+++ b/src/lib/plot/hint.js
@@ -23,11 +23,21 @@ import React from 'react';
 import PureRenderComponent from '../pure-render-component';
 import {getAttributeFunctor} from '../utils/scales-utils';
 
-const ORIENTATION_AUTO = 'auto';
-const ORIENTATION_TOPLEFT = 'topleft';
-const ORIENTATION_BOTTOMLEFT = 'bottomleft';
-const ORIENTATION_TOPRIGHT = 'topright';
-const ORIENTATION_BOTTOMRIGHT = 'bottomright';
+const ORIENTATION = {
+  AUTO: 'auto',
+  TOPLEFT: 'topleft',
+  BOTTOMLEFT: 'bottomleft',
+  TOPRIGHT: 'topright',
+  BOTTOMRIGHT: 'bottomright',
+  EDGETOP_LEFT: 'edgetopLeft',
+  EDGEBOTTOM_LEFT: 'edgebottomLeft',
+  EDGETOP_RIGHT: 'edgetopRight',
+  EDGEBOTTOM_RIGHT: 'edgebottomRight',
+  TOP_EDGELEFT: 'topEdgeleft',
+  BOTTOM_EDGELEFT: 'bottomEdgeleft',
+  TOP_EDGERIGHT: 'topEdgeright',
+  BOTTOM_EDGERIGHT: 'bottomEdgeright'
+};
 
 /**
  * Default format function for the value.
@@ -40,7 +50,7 @@ function defaultFormat(value) {
   });
 }
 
-class Hint extends PureRenderComponent {
+const Hint = class Hint extends PureRenderComponent {
 
   static get propTypes() {
     return {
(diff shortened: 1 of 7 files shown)


src/lib/plot/hint.js
```diff
@@ -215,15 +269,18 @@ class Hint extends PureRenderComponent {
 
   render() {
     const {
+      className,
+      style,
```
alee8: Will drop this addition since this requires more broader discussion for the package.

src/lib/plot/hint.js
```diff
@@ -215,15 +269,18 @@ class Hint extends PureRenderComponent {
...
 
-    const {style, className} = this._getPositionInfo();
+    const posInfo = this._getPositionInfo();
     return (
       <div
-        className={`rv-hint ${className}`}
+        className={`rv-hint ${posInfo.className} ${className}`}
         style={{
+          ... posInfo.style,
           ... style,
```
alee8: Will drop this line.  Related to comment above about adding className and style to Hint props; this merits a broader discussion for the react-vis library

src/lib/plot/hint.js
```diff
@@ -40,7 +50,7 @@ function defaultFormat(value) {
   });
 }
 
-class Hint extends PureRenderComponent {
+const Hint = class Hint extends PureRenderComponent {
```
chrisirhc: Seems like an unnecessary change

src/lib/plot/hint.js
```diff
@@ -52,75 +62,89 @@ class Hint extends PureRenderComponent {
...
    */
   _getCSSRight(x) {
     const {
       innerWidth,
-      marginRight} = this.props;
+      marginRight
+    } = this.props;
     return {
-      right: `${marginRight + innerWidth - x}px`
+      right: typeof x === 'number' ? marginRight + innerWidth - x : 0
```
chrisirhc: Seems like an alternative here is to accept top/left/right/bottom overridden by the user.

Also, rather than handling `undefined`, `null` is preferable as it looks intentional, while `undefined` looks like a user error. When handling those cases, prefer to use `x === null` / !== as it's explicit.

src/lib/plot/hint.js
```diff
@@ -52,75 +62,89 @@ class Hint extends PureRenderComponent {
...
+   * When x undefined, edge case, pin left.
    * @param {number} x X.
    * @returns {{left: *}} Mixin.
    * @private
    */
   _getCSSLeft(x) {
     const {marginLeft} = this.props;
     return {
-      left: `${marginLeft + x}px`
+      left: typeof x === 'number' ? marginLeft + x : 0
```
chrisirhc: Same as above

src/lib/utils/scales-utils.js
```diff
@@ -657,4 +657,4 @@ export default {
   getScaleObjectFromProps,
   getScalePropTypesByAttribute,
   literalScale
-}
+};
```
alee8: change reverted.

src/lib/plot/hint.js
```diff
@@ -215,15 +269,18 @@ class Hint extends PureRenderComponent {
...
   render() {
     const {
+      className,
+      style,
       value,
       format,
       children} = this.props;
 
-    const {style, className} = this._getPositionInfo();
+    const posInfo = this._getPositionInfo();
```
alee8: change reverted since this PR will not add style and className as props to Hing

src/lib/plot/hint.js
alee8: reverted.

src/lib/plot/hint.js
alee8: All these methods are not public.  Hence user cannot override.  If you think this is desirable, the API to hint should be reworked.
I will test for undefined and null. using ===

src/lib/plot/hint.js
```diff
@@ -52,75 +62,89 @@ class Hint extends PureRenderComponent {
...
    */
   _getCSSRight(x) {
     const {
       innerWidth,
-      marginRight} = this.props;
+      marginRight
+    } = this.props;
     return {
-      right: `${marginRight + innerWidth - x}px`
+      right: x === undefined || x === null ? 0 : marginRight + innerWidth - x
```
chrisirhc: Nitpick: I would early return and use only `null` rather than allow `undefined`, since allowing both allows ambiguity and `undefined` is ambiguous.

Also, early return because when it's `null`, there's no need for `innerWidth` or `marginRight`. It may improve readability slightly.

src/lib/plot/hint.js
```diff
@@ -23,11 +23,21 @@ import React from 'react';
...
-const ORIENTATION_BOTTOMLEFT = 'bottomleft';
-const ORIENTATION_TOPRIGHT = 'topright';
-const ORIENTATION_BOTTOMRIGHT = 'bottomright';
+const ORIENTATION = {
+  AUTO: 'auto',
+  TOPLEFT: 'topleft',
+  BOTTOMLEFT: 'bottomleft',
+  TOPRIGHT: 'topright',
+  BOTTOMRIGHT: 'bottomright',
+  EDGETOP_LEFT: 'edgetopLeft',
```
chrisirhc: We should be clearer what these refer to. Perhaps a quick description or some ascii illustration?

Also, an alternative to improve readability: `EDGE_TOP__LEFT`.

On reading this without checking the code, I don't find it particularly clear what `EDGETOP_LEFT` does. It aligns to the top left, and the top edge?

alee8: I just followed the style of naming that was part of this original API.  Orientation value follows this pattern:
  <top | bottom><left | right>

So, when Edgetop, Edgebottom, Edgeright, Edgeleft was introduced, I put an _ to allow better parsing of the constants name that respects the original naming scheme, that is:
  <edgetop | edgebottom | top | bottom>_<edgeleft | edgeright | left | right>

Since PR moving forwards introduce the constants, I did not change the original orientation string values by incorporating the _.  Apps using earlier version of this API choosing to update to this version would experience breaking change if they don't switch their use of the original string values 'topleft', 'topright', 'bottomleft', 'bottomright' to the constants.

I'll leave the naming as is because one might easily miss out the double underscore.

I'll add some comments to the code where the constants are defined that explains the constants.

src/lib/plot/hint.js
alee8: Sure. will do that.

alee8:
@chrisirhc Your comments [to clarify ORIENTATION constants](https://github.com/uber/react-vis/pull/200/files/b12458ce05c6325ad903a5dff00d4464cf2a8cf9#r92089079) and [to accept top/left/right/bottom css props to override Hint behavior ](https://github.com/uber/react-vis/pull/200/files/b12458ce05c6325ad903a5dff00d4464cf2a8cf9#r92051269)provoked me to re-think the existing PR.

I would like to propose the following to address and reduce your underlying concern while maintaining the original spirit of Hint orientation props in the API.  I believe your underlying concern was the long list of orientation constants options and the conflating of two axis/edges into one property.  I suggest we tease apart the orientation property into two separate properties that better address the axes/edges for placement of hints.

Specifically, the two new properties are horizontalAlign and verticalAlign. They have the following values.  See the attached image elaborating on the values and a few examples illustrating hint placements.

- horizontalAlign: auto, left, right, leftEdge, rightEdge
- verticalAlign: auto, top, bottom, topEdge, bottomEdge

As clarified earlier, in this PR, hints are placed either in one of the data point's 4 quadrants (resulting from two axes bisecting data point) or is pinned to one edge of the plot (leftEdge, rightEdge, topEdge, bottomEdge) and positioned along that edge using data point's other dimension value.

For auto, it would follow the existing notion: place hint as determined by the x or y value but aligning to fit within the inner portion of the plot.  This would also work if the other dimension were pinned to an edge.

This change will drop support of existing ORIENTATION property and its string value in favor of the above proposal.  It will be a breaking change to the Hint API.

Let me know what you think about this? If you agree with the change, I'll post an update changing the Hint API and examples.

Alison
react-vis-hint-prop

chrisirhc:
@alee8  it sounds like there's a lot of customization required for your use cases. I've written up an alternate propsal at #203 with the idea of having overrides for any of the positions. Will that be adequate for your use cases?

chrisirhc:
The reason I'm reluctant to go towards providing a lot of API surface area to deal with positioning is that there are existing libraries that already do this (e.g. http://tether.io/) , and seeing the API of these libraries, there's a certain level of functionality required before the positioning API becomes useful.

I'm proposing to leave it out of the react-vis API for now and allow the user to decide how they would like to position or override the positions so that we can serve both basic and advanced use cases without spending much time / effort on building an API that may not meet either well in its early stages.

mcnuttandrew:
Hey Alison.

I really like your diagram! Generally I feel weird about introducing a whole new system for describing the positions of things. While yours is really comprehensive, it feels like something that could easily snowball and get out of hand with time. Also, to chris's point

> The reason I'm reluctant to go towards providing a lot of API surface area to deal with positioning is that there are existing libraries that already do positioning really well (e.g. http://tether.io/) , and seeing the API of these libraries, there's a certain level of functionality required before the positioning API becomes useful to power users.

I think the optimal strategy here is to simply provide overrides so that a more developed api can do the work. Unless I'm really missing something, it seems like #203 is maybe a little more defensive in the exposure of our api footprint.Finally, can you please add a documentation task for the hint? It would be great if we could build up the pain points of documentation so that we can prioritize them when we start in on it.

alee8:
Have a conceptually cleaner proposal in [PR#204](https://github.com/uber/react-vis/pull/204) with examples in [PR#205](https://github.com/uber/react-vis/pull/205).

Closing this one out. This was valuable for thinking through and articulating the key issue for this PR (addition of user-friendly, non-programmatic means of attaching hints to edges of a plot).

PR #204, opened 2016-12-22
Hint component: Deprecate orientation prop in favor of two new prop (horizontalAlign, verticalAlign)

alee8:
This is the set of changes to realize the new proposal for Hint that I put forward about a week ago.  

Here's the proposal in its entirety:
react-vis-hint-prop

@chrisirhc Your comments [to clarify ORIENTATION constants](https://github.com/uber/react-vis/pull/200/files/b12458ce05c6325ad903a5dff00d4464cf2a8cf9#r92089079) and [to accept top/left/right/bottom css props to override Hint behavior ](https://github.com/uber/react-vis/pull/200/files/b12458ce05c6325ad903a5dff00d4464cf2a8cf9#r92051269) provoked me to re-think the previous [PR](https://github.com/uber/react-vis/pull/200).  This new PR would replace the previous [PR](https://github.com/uber/react-vis/pull/200).

I would like to propose the following to address and reduce your underlying concern while maintaining the original spirit of Hint orientation props in the API.  I believe your underlying concern was the long list of orientation constants options and the conflating of two axis/edges into one property.  I suggest we tease apart the orientation property into two separate properties that better address the placement of hints.

Specifically, the two new properties are horizontalAlign and verticalAlign. They have the following values.  See the attached image elaborating on the values and a few examples illustrating hint placements.

- horizontalAlign: auto, left, right, leftEdge, rightEdge
- verticalAlign: auto, top, bottom, topEdge, bottomEdge

This proposal extends the current option of Hint placement with a new, second option:

1. Current - around one of the data point's 4 quadrants (resulting from two axes bisecting data point).
2. New - pinned to one edge of the plot (leftEdge, rightEdge, topEdge, bottomEdge) and positioned along that edge using data point's other dimension value.

For auto, it would follow the existing notion: place hint as determined by the x or y value but aligning to fit within the inner portion of the plot.  This would also work if the other dimension were pinned to an edge.

This change will drop support of existing ORIENTATION property and its string value in favor of the above proposal.  It will be a breaking change to the Hint API.

The merits of this proposal compared to the alternate proposal from @chrisirhc - [hint positionOverrides proposal](https://github.com/uber/react-vis/pull/203) are:

1. The two placement options are practically the only two possible placement of Hints.  CSS styling and flex layout will enable offset adjustment, if desired, with each option. 
2. This will enable most users to use Hint, out of the box, by selecting the desired placement and then focusing on CSS styling.
3. @chrisirhc proposal enable a power-user option - programmatic determination of positionOverrides(left, right, top, bottom) beyond setting their values to 0.  The utils exposed in this API would facilitate mapping domain scale to screen scale for setting these values. However, the need for this option and the audience for programmatic scheme are very small compared to the placement options proposed herein.

### DIFF
--- a/src/lib/plot/hint.js
+++ b/src/lib/plot/hint.js
@@ -23,11 +23,34 @@ import React from 'react';
 import PureRenderComponent from '../pure-render-component';
 import {getAttributeFunctor} from '../utils/scales-utils';
 
-const ORIENTATION_AUTO = 'auto';
-const ORIENTATION_TOPLEFT = 'topleft';
-const ORIENTATION_BOTTOMLEFT = 'bottomleft';
-const ORIENTATION_TOPRIGHT = 'topright';
-const ORIENTATION_BOTTOMRIGHT = 'bottomright';
+/*
+ * Hint provides two options for placement of hint:
+ * a) around a data point in one of four quadrants (imagine the point bisected
+ *    by two axes -vertical, horizontal- creating 4 quadrants around a data
+ *    point).
+ * b) **New** pin to an edge of chart/plot area and position along that edge
+ *    using data point's other dimension value.
+ *
+ * To support these two options, deprecate one Hint props (orientation) with two
+ * new Hint props (horizontalAlign, verticalAlign) with following values:
+ *
+ *   horizontalAlign: auto, left, right, leftEdge, rightEdge
+ *   verticalAlign: auto, bottom, top, bottomEdge, topEdge
+ *
+ * Thus, the following ALIGN constants are the values for horizontalAlign
+ * and verticalAlign
+ */
+const ALIGN = {
+  AUTO: 'auto',
+  LEFT: 'left',
+  RIGHT: 'right',
+  LEFT_EDGE: 'leftEdge',
+  RIGHT_EDGE: 'rightEdge',
+  BOTTOM: 'bottom',
+  TOP: 'top',
+  BOTTOM_EDGE: 'bottomEdge',
+  TOP_EDGE: 'topEdge'
+};
 
 /**
  * Default format function for the value.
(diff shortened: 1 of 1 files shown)


src/lib/plot/hint.js
```diff
@@ -199,17 +285,18 @@ class Hint extends PureRenderComponent {
...
     const y = getAttributeFunctor(this.props, 'y')(value);
 
-    const orientation = initialOrientation === ORIENTATION_AUTO ?
-      this._getOrientationFromAuto(x, y) : initialOrientation;
+    const align = this._getAlign(x, y);
 
     return {
-      style: this._getOrientationStyle(orientation, x, y),
-      className: this._getOrientationClassName(orientation)
+      style: this._getAlignStyle(align, x, y),
```
chrisirhc: How about optionally taking a prop to replace this function call with something like `this.props.getPositionStyle(this.props, x, y)`.

The default prop for the `getPositionStyle` will be provided by us with these align/orientation capabilities. However, they'll be overridable.

src/lib/plot/hint.js
```diff
@@ -199,17 +285,18 @@ class Hint extends PureRenderComponent {
...
+      className: this._getAlignClassNames(align)
```
chrisirhc: Can be `this.getClassNames(this.props)`, and that method can call `_getOrientationClassName` or `_getAlignClassNames` depending on whether `align` is used.

src/lib/plot/hint.js
```diff
@@ -51,143 +84,250 @@ class Hint extends PureRenderComponent {
...
+   * Get the class names from align values.
+   * @param {Object} align object with horizontal and vertical prop strings.
+   * @returns {string} Class names.
    * @private
    */
-  _getOrientationClassName(orientation) {
-    return `rv-hint--orientation-${orientation}`;
+  _getAlignClassNames(align) {
+    return `rv-hint--horizontalAlign-${align.horizontal}
+     rv-hint--verticalAlign-${align.vertical}`;
```
chrisirhc: We should leave in the `rv-hint--orientation-` classnames from before if the user is using orientation so that users that still depend on it can still use it.

src/lib/plot/hint.js
```diff
@@ -51,143 +84,250 @@ class Hint extends PureRenderComponent {
...
+    }
+
     const {
       innerHeight,
-      marginBottom} = this.props;
+      marginBottom
+    } = this.props;
     return {
-      bottom: `${marginBottom + innerHeight - y}px`
+      bottom: marginBottom + innerHeight - y
```
chrisirhc: no `px`?

alee8: Not necessary as inline style supports bottom as an int.

src/lib/plot/hint.js
alee8: Okay, I will add it in if orientation exists.

chrisirhc:
Looks good. Will approve once changes are done.

alee8:
Thanks. 
The final change was to create new align object prop with {horizontal, vertical} props.
Maintain backward compatibility for those using orientation prop (remapping to align object).

PR #205, opened 2016-12-22
Examples only for PR204 to Deprecate orientation prop in favor of two new props (horizontalAlign, verticalAlign)

alee8:
This is the examples only portion that uses the new proposed Hint props - horizontalAlign, verticalAlign - ([PR#204](https://github.com/uber/react-vis/pull/204)) .  PR#204 contained only the change to hint.js  So, this PR is only useful if [PR#204](https://github.com/uber/react-vis/pull/204) is accepted.

Added 3 new examples to illustrate the new Hint props (see attached figure) and to serve as test of the new Hint props:
1. Dynamic Simple Edge Hints - Illustrates each point using different horizontalAlign, verticalAlign options.  The hint box includes text on which option is used.  There is no additional CSS styling. So, the example illustrates edge and quadrant placement clearly before any CSS treatment.
2. Dynamic Simple Top Edge Hints - illustrates pinning hint to top edge but using AUTO for placement of hint box along the top edge.  The Hint's AUTO algorithm uses horizontalAlign=RIGHT for the first 2 points as they are on the left side of the plot and horizontalAlign=LEFT for the last 2 points as they are on the right side of the plot.  The pole in the hint is created using additional combination of MarkSeries and Line Series to create the pole line.
3. Dynamic Complex Edge Hints - illustrates the 4 edge placement represented by the 4 data points.  More importantly, it shows how the provided children contain divs with CSS classes that allow creation of pole line (using CSS) and how the hint box with a short stem (extending from the plot edge) can be offset (using flex layout). It supports the argument that offsets can be handled with CSS.

hint-align

### DIFF
--- a/src/example/main.js
+++ b/src/example/main.js
@@ -44,6 +44,7 @@ import DynamicHints from './plot/dynamic-hints';
 import DynamicComplexEdgeHints from './plot/dynamic-complex-edge-hints';
 import DynamicSimpleEdgeHints from './plot/dynamic-simple-edge-hints';
 import DynamicSimpleTopEdgeHints from './plot/dynamic-simple-topedge-hints';
+import DynamicProgrammaticRightEdgeHints from './plot/dynamic-programmatic-rightedge-hints';
 import StaticCrosshair from './plot/static-crosshair';
 import DynamicCrosshair from './plot/dynamic-crosshair';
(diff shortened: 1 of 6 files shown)


alee8:
Thanks,
Added examples that illustrate and serve as test verification of new Hint align prop.

PR #206, opened 2016-12-29
bound _formatCrosshairTitle with this

keshavnandan:
_formatCrosshairTitle method wasn't bound to this in the constructor. This commit fixes that.


src/example/plot/complex-chart.js
```diff
@@ -81,6 +81,7 @@ export default class Example extends React.Component {
     this._updateButtonClicked = this._updateButtonClicked.bind(this);
     this._legendClickHandler = this._legendClickHandler.bind(this);
     this._formatCrosshairItems = this._formatCrosshairItems.bind(this);
+    this._formatCrosshairTitle = this.this._formatCrosshairTitle.bind(this);
```
chrisirhc: `this.this._formatCrosshairTitle` seems like a typo

keshavnandan: I have fixed it in the latest commit

PR #216, opened 2017-01-10
Added support for eslint cache

amilajack:


amilajack:
@mcnuttandrew accidentally added you to assignees instead of reviewers. My bad

Issue #215, opened 2017-01-10
Area Series opacity regards 0 as falsy 


roderickhsiao:
https://github.com/uber/react-vis/blob/master/src/lib/plot/series/area-series.js#L68

```js    
const opacity = this._getAttributeValue('opacity') || DEFAULT_OPACITY;
```

We are having use case that we are conditionally passing opacity 0 to hide the chart, the above code will regard zero as falsy and always fallback to `DEFAULT_OPACITY`;

Probably specifically check for Nil value?

Thanks

mcnuttandrew:
Good catch! Ill cut a new version tomorrow morning

mcnuttandrew:
Alright there's a new version up 0.8.1 which fixes this. Let me know if I can do anything else!

roderickhsiao:
Oh so fast! Just about to send a PR and you guys fix it :) thanks a lot

PR #218, opened 2017-01-10
Ensure opacity equals zero is interpreted 0 and not falsy

mcnuttandrew:
Responding to #215

### DIFF
--- a/src/example/plot/area-chart-elevated.js
+++ b/src/example/plot/area-chart-elevated.js
@@ -49,7 +49,7 @@ export default class AreaChartElevated extends React.Component {
           ]}/>
         <AreaSeries
           className="area-elevated-series-2"
-          color="12939a"
+          color="#12939a"
           data={[
             {x: 1, y: 5, y0: 6},
             {x: 2, y: 20, y0: 11},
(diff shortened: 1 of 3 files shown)


src/example/plot/area-chart-elevated.js
```diff
@@ -49,7 +49,7 @@ export default class AreaChartElevated extends React.Component {
           ]}/>
         <AreaSeries
           className="area-elevated-series-2"
-          color="12939a"
+          color="#12939a"
```
mcnuttandrew: this isn't connected to this task, it's an ancillary typo

src/lib/plot/series/area-series.js
```diff
@@ -65,7 +65,8 @@ class AreaSeries extends AbstractSeries {
       this._getAttributeValue('color');
     const fill = this._getAttributeValue('fill') ||
       this._getAttributeValue('color');
-    const opacity = this._getAttributeValue('opacity') || DEFAULT_OPACITY;
+    const givenOpacity = this._getAttributeValue('opacity');
+    const opacity = !isNaN(givenOpacity) ? givenOpacity : DEFAULT_OPACITY;
```
chrisirhc: `isNaN` has some quirks with coercion and its behavior is quite confusing (see how it handles true/false/undefined). How about using `Number.isFinite`? This also avoids the double negation here.

Refs:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

PR #220, opened 2017-01-10
Updated react, babel, eslint deps, fixed linter errors

amilajack:
How would I go about manually testing this? Is there any kind of screenshot testing solution?


src/example/main.js
```diff
@@ -283,6 +283,8 @@ const examples = (
 // Cannot render to body anymore: react is throwing warnings.
```
ibgreen: You could make it more clear in the comments that this is needed to support React 14 (I assume that is the reason we do this).

amilajack:
@ibgreen Updated the comments about the `React.render` deprecation to reflect your comments. Thanks!

PR #223, opened 2017-01-11
Support classname for legend

roderickhsiao:
Will be good to support class name append for legend


src/lib/legends/discrete-color-legend.js
```diff
@@ -57,11 +57,11 @@ function fillItemsWithDefaults(items) {
 }
 
 function DiscreteColorLegend({items: initialItems, width, height,
-  onItemClick, orientation}) {
+  onItemClick, orientation, className = ''}) {
```
chrisirhc: We'll need to add this into the PropTypes as well

roderickhsiao: sure thing

PR #225, opened 2017-01-11
Updated npm engines config

amilajack:
Its better that we restrict node versions to node >= 4. I believe node 0.10 was deprecated.


mcnuttandrew:
I think just up to 4 is a little bit aggressive for us? I think 0.12 would be a little safer, though I am open to argument?

package.json
```diff
@@ -80,7 +80,7 @@
     "prepublish": "npm run build"
   },
   "engines": {
-    "node": "0.10.x",
-    "npm": "3.x"
+    "node": ">=4.x",
```
amilajack: Good idea. I'll change node engine to node >= 0.10

package.json
```diff
@@ -80,7 +80,7 @@
     "prepublish": "npm run build"
   },
   "engines": {
-    "node": "0.10.x",
-    "npm": "3.x"
+    "node": ">=0.10.x",
+    "npm": ">=3.x"
```
chrisirhc: Last I checked using npm with `>=` and `.x` might be invalid. Need to `.0` or leave out to patch/minor version.

package.json
```diff
@@ -80,7 +80,7 @@
     "prepublish": "npm run build"
   },
   "engines": {
-    "node": "0.10.x",
-    "npm": "3.x"
+    "node": ">=0.10.0",
+    "npm": ">=3.x"
```
chrisirhc: missed npm

amilajack:
Also I just checked a project that I'm a collaborator of and [we use `>=3.x` for node/npm semver targets](https://github.com/chentsulin/electron-react-boilerplate/blob/master/package.json#L173-L176). I've never run into an issue so i think this is okay. But its more clear to use a fixed semver, like `>=3.0`

Issue #221, opened 2017-01-10
Undefined append on classname for Axis


roderickhsiao:
very minor but if user doesn't pass in `className` for Axis
https://github.com/uber/react-vis/blob/master/src/lib/plot/axis/axis.js#L162

`const {className} = this.props` will become `undefined`

we can either add a defaultProps (empty className) or `const {className = ''} = this.props` I guess.

mcnuttandrew:
Another good catch! Do you wanna shoot off a PR for this one?

roderickhsiao:
Sure thing, this is an easy one :)

PR #227, opened 2017-01-11
[Issue 221] Add default props class name for Axis

roderickhsiao:
Address #221 
I am adding `defaultProps` to match the pattern of `abstract-series`
https://github.com/uber/react-vis/blob/master/src/lib/plot/series/abstract-series.js#L51-L53


roderickhsiao:
Thanks!

Issue #226, opened 2017-01-11
Margin on rendered SVG


dcousens:
In the below image, screen-capped from the examples page,  there is a noticeable bottom-right margin as part of the rendered SVG.
I've noticed this even on with my own usage of the library.

Am I missing a CSS trick to center/crop this?

![svg](https://cloud.githubusercontent.com/assets/413395/21851356/da72eab8-d862-11e6-972e-9bf2ae89ace0.png)
(Source: https://uber.github.io/react-vis/)

mcnuttandrew:
No, your weren't missing anything! It was a problem on our end, thanks for catching that

PR #229, opened 2017-01-11
Demand pie chart actually be in the center

mcnuttandrew:
This addresses #226

### DIFF
--- a/src/lib/radial-chart/radial-chart.js
+++ b/src/lib/radial-chart/radial-chart.js
@@ -245,10 +245,6 @@ class RadialChart extends React.Component {
       return null;
     }
 
-    const {innerWidth, innerHeight} = getInnerDimensions(
-      this.props,
-      DEFAULT_MARGINS
-    );
     const opacityFunctor = this._getAttributeFunctor('opacity');
     const fillFunctor = this._getAttributeFunctor('fill') ||
       this._getAttributeFunctor('color');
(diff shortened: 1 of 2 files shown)


dcousens:
Tested locally,  it works! :+1: ACK

src/lib/radial-chart/radial-chart.js
```diff
@@ -245,10 +245,6 @@ class RadialChart extends React.Component {
       return null;
     }
 
-    const {innerWidth, innerHeight} = getInnerDimensions(
```
chrisirhc: Is `InnerDimensions` broken?
Looks like `DEFAULT_MARGINS` is no longer used.

mcnuttandrew: it's not broken, it was just being used incorrectly. Pie charts are by their nature centered in their containers and having their sizes modified modifying the range of the slices. The old situation lead to stuff like this 

https://cloud.githubusercontent.com/assets/413395/21851356/da72eab8-d862-11e6-972e-9bf2ae89ace0.png

which was not ideal

mcnuttandrew: also, DEFAULT_MARGINS is still in use in _getDefaultScaleProps

chrisirhc: Gotcha. Only asking this because I see that there's a top margin added to the example radial chart. Wondering whether removing this means that users need to add that margin to their usages.

mcnuttandrew: I added that to the example to indicate how it works, but it seems like we could also probably expand the documentation around the usage of margins as well

Issue #208, opened 2017-01-09
Automatic gh-pages update system


mcnuttandrew:
We want to guarantee that our gh-pages are always up to date with the latest examples! We should write a simple build script so that contributing to the docs can be easy and quick.

mcnuttandrew:
One suggestion is build a separate package.json for the example so that it can be built directly against it's own requirements, this will also thin out our devDeps & make them less menacing.

amilajack:
So for instance, if example A depends on npm modules b and c, we will need to create a `package.json `, with dependencies `b` and `c`, specifically for example A

amilajack:
Also where is the gh-pages repo?

ibgreen:
@amilajack 

For reference, I have just started to create separate examples (with separate package.json) in [`react-map-gl` dev branch](https://github.com/uber/react-map-gl/tree/dev/examples/main).

> Also where is the gh-pages repo?
`gh-pages` is not a repo, it is a branch in each repo. Anything on those branches is served by github on github pages urls.

amilajack:
@mcnuttandrew should each example in each directory have have its own `index.html`, `package.json`, `webpack.config.js`, etc? Or should we create them for each example directory (for example one `index.html`, `package.json` etc for all legends or one for each legend example)?

mcnuttandrew:
Well if we are following deck.gl's lead, then we would have a folder structure like

// examples
// // example 1
// // // package.json
// // // index.html
// // // ..etc
// // example 2
// // // package.json
// // // index.html
// // // ..etc

which would allow people to clone those examples exactly in order to get started. (That said, the barrier to entry is substantially lower for react-vis than deck.gl) 

I think we should shoot for a structure that will allow us to build a central documentation page that has each of these examples embedded within it, so it seems like the above structure would work well? I think in an ideal world out compile system would merge these all together and shove them over to the gh-pages.

amilajack:
@mcnuttandrew thanks for the clarification. I've started the migration of the examples in my local branch. I also have another question. Should the examples use the published npm version of the package or the compiled source of the git repo itself (the files in `dist`)? `react-map-gl`'s example rely on the npm package. I think we should take the npm package approach because its safer in terms of breaking changes and its easier to replicate. Is this recommended?

amilajack:
[Here's](https://github.com/amilajack/react-vis/tree/refactor-doc-examples/src/example/legends) an initial implementation of the legends examples

[...]

PR #232, opened 2017-01-13
Chore refactors

Apercu:
I abandoned the move of the compiled `main.css` file, since I can't make it compatible with both sass and vanilla css imports, we'll wait for the next major.

This introduces a change of arch by moving stuff around, doing minor es6 refactors, a nice readme, better DX with webpack, linting, named exports, coverage published on Coveralls, dedicated changelog file, and some more 😃 

This it will be easier to review separate commits, I'll wait for this to rebase.


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "react-vis",
   "version": "0.9.0",
   "license": "MIT",
-  "author": "Anton Bulyenov <antonb@uber.com>",
+  "author": "Anton Bulyenov <visualization@uber.com>",
```
mcnuttandrew: Should probably updated to someone who is not anton, as he isn't a maintainer anymore

package.json
```diff
@@ -63,14 +63,15 @@
     "react-addons-shallow-compare": "0.14.x - 15.x"
   },
   "scripts": {
+    "build-clean": "rm -rf dist && mkdir dist",
     "build-js": "babel src -d dist --copy-files",
     "build-js-example": "npm run build-js -- --source-maps inline",
     "build-docs": "(node-sass src/example/main.scss example.css) & (browserify -t brfs ./dist/example/main.js > example.js)",
-    "build-css": "node-sass src/main.scss main.css",
```
mcnuttandrew: it's my understanding that this is actually necessary as the user needs to require node-modules/react-vis/main.css to get the app running

package.json
Apercu: Yes, or more globally, "Visualization team"?

package.json
Apercu: I only changed the path of the compiled main css file for it to be inside the `dist` folder instead

package.json
mcnuttandrew: yea i think that's reasonable

package.json
mcnuttandrew: so will users have require dist/main.css instead?

Apercu: Yes, I updated the readme to make it clear

> Note that the change of the css path is breaking, but worth in my opinion.

main.css
```diff
@@ -0,0 +1 @@
+@import "./dist/main";
```
Apercu: @mcnuttandrew Importing this file in sass works, but not sure if it will using vanilla css, do you have an idea?

package.json
```diff
@@ -2,25 +2,27 @@
...
+  },
+  "scripts": {
+    "docs": "echo 'TODO'",
+    "start": "(cd examples && command -v yarn >/dev/null && yarn && npm start || npm install && npm start)",
+    "clean": "rm -rf dist && mkdir dist",
+    "build": "npm run clean && babel src -d dist --copy-files && node-sass src/main.scss main.css",
+    "lint": "eslint src tests examples --ignore-pattern node_modules",
+    "test": "tape -r babel-register tests/**/*.js",
+    "cover": "nyc --reporter=text --reporter=html --reporter=lcov npm test"
   },
```
amilajack: I see that the examples and `build-*` scripts were removed. Are they no longer needed?

Apercu: Yes, they are no longer necessary.

chrisirhc:
Doesn't removal of the prepublish script mean that this library will no longer work with es5 projects?

Apercu:
@chrisirhc Hum what can cause it not to work with es5 projects? Sorry if I'm missing something

chrisirhc:
The prepublish script ensures that there the es6 code is transpiled into es5 (in dist/) so it can work with es5 projects. Unless you're using something else to ensure interoperability?

Apercu:
The name of the `prepublish` npm hook is a bit confusing since it make people think that it will only be run before the package is published, which is not the case since it's also executed upon install.

Removing it ensure that people installing this package using the `--production` flag will be able to use the library, since they won't be able to transpile the code, and improve install time. Imagine if every package did this rather than publishing builded files directly.

We only need to pay attention to build before publishing to avoid any mistake, but that's really the only con.

chrisirhc:
I don't think that's right. That prepublish running on install is only if you install the package through the git repository URL. If you install the package from npm (not a git URI), it doesn't run the prepublish script because it assumes the package from the npm repository is already in the published state.

chrisirhc:
See https://blog.greenkeeper.io/what-is-npm-s-prepublish-and-why-is-it-so-confusing-a948373e6be1

amilajack:
I think we we're looking for is the `version` script.

Apercu:
Thanks for making my 🌍  fall apart Chris, I was pretty certain it did and encountered the behavior a few times, must be my imagination.

chrisirhc:
👍 No problem 😄 . And this stuff may have changed in npm 4/5 anyways. I haven't read up on the latest developments.

Apercu:
Note that the current version of `gh-pages` is using the generated webpack bundles.

You only need to:

    npm run docs

And it will build, checkout, commit, push and go back to `master`.

chrisirhc:
@Apercu if you intend to work on #208, you need to assign it to yourself to avoid duplication of work. It's currently assigned to @amilajack .

Apercu:
@chrisirhc Done.

chrisirhc:
Also, please update this PR's description. What's the main goal of this PR? Once it's done, let's leave it at that. This PR is growing extremely large. What's the absolute necessary changes vs nice to haves?

tests/components.js
```diff
@@ -64,6 +61,8 @@ const TABLE_PROPS = {
   data: [[1, 2, 3], [4, 5, 6]]
 };
 
+const noop = f => f;
```
mcnuttandrew: prefer screaming snake case for constants

README.md
```diff
@@ -19,16 +39,14 @@ Some notable features:
 
 Install react-vis via npm.
 
-```shell
-npm install react-vis --save
-```
+    npm install react-vis --save
 
-Include the CSS from `./node_modules/react-vis/main` in your HTML page or via SASS:
+Include the builded main CSS file in your HTML page or via SASS:
```
mcnuttandrew: typo, *built

tests/components.js
Apercu: Ah okay

PR #233, opened 2017-01-13
Build examples separately with own dependencies

amilajack:
- [x] Build `main.js` docs with webpack
- [x] Fix build steps don't conflicts with new directory structure (recommendation by @Apercu)
- [x] Added support for `eslint-plugin-import` for better refactoring safety
- [x] Make sure examples work with styles, add webpack rule for sass
- [x] Ensure examples are `0.14.0` and `0.15.0` compatible
- [x] Add build step for each set of examples
- [x] Change directory structure (pluralize `example` dir, move examples and tests to root dir)


chrisirhc:
Is this blocked on @Apercu's #232 ? 

While we're making these build refractors/changes, let's all remember that we're trying to address improving the documentation and adoption.

I'm seeing a number of RFCs for changes but I don't see how they link to the overarching goal of improving docs and adoption.

Apercu:
@chrisirhc yes, but the two are conflicting with each other, so it's either #232 or this one.

amilajack:
@chrisirhc the RFC's I made are not related to the docs. They're just future things we can do after the examples.

Also I'm not sure if the `./dist/example`'s are necessary. Can someone please clarify what they are used for. After that, I can make the necessary finishing touches and this can be merged after reviewing.

chrisirhc:
Where are you seeing the dist/example ?

amilajack:
When running `npm run build-example`, there is a resulting `dist/example` dir that has compiled examples in it. I'm wondering if we should continue compiling the examples to the `./dist/example` directory. @Apercu removed those `build-*` scripts from his fork so i assume that they're not necessary anymore. Also `npm run build-example` seems to be intended for users. It makes more sense for the examples to not be included in `./dist` and instead in the root directory, where the user can clone the repo, cd to the example and run them.

Apercu:
The generation of `dist/example` was simply due to the fact the example dir was present inside the `src` folder.

Maybe I wasn't clear enough, but I don't think this PR should be merged, as what it brings has been implemented in #232.

I'll answer on the comment

amilajack:
Okay. I can close this and reopen another PR after your's gets merged.

chrisirhc:
@Apercu could you clarify what #232 is for ? I can't tell based on the its title and description.

chrisirhc:
From what I see, this PR intends to separate the each example into their own separate folders so that it's easier to tell how to set up an example from scratch and get something up and running.

We do need to get to that later but we should also look into making sure the documentation is showing up correctly, and with the current react-vis examples.

amilajack:
I made this PR based on https://github.com/uber/react-vis/issues/208#issuecomment-271417598 and https://github.com/uber/react-vis/issues/208#issuecomment-271474378. This PR is for the gh-pages update system.

chrisirhc:
Okay, that makes sense, it should be made clear in the description of this PR.
@Apercu does #232 address #208?

Apercu:
@chrisirhc There's multiple things, and making a PR for each change would have been a pain to review/create since they are interdependent.

There is a few things more that I need to do to completely address the docs issue, will probably do it tonight or tomorrow.

chrisirhc:
@Apercu how about reviewing this PR and mentioning what can be improved here? We should try to unblock @amilajack first, then incremental improvements can come subsequently. If we need quite a few changes to be done, we can work on a separate docs branch and merge on to it so that master branch stays in a working state.

@amilajack has done most of the footwork here for splitting examples out. I think there's still some work t to be done in this PR to clean it up.

Apercu:
@chrisirhc Because what can be improved has already be done in #232, re-applying @amilajack examples split on the `dist` branch would be the easiest and more straight-forward thing to do.

But I think we are doing something wrong here, since in fact, these `examples` are our interactive demo and way to debug all react-vis components. I believe we are trying to give two purposes to the thing, and that might end up bad.

What about creating a brand-new example directory rather than trying to modify this one? Also, if we want to keep things simple, why not it just be a directory with only one plot, so that if someone is interested, he can pickup all the needed configuration and inspire on the much more complete current example? It would make much more sense to me.

amilajack:
>  I think there's still some work to be done in this PR to clean it up.

@chrisirhc What kinds of things do you have in mind? I just rebased so merge conflicts were fixed.

chrisirhc:
Agree with @Apercu, let's go with creating one example directory first and keeping the current examples as is. Separating the rest of the examples can be done later once we're more certain about the docs generation. The key thing that isn't addressed in separating all the examples into their directory is the docs generation of each of the react-vis components and integrating the current markdown documentation to gh-pages. 

@amilajack With regards to clean up, one thing I saw in this PR is that the package.json for each example has many more dependencies than what I'd expect. Let's focus on getting one right first since this PR is huge. Then we can replicate it later. This separate directory approach is also more valuable for advanced examples which we haven't yet added to the repo. The current examples in the repo are quite simple.

amilajack:
> With regards to clean up, one thing I saw in this PR is that the package.json for each example has many more dependencies than what I'd expect

Are you referring to `devDependencies` or `dependencies`? I think we should be able to remove all the `dependencies` from each example's `package.json`. The `devDeps` are required for webpack, sass compilation, and live reloading.

chrisirhc:
> Are you referring to `devDependencies` or `dependencies`? I think we should be able to remove all the `dependencies` from each example's `package.json`. The `devDeps` are required for webpack, sass compilation, and live reloading.

Not all `dependencies`. The package.json 's dependencies should contain what's needed for the example to run. I would've expected `react-vis` to be a dependency. Something lightweight like what @ibgreen mentioned in https://github.com/uber/react-vis/issues/208#issuecomment-271474378: https://github.com/uber/react-map-gl/blob/dev/examples/main/package.json

amilajack:
Ahh right. Here's what I have in mind for the plot example's `package.json`:

```
{
  "description": "Plots example for react-vis",
  "license": "MIT",
  "scripts": {
    "start": "webpack-dev-server --progress --hot --port 3001 --open"
  },
  "dependencies": {
    "react-vis": "0.9.0"
  },
  "devDependencies": {
    "buble-loader": "^0.4.0",
    "css-loader": "^0.26.1",
    "extract-text-webpack-plugin": "2.0.0-beta.4",
    "node-sass": "^4.2.0",
    "sass-loader": "^4.1.1",
    "style-loader": "^0.13.1",
    "webpack": "^2.2.0-rc.3",
    "webpack-dev-server": "^2.2.0-rc.0"
  }
}
```

chrisirhc:
That package.json looks reasonable, though I believe there are other dependencies missing in there like `react`.

---

This thread is becoming quite long and diverging to multiple tangents. We want to be able to discuss the implementation of a particular example at length, and to be able to suggest improvements to it, and to iterate on it. Since this PR is extremely large and refactors many different examples, it's difficult to suggest improvements to one particular example (and you would need to refactor many examples for each suggestion).

As such, I suggest we start a new PR with one example, so we can iterate and suggest on it.

I'm thinking we should pick heaviest example to start with, and that looks to be the complex-chart example. What do you think?

amilajack:
> I'm thinking we should pick heaviest example to start with, and that looks to be the complex-chart example. What do you think?

Makes sense. I think we should borrow some of the ideas from this PR, including
* Moving `./src/example` -> `./examples` (helps avoid conflicts with babel compilation)
* Webpack config (allows live reloading and sass imports)

Apercu:
@amilajack Please stay focus on creating a new directory to feature just the complex-chart, without doing anything to the current `example` folder as it will be handled by #232. Yours can be called something like `example` or `demo`, plurialized if it's intended to contain multiple in the long run.

Issue #177, opened 2016-11-17
Deprecate Table?


mcnuttandrew:
There are lots better react tables out there in the ecosystem than the one in our package. It seems like it would probably be pretty reasonable to deprecate the table in favor of those. Would be delighted to hear other opinions

Apercu:
Yes, yes, yes. This component seems pretty out-of-scope of the project to me.

We should start by a simple warning for now, and remove it entirely with the next major release.

I can handle this if you'd like

mcnuttandrew:
Table is now deprecated!

PR #240, opened 2017-01-17
Start warning about the imminent depreciation of the table

mcnuttandrew:
This addresses #177 in part

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,8 +31,7 @@
     "d3-shape": "^1.0.4",
     "deep-equal": "^1.0.1",
     "global": "^4.3.1",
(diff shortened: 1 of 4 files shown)


Apercu:
Should we use the `warning` package for this? Or do we want to get rid of it?

mcnuttandrew:
i had no idea we even had that installed! Ill modify

Apercu:
I think it will always log like this, looking at their code tell me that they don't have the kind of variable to keep track of what's have already been logged. We might want to create a log util method using an object indexed by unique strings indicating if something has already been shown?

PR #244, opened 2017-01-19
Added inital complex-chart example

amilajack:
This PR was made as a follow up to https://github.com/uber/react-vis/pull/233, which was too large of a PR to be reviewed. This is an example of what other following examples should look like. Comments on the example in this PR will be reflected in examples that will be written later.

Here's a summary of what this PR adds for the complex-chart example:
- [x] Creates a `./examples` directory for where the current and future examples will be located
- [x] `babel-loader` integration. https://github.com/uber/react-vis/pull/233 used `buble-loader`. I was having module loader issues with this today so I used `babel-loader` as a (temporary?) replacement.
- [x] Aliased `react-vis` to `'../../dist'`. `babel-plugin-module-resolver` was preventing webpack from resolving the `index` alias created by `babel-plugin-module-resolver`
- [x] Created a `./examples/examples.scss`, borrowed from the original `./src/example/main.scss`
- [x] Borrowed `./examples/complex-chart/complex-chart-example.js` from `./showcase/plot/complex-chart.js`
- [x] Add `react-addons-shallow-compare` to `./examples/complex-chart/package.json`'s `dependencies`
- [x] Add `react` to `./examples/complex-chart/package.json`'s `dependencies`


examples/complex-chart/README.md
```diff
@@ -0,0 +1,10 @@
+react-vis legends
```
mcnuttandrew: why is this section titled legends?

examples/complex-chart/app.js
```diff
@@ -0,0 +1,66 @@
...
+
+// Cannot render to body anymore: react is throwing warnings. React.render()
+// was deprecated in 0.15.0
+//
+// The following code is boilerplate the make the examples work with React
+// 0.14.0 and 0.15.0. Do not use it in your app.
+//
+// When using >= 0.15.0, render with ReactDOM.render(). Otherwise, use React.render()
+
+const el = document.createElement('div');
```
mcnuttandrew: maybe a little cleaner to add the div ahead of time to the index.html page and access it via a className? not wedded to that solution though

examples/complex-chart/app.js
```diff
@@ -0,0 +1,66 @@
...
+
+const render = isReactDOMSupported() ?
```
Apercu: We don't need to support react 0.14, please use the `ReactDOM` render

examples/complex-chart/app.js
```diff
@@ -0,0 +1,66 @@
...
+import {isReactDOMSupported} from '../../src/utils/react-utils';
+
+import ComplexChartExample from './complex-chart-example';
+import '../examples.scss';
+import '../../src/main.scss';
+
+export default class App extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {width: window.innerWidth};
```
Apercu: Prefer the class property for of state, and add your listener in `componentDidMount`.

examples/complex-chart/app.js
```diff
@@ -0,0 +1,66 @@
...
+
+import ReactDOM from 'react-dom';
+import React, {Component} from 'react';
+import window from 'global/window';
+import document from 'global/document';
+import {isReactDOMSupported} from '../../src/utils/react-utils';
+
+import ComplexChartExample from './complex-chart-example';
+import '../examples.scss';
+import '../../src/main.scss';
```
Apercu: You could also make `examples.scss` includes the main one, like in the showcase.

examples/complex-chart/app.js
```diff
@@ -0,0 +1,66 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import ReactDOM from 'react-dom';
+import React, {Component} from 'react';
+import window from 'global/window';
+import document from 'global/document';
+import {isReactDOMSupported} from '../../src/utils/react-utils';
```
Apercu: The module resolver was specifically added to avoid such things. Also, you shouldn't require source files, but rather use `react-vis`.

examples/complex-chart/complex-chart-example.js
```diff
@@ -0,0 +1,199 @@
...
+  }
+  return result;
+}
+
+export default class Example extends React.Component {
+
+  constructor(props) {
+    super(props);
+    const totalValues = Math.random() * 50;
+    this.state = {
```
Apercu: ditto, class property

examples/complex-chart/complex-chart-example.js
```diff
@@ -0,0 +1,199 @@
...
+          data: getRandomSeriesData(totalValues)
+        },
+        {
+          title: 'Bananas',
+          disabled: false,
+          data: getRandomSeriesData(totalValues)
+        }
+      ]
+    };
+    this._nearestXHandler = this._nearestXHandler.bind(this);
```
Apercu: If you use class properties for your methods, you won't have to bind all of them in the constructor, much cleaner.

examples/complex-chart/complex-chart-example.js
```diff
@@ -0,0 +1,199 @@
...
+        <div className="legend">
+          <DiscreteColorLegend
+            onItemClick={this._legendClickHandler}
+            width={180}
+            items={series}/>
+        </div>
+
+        <div className="chart">
+          <FlexibleXYPlot
+            animation={true}
```
Apercu: Don't need `true`, just pass the animation prop

examples/complex-chart/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+  },
+
+  devtool: 'source-maps',
+
+  // NOTE: If you are using this config for your app, do not use this 'resolve'
+  //       section. This is example specific
+  resolve: {
+    alias: {
+      'react-vis': resolve('../../dist'),
```
Apercu: This should not be done, remove all aliases to parent directories.

examples/complex-chart/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+      react: resolve('../../node_modules/react'),
+      'react-dom': resolve('../../node_modules/react-dom')
+    }
+  },
+
+  module: {
+    rules: [
+      {
+        test: /\.(sass|scss)$/,
```
Apercu: We're only using `.scss` files

examples/examples.scss
```diff
@@ -0,0 +1,287 @@
+@import url(https://fonts.googleapis.com/css?family=Sintony:400,700);
```
Apercu: This import should be extracted and put as a stylesheet in `index.html`

amilajack:
Should the examples support configurations so that they're easily migratable or that they should work only for these examples? For example, in the `webpack.config.js`, I made sure that the configuration supports both `sass` and less files. Another example is supporting react 0.14.0 and 0.15.0. A lot of the comments that @Apercu made were related to making catering the configuration only towards the examples, which has its benefits. I tried to make the examples work out of the box to help save consumers of the examples from some js-fatigue.

Which direction should we take?

examples/complex-chart/app.js
amilajack: >  I tried patching into this PR and followed the instructions and got an error:

It requires that the `./dist` directory be built. When doing a fresh clone, this doesnt happen. Ideally, we could use webpack to alias `react-vis` => `./src` instead of `./dist` but I was having issues with `babel-plugin-module-resolver`.

examples/complex-chart/README.md
amilajack: This was a mistake. I'll fix and rebase :)

Apercu:
I don't think providing loaders that are not even used because we think some users (less? maybe 15%?) might need them is a good idea. The example should be something clonable and runnable easily, without overthinking what the users might want. Most of them will have their own configuration anyway, and if they don't, they'll take what we give to tweak it after.

You should depend on react 15 and react-dom.

Also, I have a feeling creating multiples packages/webpack configs might not be a good idea since there would be a lot of code duplication. We have to think of something that would work nicely

amilajack:
> We have to think of something that would work nicely

Yup, definitly. I'm not sure how we can make it cloneable without including one config per example. @mcnuttandrew Thoughts on this?

Apercu:
@amilajack You might want to look at what I've [did](https://github.com/uber/deck.gl/blob/master/exhibits/webpack/package.json#L6) on deck.gl to prevent this issue.

examples/complex-chart/webpack.config.js
amilajack: By removing the `react-vis` alias to `../../dist` dir, the examples will be built against the latest reactt-vis npm package instead of the source. Is that okay?

Apercu: Yes, we want something clonable and runnable on its own. The dist would not be there anyway until you install all the dependencies and run the build.

examples/complex-chart/complex-chart-example.js
amilajack: I think we should use autobound methods. Is that what you're referring to?

examples/complex-chart/complex-chart-example.js
amilajack: I see. Just to confirm, we want something like this:
```js
<FlexibleXYPlot
      animation
      {/* etc */} />
```
?

Apercu: Yes

examples/complex-chart/complex-chart-example.js
Apercu: If that's how you call class properties ;)

amilajack:
Rebased the PR based on reviews:
- [x] Symlinked `webpack.config.js` and `index.html` to parent avoid duplication
- [x] Autobound methods to avoid `*.method.bind()` boilerplate
- [x] Added missing `dependencies` (react, react-dom)
- [x] Removed unnecessary `constructor()` in `examples/complex-chart/app.js`
- [x] Imported `main.scss` from `examples.scss`
- [x] Refactored font retrieval from `examples.scss` to avoid request waterfall
- [x] Dropped react 0.14.0 support
- [x] Other minor/miscellaneous changes

amilajack:
Question: Should the `./src/main.scss` be symlinked to `./examples/main.scss`?

Apercu:
What's the behavior?

Don't think the link is necessary.

examples/complex-chart/complex-chart-example.js
Apercu: Class property instead?

examples/complex-chart/package.json
```diff
@@ -0,0 +1,28 @@
+{
```
Apercu: The packages and lockfiles are gonna be the same too

examples/webpack.config.js
```diff
@@ -0,0 +1,37 @@
...
+      warnings: false
+    }
+  },
+
+  devtool: 'source-maps',
+
+  module: {
+    rules: [
+      {
+        test: /\.scss/,
```
Apercu: Re-add `$` at the end

yarn.lock
```diff
@@ -1,5 +1,7 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Apercu: Shouldn't have been touched

examples/webpack.config.js
```diff
@@ -0,0 +1,37 @@
+// NOTE: This is a Webpack 2 configuration file for react-vis
```
Apercu: I think we know that ;)

examples/webpack.config.js
```diff
@@ -0,0 +1,37 @@
...
+const {resolve} = require('path');
```
Apercu: Shouldn't need the resolve, a string will be enough.

yarn.lock
amilajack: This changed when i installed the root dependencies. I will revert the changes made to this file.

examples/complex-chart/app.js
amilajack: Should we use `componentWillMount` instead to prevent possible layout trashing?

amilajack:
> What's the behavior?

@Apercu If the user does not install the root dependencies, the following error will be thrown:

```
ERROR in ./app.js
Module build failed: ReferenceError: Unknown plugin "module-resolver" specified in ...
```

Apercu:
You should to specify a babel config in the `package.json`, otherwise it will try to find it in parent directories.

amilajack:
@Apercu Ahh i see. That works. I was under the impression that babel-loader required the babel config to be inside of the webpack config itself. I tested adding the config to package.json and it works! I'll remove the babel config from the webpack config.

examples/examples.scss
```diff
@@ -0,0 +1,287 @@
+@import "../../src/main.scss";
```
mcnuttandrew: please scss-lint this file

amilajack: Is scss-lint a tool that react-vis uses? I can't seem to find the dependency or npm script for adding sass linting, so I assume that it requires some kind of global installation to run. I can PR for [stylelint](https://github.com/stylelint/stylelint) later. Would that be welcome?

mcnuttandrew: That's a wonderful idea! I would be really excited to see that PR. Style linting can wait for that pr in that case

amilajack:
Is this PR ready to be merged?

examples/complex-chart/app.js
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
Apercu: I'm not sure if we still need the header on every file, as @ibgreen removed it from the contribution guidelines, need confirmation

examples/complex-chart/complex-chart-example.js
```diff
@@ -0,0 +1,199 @@
...
+      y
+    });
+    lastY = y;
+  }
+  return result;
+}
+
+export default class Example extends React.Component {
+
+  constructor(props) {
```
Apercu: Remove the `constructor` and the `componentWillMount` and just use the state class property. [eg](https://github.com/uber/react-vis/blob/master/showcase/radial-chart/custom-radius-radial-chart.js#L27)

examples/complex-chart/app.js
amilajack: @mcnuttandrew does this need to be in every header file?

mcnuttandrew: I'm gonna error on the side of caution and say that yes it must be in every file

amilajack:
Made all the requested changes. Since @mcnuttandrew said that header should be in every file, is this ready to merge?

examples/complex-chart/complex-chart-example.js
```diff
@@ -0,0 +1,195 @@
...
+      data: getRandomSeriesData(this.totalValues)
+    },
+    {
+      title: 'Bananas',
+      disabled: false,
+      data: getRandomSeriesData(this.totalValues)
+    }
+  ]
+
+  state = {
```
Apercu:
```
state = {
  crosshairValues: [],
  series: [
    {
      title: 'Apples',
      disabled: false,
      data: getRandomSeriesData(Math.random() * 50)
    },
    {
      title: 'Bananas',
      disabled: false,
      data: getRandomSeriesData(Math.random() * 50)
    }
  ]
}
```

Apercu:
Just need to change what I commented, maybe just put the `Math.random() * 50` in a constant at the top of the file

amilajack:
Somehow I missed the comment on that. Finished the changes

Apercu:
I don't see the change

Issue #234, opened 2017-01-13
Question: Radial Chart data level override 


roderickhsiao:
From the custom radial chart example seems like we can pass data level override, 
for instance we tried to do on hover on section to have focus style (change `opacity`, `radius` and `innerRadius`), but the chart seemly doesnt work that way.

1. Opacity: 
As long as we has opacity in the data, every arcs render as opacity `0.1`
```js
const data = [
      {angle: 2, title: '1', opacity: 1},
      {angle: 6, title: '2', opacity: 1},
      {angle: 2, title: '3', opacity: 1},
      {angle: 3, title: '4', opacity: 1},
      {angle: 1, title: '5', opacity: 1}
    ];
]
```
==> all ars render opacity 0.1

If data looks like
```js
const data = [
      {angle: 2, title: '1', opacity: .5},
      {angle: 6, title: '2', opacity: .5},
      {angle: 2, title: '3', opacity: .5},
      {angle: 3, title: '4', opacity: .5},
      {angle: 1, title: '5', opacity: 1}
    ];
]
```
==> Only the one with `opacity:1` will be correct, other render as opacity `0.1`

2. Data level `innerRadius` and `radius` doesnt seem to work if we had component level props passed in.

mcnuttandrew:
Hey @roderickhsiao 

I just merged into a pr which i think should address the issue. I think what was causing it, was the automatic generation of domains. I'm planning on doing another minor later today once sankey etc lands, would you mind checking to see if this is resolved once that happens?

roderickhsiao:
sure thing thanks!

roderickhsiao:
@mcnuttandrew I tried the version 0.10.2,
the first case works, but the second one

```js
const data = [
      {angle: 2, title: '1', opacity: .5},
      {angle: 6, title: '2', opacity: .5},
      {angle: 2, title: '3', opacity: .5},
      {angle: 3, title: '4', opacity: .5},
      {angle: 1, title: '5', opacity: 1}
    ];
]
```
the rest still showing opacity: 0.1

roderickhsiao:
also tried changing radius and innerRadius on over section doesnt' seem to work

roderickhsiao:
Found one issue,

if the data looks like
```js
const data = [
      {angle: 2, title: '1', opacity: .5},
      {angle: 6, title: '2'},
      {angle: 2, title: '3'},
      {angle: 3, title: '4'},
      {angle: 1, title: '5'}
    ];
]
```

[...]

Issue #161, opened 2016-11-08
Ability to set an active data point


greatermeans:
I have a bar graph that looks like the below image. The current color scheme only allows for one color to be chosen for all the bars; whereas I would like to set a color for an active bar and another color for the rest. Please let me know if this is currently doable!

![image](https://cloud.githubusercontent.com/assets/17058760/20119176/7fcb14e2-a5d5-11e6-98b9-34e3d31b94ec.png)

mcnuttandrew:
Hey @greatermeans I've generally found myself doing this by wrapping the chart in a stateful component that has a notion of "selected bar", using colorType='literal' for the color scale, and then modifying the data

greatermeans:
@mcnuttandrew thanks for the response, do you have an example by any chance?

mcnuttandrew:
I have one in an upcoming PR, something like this https://github.com/uber/react-vis/blob/mouse-out-radial/src/example/radial-chart/custom-radius-radial-chart.js but for barchart?

greatermeans:
@mcnuttandrew awesome dude, was able to replicate!

mcnuttandrew:
Dope! It's generally my opinion that sort of state maintaince is the job of the wrapping application, so i'm gonna close this issue as a wont fix? Feel free to re-open if you disagree

greatermeans:
@mcnuttandrew fine to close. This worked on stackedbarchart and barchart. For some reason, LineMarkSeries doesn't respond to opacity. If you have any suggestions, please let me know!

Issue #157, opened 2016-11-01
Radial chart mouse out does not get called


mcnuttandrew:

mcnuttandrew:
Specifically the occurs when animation is active

mcnuttandrew:
This is now resolved

PR #245, opened 2017-01-19
Improve Radial Chart Functionality

mcnuttandrew:
This PR addresses a number of bugs that have been present in the radial chart for a while. Chief among them are: 

- When specify a variable such as opacity that happen to all have the same value, the value default to a small value. This problem is documented in #234 

- Radial chart not updating when new props are passed. This issue was briefly referenced in #161 but didn't have a direct gh-issue about it, so I added a example that demonstrates how to build a hover, as well demonstrates the bug. 

- Untangling the mouse events in radial (#157)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ node_modules/
 .nyc_output/
 coverage/
 dist/
-showcase/node_modules/
 
 # To be removed in next major
 main.css
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ node_modules/
 .nyc_output/
 coverage/
 dist/
+showcase/node_modules/
 
 # To be removed in next major
 main.css
--- a/showcase/radial-chart/custom-radius-radial-chart.js
+++ b/showcase/radial-chart/custom-radius-radial-chart.js
@@ -18,24 +18,21 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import React from 'react';
+import React, {Component} from 'react';
 
 import RadialChart from 'radial-chart';
 
-export default class SimpleRadialChart extends React.Component {
-  constructor(props) {
-    super(props);
+export default class SimpleRadialChart extends Component {
 
-    this.state = {
-      hoveredSection: false
-    };
+  state = {
+    hoveredSection: false
   }
 
   render() {
     const {hoveredSection} = this.state;
     return (
       <RadialChart
-        animation={true}
+        animation
         radiusDomain={[0, 20]}
         data={[
           {
--- a/showcase/radial-chart/custom-radius-radial-chart.js
+++ b/showcase/radial-chart/custom-radius-radial-chart.js
@@ -23,17 +23,54 @@ import React from 'react';
 import RadialChart from 'radial-chart';
 
 export default class SimpleRadialChart extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      hoveredSection: false
+    };
+  }
+
   render() {
+    const {hoveredSection} = this.state;
     return (
       <RadialChart
+        animation={true}
         radiusDomain={[0, 20]}
         data={[
-          {angle: 1, radius: 10},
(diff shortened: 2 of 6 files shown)


showcase/radial-chart/custom-radius-radial-chart.js
```diff
@@ -23,17 +23,54 @@ import React from 'react';
 import RadialChart from 'radial-chart';
 
 export default class SimpleRadialChart extends React.Component {
+  constructor(props) {
```
Apercu: Prefer using the `state` class property.

showcase/radial-chart/custom-radius-radial-chart.js
```diff
@@ -23,17 +23,54 @@ import React from 'react';
...
+    this.state = {
+      hoveredSection: false
+    };
+  }
+
   render() {
+    const {hoveredSection} = this.state;
     return (
       <RadialChart
+        animation={true}
```
Apercu: `true` is not necessary if the animation property is present

src/radial-chart/index.js
```diff
@@ -116,7 +109,7 @@ class RadialChart extends React.Component {
    * @param {Object} event Event.
    * @private
    */
-  _triggerSectionHandler(handler, d, event) {
+  _sectionHandler(handler, d, event) {
```
Apercu: Same thing, don't need to check for the handler presence

src/radial-chart/index.js
```diff
@@ -270,18 +237,28 @@ class RadialChart extends React.Component {
...
-                opacity: opacityFunctor && opacityFunctor(d),
-                stroke: strokeFunctor && strokeFunctor(d),
-                fill: fillFunctor && fillFunctor(d)
-              },
-              onMouseOver: e => this._sectionMouseOver(pieData[i], e),
-              onMouseOut: e => this._sectionMouseOut(pieData[i], e),
-              onClick: e => this._sectionClick(pieData[i], e),
-              key: i
-            }}/>)}
+            {data.map((d, i) => {
```
Apercu: I was enjoying the direct return of the jsx element in the map

src/radial-chart/index.js
```diff
@@ -270,18 +237,28 @@ class RadialChart extends React.Component {
...
+              return (
+                <path {...{
+                  d: arc(pieData[i]),
+                  style: {
+                    opacity: opacityFunctor && opacityFunctor(d),
+                    stroke: strokeFunctor && strokeFunctor(d),
+                    fill: fillFunctor && fillFunctor(d)
+                  },
+                  onMouseEnter: e => {
```
Apercu: This could probably be kept on one line instead of 3

mcnuttandrew: true, but ill have to extend the max line width. truth be told i'd been waiting for an excuse to do so, so here we go

showcase/radial-chart/custom-radius-radial-chart.js
mcnuttandrew: i'm not sure i know what you are referring to? can you point me to an example?

src/radial-chart/index.js
mcnuttandrew: can you explain this comment a little more?

src/radial-chart/index.js
Apercu: haha 👍

showcase/radial-chart/custom-radius-radial-chart.js
Apercu:
```
class SimpleRadialChart extends Component {

  state = {
    hoveredSection: false
  }

}

File: src/radial-chart/index.js (hunk 4 above)
Comment (Apercu) [inline src/radial-chart/index.js]: `data.map((d, i)) => <path>`

File: .gitignore [hunk 6]
```diff
@@ -3,6 +3,7 @@ node_modules/
 .nyc_output/
 coverage/
 dist/
+showcase/node_modules/
```
Apercu: Isn't `node_modules/` already covering this one? That's louche!

mcnuttandrew: whoops didn't mean to leave this in, i was flailing around trying to make atom not search showcase/nodemodules

Issue #174, opened 2016-11-17
Add event listeners for tree map


mcnuttandrew:
The tree map should have onhover and on click callbacks to match the api of the other chart types

PR #246, opened 2017-01-19
Add interaction listeners for treemap

mcnuttandrew:
This PR addresses the feature enhancement from #174, open to other apis, but I feel like this one matches the the other charts reasonable well while providing access to the full node tree.

### DIFF
--- a/showcase/treemap/dynamic-treemap.js
+++ b/showcase/treemap/dynamic-treemap.js
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 import React from 'react';
-import timers from 'timers';
+import window from 'global/window';
 
 import Treemap from 'treemap';
 
@@ -34,7 +34,7 @@ export default class DynamicTreemapExample extends React.Component {
   }
 
   componentDidMount() {
-    timers.setInterval(
+    window.setInterval(
       () => this.setState({treemapData: this._getRandomData()}),
       5000
     );
--- a/showcase/treemap/dynamic-treemap.js
+++ b/showcase/treemap/dynamic-treemap.js
@@ -25,12 +25,9 @@ import Treemap from 'treemap';
 
 export default class DynamicTreemapExample extends React.Component {
 
-  constructor(props) {
-    super(props);
-    this.state = {
-      hoveredNode: false,
-      treemapData: this._getRandomData()
-    };
+  state = {
+    hoveredNode: false,
+    treemapData: this._getRandomData()
   }
 
   componentDidMount() {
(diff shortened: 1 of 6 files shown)


src/treemap/index.js
```diff
@@ -132,10 +135,24 @@ class Treemap extends React.Component {
...
   }
 
+  /**
+   * Triggers a callback on a box if the callback is set.
+   * @param {function} handler Callback function.
+   * @param {Object} d Data point of the arc.
+   * @param {Object} event Event.
+   * @private
+   */
+  _triggerHandler(handler, d, event) {
```
Apercu: Why not setting the default value of the handlers to noop, this way you don't need to do this and the code would be cleaner

showcase/treemap/dynamic-treemap.js
```diff
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 import React from 'react';
-import window from 'global/window';
+import timers from 'timers';
```
Apercu: Didn't knew this one, thanks. Don't it need to be shimmed on the client like `fs`?

src/treemap/index.js
```diff
@@ -132,10 +137,22 @@ class Treemap extends React.Component {
...
   }
 
+  /**
+   * Triggers a callback on a box if the callback is set.
+   * @param {function} handler Callback function.
+   * @param {Object} d Data point of the arc.
+   * @param {Object} event Event.
+   * @private
+   */
+  _triggerHandler(handler = NOOP, d, event) {
```
Apercu: I was more thinking about using the default props so you don't need to create a new method

showcase/treemap/dynamic-treemap.js
```diff
@@ -28,6 +28,7 @@ export default class DynamicTreemapExample extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
+      hoveredNode: false,
```
Apercu: Class prop too?

src/treemap/index.js
mcnuttandrew: oh gotcha

src/treemap/index.js
```diff
@@ -20,12 +20,12 @@
...
 import React, {PropTypes} from 'react';
 import * as d3Hierarchy from 'd3-hierarchy';
-import * as d3Color from 'd3-color';
 
 import Animation from 'animation';
 import {CONTINUOUS_COLOR_RANGE, DEFAULT_COLOR, OPACITY_RANGE} from 'theme';
 import {AnimationPropType} from 'utils/animation-utils';
-import {getAttributeFunctor, getMissingScaleProps} from 'utils/scales-utils';
+import {getAttributeFunctor, getFontColorFromBackground, getMissingScaleProps} from 'utils/scales-utils';
+import {} from 'utils/chart-utils';
```
Apercu: ?

PR #247, opened 2017-01-19
allow custom crosshair orientation

fastfrwrd:
if the user has custom requirements for crosshair positioning, they should be able to provide an override prop for the "left" or "right" value of `<Crosshair />`.

### DIFF
--- a/src/plot/crosshair.js
+++ b/src/plot/crosshair.js
@@ -70,6 +70,7 @@ class Crosshair extends PureRenderComponent {
       innerHeight: React.PropTypes.number,
       marginLeft: React.PropTypes.number,
       marginTop: React.PropTypes.number,
+      orientation: React.PropTypes.oneOf(['left', 'right']),
       itemsFormat: React.PropTypes.func,
       titleFormat: React.PropTypes.func
     };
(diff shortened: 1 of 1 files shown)


src/plot/crosshair.js
```diff
@@ -132,14 +133,19 @@ class Crosshair extends PureRenderComponent {
...
     if (!value) {
       return null;
     }
     const x = getAttributeFunctor(this.props, 'x');
     const innerLeft = x(value);
 
-    const orientation = (innerLeft > innerWidth / 2) ? 'left' : 'right';
+    // infer orientation from x value if not overridden
+    if (!orientation) {
+      orientation = (innerLeft > innerWidth / 2) ? 'left' : 'right';
```
Apercu: Can't you do this stuff in `getDefaultProps`? Not a fan of overriding variables ;)

fastfrwrd: no, since `getDefaultProps` shouldn't access `this.props`. If this http://stackoverflow.com/a/33127059/1214462

I could create an instance method for this called `getOrientation()` that either returns the prop or the inference, which would prevent reassignment in `render()`.

fastfrwrd: Biggest problem with that is that I'll have to do lines 140 and 141 in `getOrientation()`.

Apercu: Could this work in the render?

```
const {
      children,
      values,
      marginTop,
      marginLeft,
      innerWidth,
      innerHeight
} = this.props;

const {orientation = (innerLeft > innerWidth / 2) ? 'left' : 'right'} = this.props;
```

fastfrwrd: I like it, updating.

Issue #39, opened 2016-03-22
Add the ability to show labels on the chart (e.g. label series)


bulyonov:

mikaa123:
Hello!
Just to make sure: would this feature allow to show labels on `RadialChart`?
Are there currently any way to display labels on it?

Thanks

bulyonov:
@mikaa123 , currently there's no way to place labels on radial chart. We will add them as soon as the label series will be supported.

I also realized that we don't support hints for the radial as well. I created a task for them here — #46.

stolendog:
it would be great to have text labels on `RadialChart`, waiting for this too.

ranjithnori:
+1

mcnuttandrew:
The labelSeries is now available on the 2.0.0beta release, this ticket can be closed.

PR #249, opened 2017-01-20
Add labels to radial plot

mcnuttandrew:
This PR addresses the omnipresent need to add labels to the radial plot (#39). To do so we modify the api of the radial plot in a couple of ways. We add a top level prop called showLabels to engage the labels, and two new data props, label and subLabel (See below screencap)

screen shot 2017-01-20 at 8 36 49 am

I think the long term solution for labels & radial plots will be to develop a labelSeries that accepts regular coordinates, and then use that in conjunction with the r-theta plot. In the mean time, this solution makes sense to me as a way to make our pie chart more ethical (label-less pie charts aint great).

### DIFF
--- a/src/radial-chart/index.js
+++ b/src/radial-chart/index.js
@@ -192,13 +192,102 @@ class RadialChart extends React.Component {
       .innerRadius(innerRadiusFunctor);
   }
 
+  /**
+   * Generate the svg pie slices to be rendered
+   * @param {Array} pieData - the d3 generate information for drawing slices
+   * @param {Func} arc - the arc generator
+   * @returns {function} the react content functor
+   * @private
+   */
+  _renderPieSlice(pieData, arc) {
+    const opacityFunctor = this._getAttributeFunctor('opacity');
+    const fillFunctor = this._getAttributeFunctor('fill') || this._getAttributeFunctor('color');
+    const strokeFunctor = this._getAttributeFunctor('stroke') || this._getAttributeFunctor('color');
+    return (d, i) => {
+      return (<path {...{
+        className: 'rv-radial-chart__series--pie__slice',
+        d: arc(pieData[i]),
+        style: {
+          opacity: opacityFunctor && opacityFunctor(d),
+          stroke: strokeFunctor && strokeFunctor(d),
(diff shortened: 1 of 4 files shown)


src/radial-chart/index.js
```diff
@@ -235,19 +238,64 @@ class RadialChart extends React.Component {
...
+                  style: {
+                    opacity: opacityFunctor && opacityFunctor(d),
+                    stroke: strokeFunctor && strokeFunctor(d),
+                    fill: fillFunctor && fillFunctor(d)
+                  },
+                  key: i
+                }}/>)}
+            </g>
+            <g className="rv-radial-chart__series--pie-labels-wrapper">
+              {showLabels && data.map((d, i) => {
```
Apercu: Might be a bit more readable to declare a method for this one

src/radial-chart/index.js
```diff
@@ -235,19 +238,64 @@ class RadialChart extends React.Component {
...
+                // reject the label if its not affixed to the section
+                const canRenderMainLabel = d.label && typeof d.label === 'string';
+                const canRenderSubLabel = d.subLabel && typeof d.subLabel === 'string';
+                if (!canRenderMainLabel && !canRenderSubLabel) {
+                  return;
+                }
+                const angle = (pieData[i].startAngle + pieData[i].endAngle) / 2 - Math.PI / 2;
```
javidhsueh: It would be better
const angle = (pieData[i].startAngle + pieData[i].endAngle) / 2 - (Math.PI / 2);
also explain what does this equation do.

src/radial-chart/index.js
```diff
@@ -235,19 +238,64 @@ class RadialChart extends React.Component {
...
+                const xTrans = 1.1 * radiusFunctor(d) * Math.cos(angle);
+                const yTrans = 1.1 * radiusFunctor(d) * Math.sin(angle);
+                const textAnchor = (angle > 0.5 * Math.PI) && angle < (1.5 * Math.PI) ? 'end' : 'start';
```
javidhsueh: maybe add more comment how the angle affect the text anchor.

PR #255, opened 2017-01-24
Added stylelint and stylefmt support

amilajack:
- [x] Added stylelint and stylefmt support
- [x] Conformed all styles to convention


.stylelintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "stylelint-config-standard",
+	"rules": {
+		"color-hex-case": "upper",
```
mcnuttandrew: i kinda prefer lower cased hexes personally

.stylelintrc
```diff
@@ -0,0 +1,7 @@
...
+    "selector-list-comma-newline-after": "never-multi-line"
```
mcnuttandrew: also i prefer multi-lining multi-selectors

package.json
```diff
@@ -23,6 +23,8 @@
     "clean": "rm -rf dist bundle.* index.html && mkdir dist",
     "build": "npm run clean && babel src -d dist --copy-files && node-sass src/main.scss main.css",
     "lint": "eslint src tests showcase --ignore-pattern node_modules --ignore-pattern bundle.js",
+    "lint-styles": "stylelint examples/*.scss showcase/*.scss src/*.scss src/styles/*.scss --syntax scss",
+    "lint-styles-fix": "stylefmt -r examples/*.scss showcase/*.scss src/*.scss src/styles/*.scss --syntax scss",
```
mcnuttandrew: i'm not sure i understand what this does?

examples/examples.scss
```diff
@@ -7,7 +7,7 @@ body {
   font-size: 14px;
 }
 
-h1, h2, h3, h4, h5 {
+h1,h2,h3,h4,h5 {
```
Apercu: What's the purpose of this one? Minifying our source?

.stylelintrc
amilajack: The convention in the project was mixed so i wasnt sure what to choose. I'll fix this

.stylelintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "stylelint-config-standard",
+	"rules": {
```
Apercu: Can you format this config instead of keeping mixed tab and spaces?

package.json
Apercu: Kinda the same as `eslint --fix`

Apercu: Lot of repetition here, can't we avoid that? Also unsure if we need to add `stylefmt`, it's useful the first time, but not so much after.

package.json
```diff
@@ -61,6 +63,9 @@
     "react-addons-shallow-compare": ">=15.4.2",
     "react-addons-test-utils": ">=15.4.2",
     "react-dom": ">=15.4.2",
+    "stylefmt": "^5.1.1",
+    "stylelint": "^7.7.1",
+    "stylelint-config-standard": "^15.0.1",
```
Apercu: Note: there is no such thing as standard in the js world 😁

.stylelintrc
Apercu: Same

examples/examples.scss
```diff
@@ -57,16 +57,19 @@ nav {
   float: right;
   padding: 0;
   font-weight: 400;
+
```
Apercu: Do we want this @mcnuttandrew ?

src/styles/plot.scss
```diff
@@ -1,36 +1,28 @@
...
-  color: #c3c3c3;
+  color: #C3C3C3;
   position: relative;
 }
 
 .rv-xy-plot__inner {
   display: block;
 }
 
-.rv-xy-plot__axis {
```
Apercu: I have a feeling this was here as a reminder

showcase/index.scss
```diff
@@ -1,4 +1,4 @@
-@import "../main";
+@import "../src/main.scss";
```
Apercu: Should keep this as it is

Apercu: Hum actually not sure now, maybe it's fine

Apercu:
The showcase stylesheet and example one are exactly the same, we should probably want to merge them.

amilajack:
What about creating a `examples.scss` file in the root and removing the existing `./showcase/index.scss` and `./examples/examples.scss`?

Apercu:
Not a fan of throwing too much files in the root, it should be clean. `src/styles` would be a better place.

amilajack:
Removed duplication by creating a single `./src/styles/examples.scss`

Apercu:
Nice! ;) And could you remove the two rule overrides?

amilajack:
Which rules are you referring to?

amilajack:
I added the second rule because of @mcnuttandrew 's [comment](https://github.com/uber/react-vis/pull/255#discussion_r97651011)

Apercu:
My bad, keep the multi-line, but remove the uppercase

src/styles/plot.scss
```diff
@@ -1,36 +1,30 @@
...
-$rv-xy-plot-axis-line-color: #e6e6e9;
+$rv-xy-plot-axis-line-color: #E6E6E9;
 $rv-xy-plot-axis-font-size: 11px;
-
 $rv-xy-plot-tooltip-background: #3A3A48;
-$rv-xy-plot-tooltip-color: #fff;
+$rv-xy-plot-tooltip-color: #FFF;
 $rv-xy-plot-tooltip-font-size: 12px;
 $rv-xy-plot-tooltip-border-radius: 4px;
 $rv-xy-plot-tooltip-shadow: 0 2px 4px rgba(0, 0, 0, 0.5);
```
mcnuttandrew: while were linting stuff, prefer rbga with hexes: rgba(#000, 0.5) would be better here

amilajack: Should this be done for all hexes or just this one?

package.json
```diff
@@ -23,6 +23,7 @@
     "clean": "rm -rf dist bundle.* index.html && mkdir dist",
     "build": "npm run clean && babel src -d dist --copy-files && node-sass src/main.scss main.css",
     "lint": "eslint src tests showcase --ignore-pattern node_modules --ignore-pattern bundle.js",
+    "lint-styles": "stylelint src/*.scss src/styles/*.scss --syntax scss",
```
Apercu: We should probably be fine with only `stylelint src/styles/*.scss --syntax scss`

amilajack: Fixed

Apercu:
Nice 😃

PR #259, opened 2017-01-27
Initial gitbook working prototype

amilajack:
- [x] Uses the existing `./docs` examples as content for gitbook


amilajack:
screen shot 2017-01-26 at 8 38 00 pm

Setup script:
```
git clone https://github.com/amilajack/react-vis.git
cd react-vis
git checkout gitbook-support
npm i -g gitbook-cli
gitbook install
gitbook serve
```

hanjunx:
sgtm

mcnuttandrew:
This seems fine so far, can confirm that it works! Pretty neat

mcnuttandrew:
What specifically is the plan for making the gitbook integrate the examples/ folder?

amilajack:
We can include the examples files with gitbook [with this plugin](https://github.com/azu/gitbook-plugin-include-codeblock).

mcnuttandrew:
Maybe i don't get that plug-in, but I'm not sure that it will allow us to import the examples folder as it stands? Like it looks kinda like it's expecting handlebars?

amilajack:
Yea, its similar to handlebars in how it allows certain files to be included into readme files. I was thinking that we could use it to import the `./examples/**/*.js` files into the cookbook style documentation.

hanjunx:
sgtm

Hanjun

package.json
```diff
@@ -19,6 +19,7 @@
   },
   "scripts": {
     "docs": "(cd showcase && npm run build && cp bundle.* .. && cd .. && git checkout gh-pages && git commit -am 'Upgrade docs' && git push && git checkout master)",
```
mcnuttandrew: i guess my other question is how with the gitbook interface with the doc script? we need to make sure that this pipeline is easy to run

hanjunx:
lgtm

so lazy, can't move

talk is cheap, show me the code

stay hungry, hunting for souls and love

stay foolish, learning to kill and hate

Hanjun

hanjunx:
it is what it is
i am who i am
slow to anger
quick to turnaround

Hanjun

hanjunx:
Close other issues. I want a clean interface.

Hanjun

package.json
amilajack: I see. Maybe we can run the `gitbook-setup` script in the `docs` script?

amilajack: ?

mcnuttandrew:
I spent sometime over the weekend looking at the plug-in you mention above, and it's not immediately clear to me if it will work for our purposes. I worry that our web-packing solution will not play nice with something that attempts to inject an iframe. (also i think you mean this plug in https://github.com/zingchart/gitbook-plugin-include-display-run, the one you link to would definitely not play nice with our examples). I'm still pretty hesitant about the long term viability of gitbook, especially if it means needing to code up our plugin in order to work with webpack

amilajack:
I'm not sure why webpack is required for the examples. For the gitbook examples, i was thinking that we could import code as txt files and not actually run the examples because we're not aiming for live and editable exampels

mcnuttandrew:
We are definately aiming to be able have our example be live. I don't think we really need them to be editable, but think we really ought to be able to have live versions of our stuff from the /examples folder

amilajack:
I see. But even to have live examples, I don't think the live editor needs to know about our webpack config. esnextb.in is a good example of this. I was looking at [FormidableLabs/component-playground](https://github.com/FormidableLabs/component-playground) over the weekend. Seems pretty nice because live examples can be added adding a few scripts.

mcnuttandrew:
Now that we have the homespun docs-app living in the repo, I'm gonna close this as we no longer need a gitbook

PR #260, opened 2017-01-31
Initial docs 'getting started guide'

amilajack:
Here's an initial demo of the getting started guide. Although its not finished, I would like some early reviews so that I can get an idea for that the rest of the docs should look like


mcnuttandrew:
Looks fine

amilajack:
This actually 100% finished. I made the PR so i could get a review. I'll make a PR that adds some finishing touches to this PR

Issue #236, opened 2017-01-14
demo plot on your github page has a broken plot


Frazer:
on http://uber.github.io/react-vis/  there is a plot that displays really badly:

![screen shot 2017-01-14 at 10 16 42](https://cloud.githubusercontent.com/assets/374464/21954082/f6e5f014-da47-11e6-80d3-cbba04ec5ffa.png)

it appears like that for me on safari and firefox.  

Also the x-axis is displaying decimals (including negative decimals) whenever there are less than 15 columns, which isn't sensible for a bar chart.

![screen shot 2017-01-14 at 11 00 07](https://cloud.githubusercontent.com/assets/374464/21954111/c357bf60-da48-11e6-92c7-4289e1edc84c.png)

mcnuttandrew:
Alt: maybe I'm misunderstanding the basis of your comment, can you provide a gif?

Apercu:
You can see on his first screenshot that the y-axis is entirely hidden, you can see the top of a value just at the bottom of the word plot. To me it's more an issue about data (and the random generation process) rather than the library itself.

Frazer:
I guess the problem with the decimals on the x-axis is that you will have X
discrete columns, but 2X or 3X ticks on the x-axis.

amilajack:
I was thinking of limiting the bar count to >= 2. Having a single bar somewhat defeats the purpose of having a bar chart, which really only makes sense when comparing two or more things.

mcnuttandrew:
@amilajack that's reasonable, i'd also endorse using the rect-series rather than the bar-series, as this demo is using a continuous variable for the axis, and is thus a histogram (and not a bar chart)

mcnuttandrew:
Hey @Frazer,

The demo page has been pretty radically updated! I think the issue should now be resolved. I'm marking this as resolved, but if you see the demo acting in a way that doesn't make sense again, feel free to re-open

PR #276, opened 2017-02-20
Enforce a minimum number of boxes in the complex chart

mcnuttandrew:
When there are less than two bars in the complex chart it looks pretty crumby! This is a simple work around. It addresses #236

### DIFF
--- a/showcase/plot/complex-chart.js
+++ b/showcase/plot/complex-chart.js
@@ -174,8 +174,16 @@ export default class Example extends React.Component {
             xDomain={[0, series[0].data.length - 1]}
             height={300}>
             <HorizontalGridLines />
-            <YAxis className="cool-custom-name"/>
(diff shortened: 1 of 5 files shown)


showcase/plot/complex-chart.js
```diff
@@ -171,12 +171,16 @@ export default class Example extends React.Component {
...
             onMouseLeave={this._mouseLeaveHandler}
+            xDomain={[0, series[0].data.length - 1]}
             height={300}>
             <HorizontalGridLines />
             <YAxis className="cool-custom-name"/>
             <XAxis className="even-cooler-custom-name"/>
-            <VerticalBarSeries
-              data={series[0].data}
+            <VerticalRectSeries
+              data={series[0].data.map(row => {
```
Apercu: `series[0].data.map(({x, y}) => ({x0: x - 0.5, x + 0.5, y})`

PR #277, opened 2017-02-20
tickSizeInner should be tickSizeOuter and vice versa.  

alee8:
The plot is the reference point. The old version had it inverted. 
Updated the complex-chart example to show that the changes work.
Attached image shows the example with outTicks of size 8.
pr179

### DIFF
--- a/showcase/plot/complex-chart.js
+++ b/showcase/plot/complex-chart.js
@@ -173,8 +173,16 @@ export default class Example extends React.Component {
             onMouseLeave={this._mouseLeaveHandler}
             height={300}>
             <HorizontalGridLines />
-            <YAxis className="cool-custom-name"/>
-            <XAxis className="even-cooler-custom-name"/>
+            <YAxis
+              className="cool-custom-name"
+              tickSizeInner={ 0 }
+              tickSizeOuter={ 8 }
+            />
+            <XAxis
+              className="even-cooler-custom-name"
+              tickSizeInner={ 0 }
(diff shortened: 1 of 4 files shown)


src/styles/examples.scss
```diff
@@ -158,6 +158,9 @@ article {
 
   .chart {
     margin-right: 200px;
+    & .rv-xy-plot__axis__tick__line {
+      stroke: #6b6b76;
```
mcnuttandrew: nits: 
- i dont think you need the &
- can you use a scss variable instead for the hex value?

mcnuttandrew:
Can you also add some notes in the docs about this?

showcase/plot/complex-chart.js
```diff
@@ -173,8 +173,16 @@ export default class Example extends React.Component {
             onMouseLeave={this._mouseLeaveHandler}
             height={300}>
             <HorizontalGridLines />
-            <YAxis className="cool-custom-name"/>
-            <XAxis className="even-cooler-custom-name"/>
+            <YAxis
+              className="cool-custom-name"
+              tickSizeInner={ 0 }
```
Apercu: Could you please remove the spacing between the brackets?

alee8: too late.  You don't like the clarity this gives?

Apercu: Fixed on master. Consistency is my issue

PR #289, opened 2017-03-06
Modified `AxisTitle` to use attribute transform instead of css transform

### LINKED ISSUE #286: [AxisTitle] Axis titles in Microsoft Edge are incorrectly positioned
[Currently CSS transforms are not supported on SVG elements by Microsoft Edge](https://docs.microsoft.com/en-us/microsoft-edge/dev-guide/graphics/svg) which causes an issue with the positioning of the `AxisTitle` component.

The fix I'm suggesting is to take the transform out of the [`style` attribute](https://github.com/uber/react-vis/blob/master/src/plot/axis/axis-title.js#L62) and instead use a `tranform` attribute directly on that group.

Will whip up a quick PR if you're happy with this approach!

samhogg:
Fixes #286

### DIFF
--- a/src/plot/axis/axis-title.js
+++ b/src/plot/axis/axis-title.js
@@ -22,6 +22,8 @@ import React from 'react';
 
 import {ORIENTATION} from 'utils/axis-utils';
 
+// Assuming that 16px = 1em
+const ADJUSTMENT_FOR_TEXT_SIZE = 16;
 const {LEFT, RIGHT, TOP, BOTTOM} = ORIENTATION;
 
 /**
@@ -32,20 +34,20 @@ const {LEFT, RIGHT, TOP, BOTTOM} = ORIENTATION;
  */
 const transformation = (width, height) => ({
   [LEFT]: {
-    x: 16,
+    x: ADJUSTMENT_FOR_TEXT_SIZE,
     y: 0,
     rotation: -90,
     textAnchor: 'end'
   },
   [RIGHT]: {
-    x: -8,
+    x: ADJUSTMENT_FOR_TEXT_SIZE * -0.5,
     y: height,
     rotation: -90,
     textAnchor: 'start'
   },
   [TOP]: {
     x: 0,
-    y: 16,
+    y: ADJUSTMENT_FOR_TEXT_SIZE,
     rotation: 0,
     textAnchor: 'start'
   },
--- a/src/plot/axis/axis-title.js
+++ b/src/plot/axis/axis-title.js
@@ -23,29 +23,36 @@ import React from 'react';
 import {ORIENTATION} from 'utils/axis-utils';
 
 const {LEFT, RIGHT, TOP, BOTTOM} = ORIENTATION;
+
+/**
+ * Compute transformations, keyed by orientation
+ * @param {number} width - width of axis
+ * @param {number} height - height of axis
+ * @returns {Object} Object of transformations, keyed by orientation
+ */
 const transformation = (width, height) => ({
   [LEFT]: {
     x: 16,
     y: 0,
-    rotation: '-90',
+    rotation: -90,
     textAnchor: 'end'
   },
   [RIGHT]: {
     x: -8,
     y: height,
-    rotation: '-90',
+    rotation: -90,
     textAnchor: 'start'
   },
   [TOP]: {
     x: 0,
-    y: 12,
-    rotation: '0',
+    y: 16,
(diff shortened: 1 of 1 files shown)


mcnuttandrew:
Hey @samhogg,

You changes are super reasonable! That said, I think the axis-title component is kind of a mess, would you mind if I redid this PR to bring it into a more modern style (in addition to addressing your changes)?

samhogg:
@mcnuttandrew yeah I was thinking of a bigger refactor but just wanted to solve the issue at hand - no problems from my end if you want to do that!

src/plot/axis/axis-title.js
```diff
@@ -36,30 +36,23 @@ const propTypes = {
...
-      transform: `translate(${width}px, -6px)`,
-      textAnchor: 'end'
-    };
+    transform = `translate(${width}px, -6px)`;
+    textAnchor = 'end';
   } else {
-    style = {
-      transform: 'translate(0px, 1em)'
-    };
+    transform = 'translate(0px, 1em)';
```
mcnuttandrew: does this work on edge? my understanding of svg transforms was that was done in unitless values? I'm fiddling around with my own version and this technique throws a lot of errors

samhogg: Oops, good catch. I had it unitless, obviously forgot to save or something.

samhogg: (made a best guess at pixel count, not sure what the best approach is though)

mcnuttandrew:
Any word on this guy?

samhogg:
@mcnuttandrew - oops, realized I didn't push - did you still want to redo this entirely though, as per your comment above?

samhogg:
@mcnuttandrew not sure what you had in mind in terms of modernizing, but I've had a go at a quick refactor to get rid of all the if statements, figured a map of options was probably better.

samhogg:
@mcnuttandrew can you take another look at this one, please?

mcnuttandrew:
Looks like there are some problems with this commit, here's the expected behavior
screen shot 2017-03-30 at 9 45 22 am

and here's the behavior from this pr
screen shot 2017-03-30 at 9 45 02 am

samhogg:
Hmmmmm, I'll do some digging and get back to you

samhogg:
@mcnuttandrew fixed in latest commit, silly typo 😬

src/plot/axis/axis-title.js
```diff
@@ -23,6 +23,32 @@ import React from 'react';
 import {ORIENTATION} from 'utils/axis-utils';
 
 const {LEFT, RIGHT, TOP, BOTTOM} = ORIENTATION;
+const transformation = (width, height) => ({
```
mcnuttandrew: jsdoc?

src/plot/axis/axis-title.js
```diff
@@ -23,6 +23,32 @@ import React from 'react';
...
+  [LEFT]: {
+    x: 16,
```
mcnuttandrew: magic numbers?

src/plot/axis/axis-title.js
samhogg: Sure thing.

src/plot/axis/axis-title.js
samhogg: Essentially, yes - I just converted the em values given to an approximation in pixels (assuming 1em === 16px)

Just realised that I messed up the top y coordinate though, should be 16, will fix that. Will also change the rotations to number values, they're strings as a remnant of the CSS syntax having the `deg` suffix.

mcnuttandrew: I hear that, i think we will most likely run into some collision (eg users trying to manipulate distances using em) later around this kind of hard constant setting, but i this is alright for now. What i am suggesting with this nit is that you pulled these numbers up to the top of the file as constants, so it is explicitly clear why they are set this way

samhogg: Ok, so I'm not entirely clear on your suggested fix, should I pull out the assumed base `em`  as a constant and multiply that?

mcnuttandrew: no what i'm saying is add a constant on line 24

const ADJUSTMENT_FOR_TEXT_SIZE = 16;

and then use that in your transformation function

samhogg: Ok, cool - done.

samhogg:
Thanks!

PR #291, opened 2017-03-15
Fancy docs page

mcnuttandrew:
We have a small, but growing, collection of examples that we should be able to show off! Count this alongside how intensely unpleasant it is to browse our documentation, this PR re-creates the docs app from deck.gl in order to provide a better front page for react-vis. 

This is very much a WIP Pr, it still needs
- some design love
- a home page
- ???

It also breaks apart many markdown files in the interest of a greater granularity of documentation. This will allow us to iterate on these articles much faster.


docs/docs-app/stylesheets/mixins.scss
```diff
@@ -0,0 +1,25 @@
+@mixin linearGradientV($property_name, $top, $bottom){
```
Apercu: Can't we use autoprefixer instead of creating these mixins?

mcnuttandrew: i'm not sure i know what you are referring to?

mcnuttandrew:
The app still needs some more love, but this pr is already pretty sprawling so i'm going to iterate on the design and organization of the docs in subsequent prs

Issue #184, opened 2016-11-22
HorizontalBarSeries does not render correctly 


nbryant:
Given a typical horizontal bar chart setup:

```javascript
<XYPlot
  height={height}
  width={width}
  yType="ordinal"
>
  <XAxis />
  <YAxis />
  <HorizontalBarSeries
    color="steelblue"
    data={[
      { x: 4, y: 'A' },
      { x: 5, y: 'B' },
      { x: 3, y: 'C' },
   ]}
  />
</XYPlot>
```

The chart and axes render, but the bars do not. This error is in the console: `DOMPropertyOperations.js:142 Error: <rect> attribute height: A negative value is not valid. ("-99.16666666666669")`.

It appears that the `<rect>`'s `y` and `height` attributes are not being computed correctly.

FWIW, manually reversing the y-axis domain will solve this problem:

```javascript
<XYPlot
  height={height}
  width={width}
  yType="ordinal"
  // manually reverse the y-axis range
  // subtract top and bottom margins
  yRange={[-10, height - 40]}
>
```

danhipke:
@nbryant I ran into this issue too and the fix is in the master branch.

nbryant:
nice! not sure when I'll have time to confirm, but the fix is much appreciated.

mcnuttandrew:
This should now be fixed in v1.1.0.

PR #300, opened 2017-03-20
Fixed scale distance calculation for ordinal/category type to prevent…

danhipke:
… negative values (when horizontal bar used). Fix for #184


danhipke:
Definitely 👍

danhipke:
@mcnuttandrew Done! Let me know if anything needs changing

Issue #147, opened 2016-09-13
Provide a better support of the animations


bulyonov:
Currently animations lack good API. They are missing following things:

- the parameters of the animation (either duration/mode or the react-motion parameters);
- event listeners such as `onAnimationStart` or `onAnimationEnd`.

Also the `makeWidthFlexible` function should probably turn animation off on the first initialization (it causes the "slide" animation for the entire chart)

mcnuttandrew:
issues: 

- [X] the parameters of the animation
- [ ] event listeners
- [X] Also the makeWidthFlexible function should probably turn animation off on the first initialization

jckr:
we're going to deprecate make...Flexible. so the only thing we need is event listeners.

PR #301, opened 2017-03-20
Disable animation on flexible width when width is zero

mcnuttandrew:
This addresses part of #147. The logic here is that the flexibility of width shouldn't be animated on first load. Also, At this point #147 is pretty stale, so I'd be willing to drop this pr if others felt it was unnecessary.


mcnuttandrew:
What does this do? It prevents the slide in animation on load

PR #302, opened 2017-03-20
Style up the home page

mcnuttandrew:


mcnuttandrew:
styleup

I'm thinking of this PR less as a final iteration, and more as a step on the way to completion, once this lands we will be able to publish the new docs! Open to other's comments and suggestions on the styling.

Apercu:
Looks good, but I think the banner is way to big, trying to fit the graphs into it might not be a good idea, we also loose some space on the right of the text.

Why not dispatch each graph next to each paragraph, like a placeholder thing. We could even alternate the order paragraph / graph left and right, might be interesting to try out.

docs/docs-app/components/header.js
```diff
@@ -19,16 +19,21 @@ const renderTab = (tab, index) => (
...
 );
 
+const renderGithubLink = () => (
+  <a href="https://github.com/uber/react-vis" className="link">GITHUB</a>
+);
+
 export default class Header extends Component {
 
   render() {
+    const mappedTabs = tabs.map(renderTab).concat(renderGithubLink());
```
Apercu: Could be taken out of the render

docs/docs-app/demo.scss
```diff
@@ -52,20 +54,38 @@ a:visited {
   display: flex;
 }
 
+.f-down {
```
Apercu: `.fcol` could be a better naming

mcnuttandrew:
homepage

jckr:
a word of caution - the colors of the default react-vis palette work best against a white background. 
2nd best against a very dark background. 
against a grey background, you'll always lose contrast so you'd have to compensate somehow. maybe use a slightly off-balance grey (ie more red or more blue or more green) and a complementary tint? 
against a colored background, some of these colors will definitely look weird.

PR #304, opened 2017-03-21
Gl scatterplot

mcnuttandrew:
gl-scatterplot

This is an initial pass at making use of deck.gl for a abstract vis rendering engine. This PR adds support for a new series type called MarkSeriesGL, what it's still missing is

- Support for mouse events
- Opacity finesse
- Tests

Delighted to hear thoughts on this approach

### DIFF
--- a/src/plot/series/mark-series-gl.js
+++ b/src/plot/series/mark-series-gl.js
@@ -37,8 +37,8 @@ class MarkSeriesGL extends AbstractSeries {
     return false;
   }
 
-  _renderScatterplotLayer(props) {
-    const {data, _renderKey, seriesId} = props;
+  _renderScatterplotLayer() {
+    const {data, _renderKey, seriesId} = this.props;
     const xFunctor = this._getAttributeFunctor('x');
     const yFunctor = this._getAttributeFunctor('y');
     const sizeFunctor = this._getAttributeFunctor('size');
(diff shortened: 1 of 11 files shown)


ibgreen:
@mcnuttandrew Very cool. Not sure if this is of interest, but an idea we have been playing with in deck.gl is to provide better React wrappers allowing deck.gl layers be specified as React components.  The idea would be to allow the following type code that might make deck.gl easier to use for your purposes.
```
<DeckGL ...>
    <ScatterplotOverlay ...>
</DeckGL>
```
Perhaps implemented using something like: 
```
class ScatterplotOverlay extends React.Component {
   constructor() {
      this.Layer = ScatterplotLayer;
   }
   render() { return null; }
}

class DeckGL extends React.Component {
  render() {
    let layers = [];
    React.children.forEach(child => {
      if (child.Layer) {
         layers.push(new child.Layer(child.props));
       }
     }
  }
}
```
Maybe an adaptation of this idea could be helpful in your integration?

mcnuttandrew:
@ibgreen That's a great idea! I feel like that would ease legibility lots of different places

src/plot/series/mark-series-gl.js
```diff
@@ -0,0 +1,118 @@
...
+    const glViewport = new OrthographicViewport({
+      width: width || 0,
+      height: height || 0,
+      left: -marginLeft,
+      top: -marginTop
+    });
+    return (
+      <div className={`${predefinedClassName} ${className}`}>
+        {innerHeight && innerWidth && <DeckGL width={width} height={height} viewport={glViewport}
+          style={{position: 'absolute', top: '0px', left: '0px'}}
```
Apercu: `top: 0, left: 0`

src/plot/series/mark-series-gl.js
```diff
@@ -0,0 +1,118 @@
...
+          layers={[this._renderScatterplotLayer(this.props)]}/>}
```
Apercu: Why passing the props here? Juste get them from the method

Apercu:
Alternative implementation over ib comment

```
const DeckGL = ({children}) => 
  children.map(chid => child.Layer && new child.Layer(child.props)).filter(l => l)
```

mcnuttandrew:
homepage

PR #306, opened 2017-03-22
Reorganize showcase so that it can be used as an example

mcnuttandrew:
showcase

Also address @jckr's comment about colors from the previous PR


showcase/app.js
```diff
@@ -0,0 +1,38 @@
...
+  render() {
+    return <ShowcaseApp />;
+  }
+}
+
+const el = document.createElement('div');
+document.body.appendChild(el);
+
+ReactDOM.render(React.createElement(App), el);
+// ReactDOM.render(<App />, document.querySelector('#index'));
```
Apercu: we can get rid of this

PR #309, opened 2017-03-22
Create docs deployment script

mcnuttandrew:


deploy-docs.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+rm -rf docs-dist &&
```
Apercu: you kept the `&&`?

mcnuttandrew: it didn't run in sequence if i didn't?

PR #317, opened 2017-03-24
MarkSeries Gl and LineSeries GL

mcnuttandrew:
line-gl-example
scatterplot-gl-example

This PR adds two new series, a gl mark series, and a gl line series. They are still waiting on some completion events. The mark series still needs some finesse on the stroke properties, and it needs to be determined whether or not the path series can generate dashed lines.

...Previous branch got fried in a clean up


shaojingli:
deck.gl v4 in action!

mcnuttandrew:
So i've hit a snag that I am kinda confused on, which is how to handle multiple instances of deck gl in a single XYPlot.

- One approach has us bunching them up at the bottom of XY, just the same way we do for SVG. I tried this out here: https://github.com/mcnuttandrew/react-vis/tree/alt-deck. This approaches loses a lot of the react magic as it runs a lot closer to how deck.gl actually runs. It runs and renders fine, but loses the animation aspect
- Another just says, disallow this behavior?

ibgreen:
> . It runs and renders fine, but loses the animation aspect

I wonder why? Can you provide a link to actual code lines? I'd be happy to take a look at it.

mcnuttandrew:
https://github.com/mcnuttandrew/react-vis/pull/2/files

src/plot/series/deck-gl-wrapper.js
```diff
@@ -0,0 +1,96 @@
...
+      marginLeft,
+      marginTop,
+      marginBottom,
+      marginRight,
+      innerHeight,
+      innerWidth,
+      _renderKey
+    } = this.props;
+
+    if (animation) {
```
mcnuttandrew: the crucial question now is how to make this section animate the layers correctly! am very confused

showcase/plot/scatterplot-gl.js
```diff
@@ -0,0 +1,104 @@
...
+    const {glMode, data, colorType, value} = this.state;
+    const markSeriesProps = {
+      animation: true,
+      className: 'mark-series-example',
+      sizeRange: [5, 15],
+      seriesId: 'my-example-scatterplot',
+      colorRange: colorRanges[colorType],
+      opacityType: 'literal',
+      data,
+      onValueMouseOver: val => this.setState({value: val})
```
Apercu: maybe rename to value

src/plot/series/deck-gl-wrapper.js
```diff
@@ -0,0 +1,73 @@
...
+      top: -marginTop
+    });
+    if (!innerHeight || !innerWidth) {
+      return null;
+    }
+
+    const layers = children.reduce((res, layer) => {
+      if (!layer.type.renderLayer) {
+        const trueLayer = layer.props.children;
+        return res.concat(trueLayer.type.renderLayer({...trueLayer.props}));
```
Apercu: do we need to clone props there or can we just pass the object?

src/plot/series/deck-gl-wrapper.js
```diff
@@ -0,0 +1,73 @@
...
+      }
+
+      return res.concat(layer.type.renderLayer({...layer.props}));
```
Apercu: same

mcnuttandrew:
Now that deck.glv4 is released i think this guy is ready for a real review!

src/plot/series/line-series-gl.js
```diff
@@ -0,0 +1,89 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import {PropTypes} from 'react';
```
btford: PropTypes will soon be its own package. Again, this is perhaps out of scope, but we should think about moving this repo to flowtype or using the new prop-types package.

src/plot/series/mark-series-gl.js
```diff
@@ -0,0 +1,100 @@
...
+      _renderKey,
+      seriesId,
+      colorDomain,
+      colorRange,
+      sizeDomain,
+      sizeRange,
+      fp64
+    } = props;
+
+    const xFunctor = getAttributeFunctor(props, 'x');
```
btford: This is also a bit out of scope but for posterity...

Not sure if I'd call this a `functor` unless I wanted to evoke its meaning from category theory (and the corresponding fantasy-land spec for JS). I know this is what devs in C++ call function objects, but that terminology isn't really useful in JS since we already have first-class functions.

mcnuttandrew: tbh, this is a copy-pasta labeling. i 100% support a relabeling of functors as functions

src/plot/series/line-series-gl.js
mcnuttandrew: Will you open an issue about this? I think this is something worth discussing

showcase/plot/line-chart-gl.js
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
mcnuttandrew: We've actually run into this issue before! i think @ibgreen suggested that we just remove the copyright and have a single copy right page at the root of the library, which would probably be better

src/plot/series/line-series-gl.js
btford: Yes– and it looks like someone already has a PR for this :) https://github.com/uber/react-vis/issues/369

src/plot/series/mark-series-gl.js
btford: Will open follow up ticket.

PR #335, opened 2017-03-31
Allow examples to have associated markdown files

mcnuttandrew:
That way we can describe why we are presenting these examples without maintaining a bunch of html!

### DIFF
--- a/docs/docs-app/components/utils.js
+++ b/docs/docs-app/components/utils.js
@@ -1,0 +1,39 @@
+import React from 'react';
+import marked from 'marked';
+
+import {docsRouting} from '../constants/pages';
+import {showCase} from '../../../showcase/index';
+
+const INJECTION_REG = /<!-- INJECT:"(.+)\" -->/g;
+
+export function injectExamplesIntoHtml(content) {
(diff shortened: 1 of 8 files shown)


docs/docs-app/components/utils.js
```diff
@@ -0,0 +1,39 @@
...
+        <Example />
+      </div>
+    );
+  });
+}
+
+export function convertMarkdownToReact(content) {
+  const renderer = new marked.Renderer();
+  renderer.link = (href, title, text) => {
+
```
Apercu: extra newline

docs/docs-app/components/utils.js
```diff
@@ -0,0 +1,39 @@
...
+    if (docsRouting[href]) {
+      return `<a href=${docsRouting[href]} title=${title}>${text}</a>`;
+    }
+    return `<a href=${href} title=${title}>${text}</a>`;
```
Apercu: Might be better to do

```
renderer.link = (href, title, text) =>
  `<a href=${docsRouting[href] || href} title=${title}>${text}</a>`
```

Issue #299, opened 2017-03-20
Radar Chart?


nikolay-g:
I couldn't see any example of a [Radar Chart](https://en.wikipedia.org/wiki/Radar_chart). Does react-vis support it?

Apercu:
Not at the time, but a PR would be appreciated! 😃

mcnuttandrew:
I've been thinking Radar charts would be part of a larger story around r-theta plots (#209), which would enable us to do a whole variety of additional polar coordinate based charts!

mcnuttandrew:
At last radar charts are part of the library! Woot! They'll be up int he next release

PR #340, opened 2017-04-03
Circular grid lines

mcnuttandrew:
This PR adds a circular grid line component. This is a valuable component to have if users wish to make radial scatter plots (see below), or when making radar charts (#299)

screen shot 2017-04-02 at 5 31 25 pm

### DIFF
--- a/src/plot/circular-grid-lines.js
+++ b/src/plot/circular-grid-lines.js
@@ -43,9 +43,7 @@ const propTypes = {
   tickTotal: PropTypes.number,
 
   animation: AnimationPropType,
-
-  // Not expected to be used by the users.
-  // TODO: Add underscore to these properties later.
+  // generally supplied by xyplot
   marginTop: PropTypes.number,
   marginBottom: PropTypes.number,
   marginLeft: PropTypes.number,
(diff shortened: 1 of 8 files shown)


src/plot/circular-grid-lines.js
```diff
@@ -0,0 +1,131 @@
...
+  height: PropTypes.number,
+  top: PropTypes.number,
+  left: PropTypes.number,
+
+  tickValues: PropTypes.array,
+  tickTotal: PropTypes.number,
+
+  animation: AnimationPropType,
+
+  // Not expected to be used by the users.
```
Apercu: but where does they come from?

src/plot/circular-grid-lines.js
```diff
@@ -0,0 +1,131 @@
...
+      height: innerHeight,
+      tickTotal: getTicksTotalFromSize(Math.min(innerWidth, innerHeight))
+    };
+  }
+
+  render() {
+    const {animation, centerX, centerY} = this.props;
+    if (animation) {
+      return (
+        <Animation {...this.props} {...{animatedProps}}>
```
Apercu: hum, don't know how I feel about spreading an object just to avoid doing `animatedProps={animatedProps}`

src/plot/circular-grid-lines.js
mcnuttandrew: this is just copy-pasta pasta comments. regardless it comes form XYPlot

Issue #25, opened 2016-03-11
Is there some graph map with edges?


ardok:
Let's say we start with one node, attach a click listener in which it will fetch some data, then we will expand that "graph map" with more edges connected with these other nodes that we get from our new data.

bulyonov:
@ardok , we don't have any graph support in the library now. The next release will be definitely extending the existing functionality for charts. Different types of graphs are in our roadmap, but we don't have any estimate for that yet.

williaster:
@bulyonov is there a published roadmap anywhere?

mcnuttandrew:
@ardok you might be interested in this library https://github.com/uber/react-vis-force

mcnuttandrew:
Also of interest, https://github.com/uber/react-digraph

jckr:
no plan to build that in react-vis for now.

PR #341, opened 2017-04-03
Force directed graph demo

mcnuttandrew:
Not everything react-vis can draw has to be a chart per say! I feel that one of the core value adds is the XYPlot's ability move things in coordinate space into pixel space by smartly handling the domain and range manipulation. Taking this notion to the extreme I tried throwing out the coordinate system all together, and it seems fine

force

This connects to #25


mcnuttandrew:
> Neat, do you know exactly what react-vis-force is doing more? It might be a good opportunity to merge both of them if react-vis is supporting force

I dunno! That's a larger effort (which we should totally do!), the point of this demo is just to show that you can do a lot more than just charts in react-vis

Issue #209, opened 2017-01-09
r-theta plot


mcnuttandrew:
Our current stratagey for radial plots works alright for donuts/pies/radial charts, but there is a constellation of additional chart types we might want. I propose that we add a notion of R-Theta Plots (to match the XY Plots) for controlling charts the express themselves in polar coordinates. This will allow us to make partition graphs, radial scatterplots, polar rose diagrams.

mcnuttandrew:
For anyone watching this thread, my current plan of attack to address this feature is as follows 

- Introduce huge amount of tests
- Add a concept of accessors to series, so that users can provide things like 
```
xAccessor={row => row.get('my-x-col-name')}
```
This in turn will open us up to immutable.js, (which is a super cool feature by itself but will require some refactoring of data processing logic, again the important of the first step being tests)
- Introduce a new object called RThetaPlot that wraps all of it's children in a coordinate transform moving polar to XY

mcnuttandrew:
After a lot of thinking, I think this is the wrong tree to be barking up. As seen in arc-series it is perfectly possible to get all the functionality we need from the XYPlot! I still think introducing accessors is a good project, but I think it deserves it's own ticket. Marking as invalid.

Issue #46, opened 2016-04-06
Add hints for radial chart


bulyonov:

stolendog:
it would be great to have hints on radial chart.

mcnuttandrew:
The functionality is now available on the 2.0.0beta release, this ticket can be closed.

PR #344, opened 2017-04-03
Arc Series

mcnuttandrew:
This PR enables the plotting of arcs in cartesian space, while still allowing users to specify arcs in polar coordinates! This is a significant PR as it means **we do not have to do the r-theta plot** (#209)!

arc-series

Eventually the radial chart should be re-written to just be a wrapper on the XYPlot and arcs, as this will allow us to finally add hints (#46)


docs/markdown/arc-series.md
```diff
@@ -0,0 +1,133 @@
...
+  <ArcSeries
+    animation
+    radiusType={'literal'}
+    center={{x: -2, y: 2}}
+    data={myData}
+    colorType={'literal'}/>
+</XYPlot>
+```
+
+A key point: XYPlot infers the necessary x and y domains by converting the angular coordinates to cartesian. If you want to guarentee a centered plot it is advisable to set the x and y domain's yourself, as above.
```
Apercu: garantee

docs/markdown/arc-series.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Data format Reference
+
+Arc series works a little bit different than other series. The most natural language to describe arcs is using polar coordinates, so we allow
```
Apercu: > The most natural way to describe arcs is by using polar coordinates, thus our decision to go this way.

maybe?

showcase/radial-chart/arc-series-example.js
```diff
@@ -0,0 +1,100 @@
...
+  YAxis
+} from 'index';
+
+import {EXTENDED_DISCRETE_COLOR_RANGE as COLORS} from 'theme';
+
+const PI = Math.PI;
+
+function updateData() {
+  const divider = Math.floor(Math.random() * 8 + 3);
+  const newData = (new Array(5)).fill(0).map((row, index) => {
```
Apercu: new way to do the map in 3 less chars `[...new Array(5)].map()`

src/utils/series-utils.js
```diff
@@ -68,6 +68,33 @@ function collectSeriesTypesInfo(children) {
 }
 
 /**
+ * Check series to see if it has angular data that needs to be converted
+ * @param {Array} data - an array of objects to check
+ * @returns {Boolean} whether or not this series contains polar configuration
+ */
+function seriesHasAngleRadius(data) {
```
Apercu: default param instead of the `||` maybe

PR #345, opened 2017-04-04
Pass style

jckr:
In this PR I pass style as a prop to any react-vis component (series, axis, gridlines, hint etc) and that component will take this styling information into account. With this it's possible to completely override style info from the stylesheet or even to style a chart without importing the stylesheet.
I've added one example where axes, gridlines and series are restyled through props.

### DIFF
--- a/docs/markdown/axes.md
+++ b/docs/markdown/axes.md
@@ -102,3 +102,18 @@ Height of the axis in pixels. **Already set by default**, but can be overriden b
 
 #### animation (optional)
 See the [XYPlot](xy-plot.md)'s `animation` section for more information.
+
+### style (optional)
+Type: `object`
+An object that contains CSS properties with which the axis component can be entirely re-styled.
+As the Axis component is composite, it is possible to style its different parts individually. See [style](style.md)
+
+The various parts of the axis can be styled by passing an object to the `line`, `ticks`, `text` and `title` properties:
+
+```jsx
+<XAxis title="X Axis" style={{
+  line: {stroke: '#ADDDE1'},
+  ticks: {stroke: '#ADDDE1'},
+  text: {stroke: 'none', fill: '#6b6b76', fontWeight: 600}
+}}/>
+```
--- a/docs/markdown/bar-series.md
+++ b/docs/markdown/bar-series.md
@@ -117,3 +117,7 @@ The opacity for all elements in the series, this property will be over-ridden by
 #### stroke
 Type: `string|number`
 The outer color for all elements in the series, this property will be over-ridden by color specified in the data attribute.
+
+### style
+Type: `object`
+A list of CSS properties to style the series outside of the explicitly set properties. Note that it will override all other properties (ie fill, stroke, opacity, color). See [style](style.md)
--- a/docs/markdown/hint.md
+++ b/docs/markdown/hint.md
@@ -11,3 +11,24 @@ In case if custom representation of is needed, the component is also able to wra
   </div>
 </Hint>
 ```
+
+### Style (optional)
+Type: `Object`
+You can pass a style object to your Hint component to apply your own styles. See [style](style.md)
+```jsx
+<Hint value={value} style={{fontSize: 14}}/>
+```
+
+Style is a composite component, and individual parts of it can receive different parts.
+The different parts are: content, row, title, value. To style a specific part, you can pass an object to the property with that name.
+```jsx
+<Hint value={value} style={{
+  fontSize: 14,
+  text: {
+    display: 'none'
+  },
+  value {
+    color: 'red'
+  }
+}}/>
+```
--- a/docs/markdown/series.md
+++ b/docs/markdown/series.md
@@ -75,5 +75,43 @@ Type: `function(info)`
 Type: `function(info)`
 `click` event handler for the entire series. Received `info` object as argument with the only `event` property.
 
+#### style (optional)
+Type: `object`
+An object which holds CSS properties that will be applied to the SVG element(s) rendered by the series. This allows you to style series beyond the other explicitly defined properties and without having to use CSS classnames and stylesheets. For instance, you can set the stroke-linejoin style of a line series to "round":
+```jsx
+<LineSeries
+  data={data}
+  style={{strokeLinejoin: "round"}}
+/>
+```
+LineMark series is a composite series, and as such, it's possible to separate style instructions for the line and the mark part by putting them under a "line" and a "mark" property respectively: 
+
+```jsx
+<LineMarkSeries
+  data={data}
+  style={{
+  	// affect both the line and the mark part
+  	stroke: "red",
+  	.line: {
+  	  // affects just the line series
+  	  strokeWidth: 2
+  	},
+  	.mark {
+  	  // affects just the mark series
+  	  strokeWidth: 4
+  	}
+  }}
+/>
+```
+
+Note that style information passed through the style property will override those passed through props.
+```jsx
+<MarkSeries
+  data={data}
+  /// all the points are red
+  style={{fill: "red"}}
+/>
+```
+
 #### animation (optional)
 See the [XYPlot](xy-plot.md)'s `animation` section for more information.
--- a/src/plot/axis/axis-line.js
+++ b/src/plot/axis/axis-line.js
@@ -27,14 +27,19 @@ import {ORIENTATION} from 'utils/axis-utils';
 const {LEFT, RIGHT, TOP, BOTTOM} = ORIENTATION;
 
 const propTypes = {
-  width: PropTypes.number.isRequired,
   height: PropTypes.number.isRequired,
+  style: PropTypes.object,
   orientation: PropTypes.oneOf([
     LEFT, RIGHT, TOP, BOTTOM
-  ]).isRequired
+  ]).isRequired,
+  width: React.PropTypes.number.isRequired
 };
 
-function AxisLine({orientation, width, height}) {
+const defaultProps = {
+  style: {}
+};
+
+function AxisLine({orientation, width, height, style}) {
   let lineProps;
   if (orientation === LEFT) {
     lineProps = {
@@ -66,12 +71,12 @@ function AxisLine({orientation, width, height}) {
     };
   }
   return (
-    <line {...lineProps} className="rv-xy-plot__axis__line"/>
+    <line {...lineProps} className="rv-xy-plot__axis__line" style={style}/>
   );
 }
 
+AxisLine.defaultProps = defaultProps;
 AxisLine.displayName = 'AxisLine';
 AxisLine.propTypes = propTypes;
 
 export default AxisLine;
-
--- a/src/plot/axis/axis-ticks.js
+++ b/src/plot/axis/axis-ticks.js
@@ -28,11 +28,16 @@ import {getAttributeScale} from 'utils/scales-utils';
 const {LEFT, RIGHT, TOP, BOTTOM} = ORIENTATION;
 
 const propTypes = {
-  width: PropTypes.number.isRequired,
   height: PropTypes.number.isRequired,
   orientation: PropTypes.oneOf([
     LEFT, RIGHT, TOP, BOTTOM
-  ]).isRequired
+  ]).isRequired,
+  style: PropTypes.object,
+  width: PropTypes.number.isRequired
+};
+
+const defaultProps = {
+  style: {}
 };
 
 function _getTickFormatFn(scale, tickTotal, tickFormat) {
@@ -151,6 +156,7 @@ class AxisTicks extends React.Component {
       orientation,
       width,
       height,
+      style,
       tickFormat,
       tickTotal,
       tickValues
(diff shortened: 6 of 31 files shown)


main.css
```diff
@@ -0,0 +1,139 @@
+@import url(https://fonts.googleapis.com/css?family=Sintony:400,700);
```
mcnuttandrew: delete this file

src/plot/axis/axis-line.js
```diff
@@ -26,13 +26,14 @@ const {LEFT, RIGHT, TOP, BOTTOM} = ORIENTATION;
 
 const propTypes = {
```
mcnuttandrew: alphabetize the props

src/plot/axis/axis-line.js
```diff
@@ -26,13 +26,14 @@ const {LEFT, RIGHT, TOP, BOTTOM} = ORIENTATION;
...
   width: React.PropTypes.number.isRequired,
+  style: React.PropTypes.object,
   height: React.PropTypes.number.isRequired,
   orientation: React.PropTypes.oneOf([
     LEFT, RIGHT, TOP, BOTTOM
   ]).isRequired
 };
 
-function AxisLine({orientation, width, height}) {
+function AxisLine({orientation, width, height, style = {}}) {
```
mcnuttandrew: prefer defaultProps

src/plot/axis/axis-ticks.js
```diff
@@ -28,6 +28,7 @@ const {LEFT, RIGHT, TOP, BOTTOM} = ORIENTATION;
 const propTypes = {
   width: React.PropTypes.number.isRequired,
```
mcnuttandrew: alphabetize

src/plot/axis/axis-ticks.js
```diff
@@ -171,9 +173,17 @@ class AxisTicks extends React.Component {
...
       return (
-        <g key={i} {...translateFn(pos, 0)} className="rv-xy-plot__axis__tick">
-          <line {...pathProps} className="rv-xy-plot__axis__tick__line"/>
-          <text {...textProps} className="rv-xy-plot__axis__tick__text">
+        <g key={i} {...translateFn(pos, 0)}
+        className="rv-xy-plot__axis__tick"
+        style={style}>
+          <line {...pathProps}
+            className="rv-xy-plot__axis__tick__line"
+            style={{...style, ...style.line}}
```
mcnuttandrew: add documentation around this?

src/plot/axis/axis.js
```diff
@@ -47,6 +47,8 @@ const propTypes = {
   left: PropTypes.number,
   title: PropTypes.string,
 
+  style: PropTypes.object,
```
mcnuttandrew: alpha

src/plot/axis/axis.js
```diff
@@ -165,17 +168,21 @@ class Axis extends PureRenderComponent {
...
+        style={style}>
         {!hideLine && (<AxisLine
           height={height}
           width={width}
-          orientation={orientation}/>)}
-        {!hideTicks && (<AxisTicks {...props} />)}
+          orientation={orientation}
+          style={{...style, ...style.line}}
+          />)}
+        {!hideTicks && (<AxisTicks {...props} style={{...style, ...style.ticks}}/>)}
```
mcnuttandrew: document the use of the style object

src/plot/hint.js
```diff
@@ -361,25 +362,27 @@ class Hint extends PureRenderComponent {
...
         style={{
-          ... style,
+          ...style,
+          ...position,
           position: 'absolute'
         }}>
         {children ?
           children :
-          <div className="rv-hint__content">
+          <div className="rv-hint__content" style={style.content}>
```
mcnuttandrew: document

src/plot/series/area-series.js
```diff
@@ -45,7 +45,7 @@ class AreaSeries extends AbstractSeries {
 
   render() {
     const {
-      animation, className, curve, data, marginLeft, marginTop
+      animation, className, curve, data, marginLeft, marginTop, style
```
mcnuttandrew: should also add this to the abstract series propTypes and and defaultProps

src/plot/series/line-mark-series.js
```diff
@@ -31,10 +31,11 @@ class LineMarkSeries extends AbstractSeries {
   }
 
   render() {
+    const {style = {}} = this.props;
     return (
       <g className="rv-xy-plot__series rv-xy-plot__series--linemark">
-        <LineSeries {...this.props} />
-        <MarkSeries {...this.props} />
+        <LineSeries {...this.props} style={{...style, ...style.line}}/>
```
mcnuttandrew: add documentation

src/plot/xy-plot.js
```diff
@@ -289,7 +290,8 @@ class XYPlot extends React.Component {
           className={`rv-xy-plot ${className}`}
           style={{
             width: `${width}px`,
-            height: `${height}px`
+            height: `${height}px`,
+            ...this.props.style
```
mcnuttandrew: why this.props and not just style?

jckr: imo, destructuring this.props doesn't make things more legible, but in the name of consistency I did it.

mcnuttandrew:
you need to run "npm run remove-unpm-rfs" to remove the unpm junk from the yarn

mcnuttandrew:
It also looks like you are failing lint

docs/markdown/bar-series.md
```diff
@@ -117,3 +117,7 @@ The opacity for all elements in the series, this property will be over-ridden by
 #### stroke
 Type: `string|number`
 The outer color for all elements in the series, this property will be over-ridden by color specified in the data attribute.
+
+### style
+Type: `object`
+A list of CSS properties to style the series outside of the explicitly set properties. Note that it will override all other properties (ie fill, stroke, opacity, color).
```
Apercu: I don't agree with the use of the word "CSS" here, "style properties" would be more fitting in my opinion.
Also, we should have one consistent definition and not attempt to change it between components. I'm not talking about possible different examples between the components.

The only exception to this would be the axes one (and other ones below I haven't saw yet that function the same way). One other thing I would say would be that in the case of axe, rename the `style` prop into something else, like `styles`. This way, it will allow users to tell immediately if it's a simple style, or an object containing multiple.

docs/markdown/axes.md
```diff
@@ -74,3 +74,18 @@ Height of the axis in pixels. **Already set by default**, but can be overriden b
 
 #### animation (optional)
 See the [XYPlot](xy-plot.md)'s `animation` section for more information.
+
+### style (optional)
```
Apercu: should use consistent number of headings, and wording (`Style` instead of `style` in `docs/markdown/hint.md`)

src/plot/xy-plot.js
Apercu:
```
style={{width, height, ...style}}
```

docs/markdown/hint.md
```diff
@@ -11,3 +11,24 @@ In case if custom representation of is needed, the component is also able to wra
...
+```jsx
+<Hint value={value} style={{fontSize: 14}}/>
+```
+
+Style is a composite component, and individual parts of it can receive different parts.
+The different parts are: content, row, title, value. To style a specific part, you can pass an object to the property with that name.
+```jsx
+<Hint value={value} style={{
+  fontSize: 14,
+  .text: {
```
akre54: Is the `text` a property or a class? Either way looks like the `.` is a mistake

docs/markdown/line-series.md
```diff
@@ -43,3 +43,13 @@ const stringCurveProp = <LineSeries data={data} curve={'curveMonotoneX'} .../>;
 const configuredCurve = d3Shape.curveCatmullRom.alpha(0.5);
 const funcCurveProp = <LineSeries data={data} curve={configuredCurve} .../>;
 ```
+
+#### style (optional)
+Type: `object`
+An object which holds CSS properties that will be applied to the SVG element(s) rendered by the series. This allows you to style series beyond the other explicitly defined properties and without having to use CSS classnames and stylesheets.
```
akre54: missing word: "style series" => "style the series"

docs/markdown/series.md
```diff
@@ -75,5 +75,43 @@ Type: `function(info)`
 Type: `function(info)`
 `click` event handler for the entire series. Received `info` object as argument with the only `event` property.
 
+#### style (optional)
+Type: `object`
+An object which holds CSS properties that will be applied to the SVG element(s) rendered by the series. This allows you to style series beyond the other explicitly defined properties and without having to use CSS classnames and stylesheets. For instance, you can set the stroke-linejoin style of a line series to "round":
```
akre54: "style series" => "style the series"

docs/markdown/series.md
```diff
@@ -75,5 +75,43 @@ Type: `function(info)`
...
+```
+LineMark series is a composite series, and as such, it's possible to separate style instructions for the line and the mark part by putting them under a "line" and a "mark" property respectively: 
+
+```jsx
+<LineMarkSeries
+  data={data}
+  style={{
+  	// affect both the line and the mark part
+  	stroke: "red",
+  	.line: {
```
akre54: Is the `.` intentional here?

src/plot/axis/axis-ticks.js
akre54: Isn't there potential for collision between the `style` and `style.line` objects here? Why not only use the more specific version?

docs/markdown/bar-series.md
jckr: 1) we can iterate on the wording. React doc (which is about the same concept) doesn't use either. They specify, however, that the properties are camel cased and not hyphenated. I'm not saying that we should use the react docs wording, but rather come up with something which is completely clear and unambiguous. 
2) I disagree with this though. Possible API attributes should be repeated, and possibly rephrased, for each and every component. The style attribute should also has its own page that goes into more detail. But eventually, for every component, such as every series type, there should be an exhaustive list of all the attributes that can be passed, phrased in the way that makes the most sense in this context. When a user wants to know how to use, ie, LineSeries, they should have access to the entirety of information on how to use it, as opposed to have to jump back and forth between pages.

docs/markdown/axes.md
jckr: ok

docs/markdown/line-series.md
jckr: I can add a the in that sentence, but I intended that sentence without a the

akre54: Ah. You're using `series` in its plural form here (English is weird). You're right it's grammatically correct in this case to use without a definite or indefinite article, but given that you refer to `the series` right before I think it makes this plural use weird here.

docs/markdown/bar-series.md
Apercu: I'm not saying that we shouldn't repeat the props that are available to use, I said that I think we should not use the same property if it behaves differently. Passing multiples styles objects in a `style` property shouldn't be allowed

samhogg:
@jckr @mcnuttandrew @Apercu 

I know you guys probably have an internal dialog going on about this at the moment but would like to offer some (unsolicited) outside perspective on this PR.

Adding child element components into the `style` object goes against React (and DOM) convention - it would break any tooling that assumes style objects contain just CSS property names such as linters or helpers.

I propose two options worth considering (using the Hint example in the PR)

1. Use a `styles` object that explicitly validates subcomponents using PropTypes.

```jsx
// src/plot/hint.js
Hint.propTypes = {
  ...,
  styles: React.PropTypes.shape({
    container: React.PropTypes.object,
    text: React.PropTypes.object.
    value: React.PropTypes.object
  })
}
```

```jsx
// example render
<Hint
  value={value}
  styles={{
    container: {
      fontSize: '14px'
    },
    text: {
      display: 'none'
    },
    value: {
      color: 'red'
    }
  }}
/>
```

2. Pass separate `style` props to the component, with the `style` prop being the container or outermost DOM element.

```jsx
// src/plot/hint.js
Hint.propTypes = {
  ...,
  style: React.PropTypes.object
}
```

```jsx
// example render
<Hint
  value={value}
  style={{
    fontSize: '14px'
  }}
  textStyle={{
    display: 'none'
  }}
  valueStyle={{
    color: 'red'
  }}
/>
```

src/plot/axis/axis-ticks.js
samhogg: This could potentially make for weird results if you use an `em` based measurement in the base style object.

Apercu:
@samhogg Feedback is always much appreciated, thanks ;)

Your first point is exactly what I was referring to, thanks for putting all these examples together 😄 
I'm not really in favor of the second solution given it might create a lot of new props on some components, I prefer having a unified `styles` api indexed by sub-component key but that's personal I guess.

jckr:
@Apercu @samhogg I'm seeing this after having pushed another commit. Between the 2 options I actually prefer option 2. A styles prop is pretty error-prone imo. There are already some iffy camelcase decisions in react-vis which make debugging harder than it should be. ie <GridLines /> vs <Gridlines />.
I get what you mean with linters. the newer voronoi layer already implements the option 2 you describe with 2 separate style props.

samhogg:
@jckr to be honest, I'd be happy with either as long as it was consistently implemented throughout the library with validations/docs where appropriate.

I did notice the new voronoi plot, I think that's nice and clean, personally. We probably shouldn't get to the point when there's more than a couple subcomponents to be styling anyway, right?

jckr:
yes. unlikely to have something more complex than Axis for instance.

Issue #322, opened 2017-03-28
Documentation on animation is incorrect


mcnuttandrew:
The documentation claims that it accepts an object with a property duration, but it appears more realistically that it only checks for the presence or non-presence of that property.

PR #347, opened 2017-04-04
Add animation docs, expose more animation controls

mcnuttandrew:
This PR exposes more access to the react-motion library from the animation prop!

animation

This addresses #322 and some of the spirit of #147.


src/treemap/treemap-leaf.js
```diff
@@ -21,7 +21,7 @@
 import React, {PropTypes} from 'react';
 
 import Animation from 'animation';
-import {AnimationPropType} from 'utils/animation-utils';
+import {AnimationPropType} from 'animation';
```
Apercu: You have two imports from the same file, could be merged.

src/radial-chart/index.js
```diff
@@ -23,7 +23,7 @@ import {arc as arcBuilder, pie as pieBuilder} from 'd3-shape';
 
 import Animation from 'animation';
 import {getInnerDimensions, MarginPropType} from 'utils/chart-utils';
-import {AnimationPropType} from 'utils/animation-utils';
+import {AnimationPropType} from 'animation';
```
Apercu: same

PR #356, opened 2017-04-06
Add parallel coordinates example

mcnuttandrew:
Build a parallel coordinates example. On the way create a new component called manipuableAxis that allows users to specify arbitrary start and endpoints, as well as a axis domain, and magically work from there!

parallel-coords


examples/parallel-coordinates/manipulable-axis.js
```diff
@@ -0,0 +1,144 @@
...
+          <line {...tickProps} className="rv-xy-plot__axis__tick__line"/>
+          <text {...textProps} className="rv-xy-plot__axis__tick__text">
+            {tickValue(point.text)}
+          </text>
+        </g>
+      );
+    });
+}
+
+class ManipulableAxis extends AbstractSeries {
```
mcnuttandrew: This file is the meat.

examples/parallel-coordinates/manipulable-axis.js
```diff
@@ -0,0 +1,144 @@
...
+  };
+}
+
+function renderTicks(props) {
+  const {axisDomain, numberOfTicks, axisStart, axisEnd, tickValue} = props;
+  const {points} = generatePoints({axisStart, axisEnd, numberOfTicks, axisDomain});
+
+  return points
+    .map((point, index) => {
+      const tickProps = {
```
mcnuttandrew: in order to productionalize this component, this needs to manipuable from the props, specifically it needs to offer the length and automatically find the tangent to the slope

mcnuttandrew: also need to add tests and docs

samhogg:
@mcnuttandrew this looks really solid - I think your hover effect would probably look better if you bought the hovered series to the front (not sure if you can achieve that with CSS+SVG or you'd have to reorder)

When you say "productionalize this component" are you talking about integrating it into the library proper?

mcnuttandrew:
> When you say "productionalize this component" are you talking about integrating it into the library proper?

yea exactly! I think a few of the one off components i've written for the demo could be pulled into the main library. i am interested in having the manipuableAxis be in the main library as it will allow us to make radar charts and parallel coordinates. In addition I think the labelSeries from responsiveVis demo could be moved into the main library

samhogg:
Nice! I really like the approach of building things in examples first, shows how extensible the core library is. Sold work :)

examples/parallel-coordinates/app.js
```diff
@@ -0,0 +1,48 @@
...
+
+import ReactDOM from 'react-dom';
+import React, {Component} from 'react';
+import window from 'global/window';
+import document from 'global/document';
+
+import ParallelCoordinates from './parallel-coordinates';
+
+export default class App extends Component {
+  componentWillMount() {
```
javidhsueh: should remove the event listener when componentWillUnMount

examples/parallel-coordinates/app.js
```diff
@@ -0,0 +1,48 @@
...
+    window.addEventListener(
+      'resize',
+      () => this.setState({width: window.innerWidth})
```
javidhsueh: Do you actually use the state.width? Also, it would be great to have this component as a reusable wrapper called 'WidthProvider'

examples/parallel-coordinates/manipulable-axis.js
```diff
@@ -0,0 +1,144 @@
...
+      axisEnd,
+      axisDomain,
+      numberOfTicks,
+      tickValue
+    } = this.props;
+
+    if (animation) {
+      return (
+        <Animation {...this.props} {...{animatedProps}}>
+          <ManipulableAxis {...this.props} animation={null}/>
```
javidhsueh: Interesting, so here is to reuse the component itself?

mcnuttandrew: Yep! That's the way animation works everywhere in react-vis

Issue #343, opened 2017-04-03
Make docs page faster


mcnuttandrew:
- [X] Dont need to load all of the markdown at once, it can either be loaded in the background (maybe even in a webworker) or just be loaded ad-hoc
- [X] Break apart the showcase app example page

PR #357, opened 2017-04-07
Sunburst

mcnuttandrew:
sunburst

Add a top level chart type, sunbursts. This chart makes use of the recently created arcSeries to handle all of the data processing for it. Still to do

- [ ] Tests
- [ ] Docs
- [ ] JSdocs Where appropriate
- [ ] Animation is currently broken, the interpolator doesn't know how to address deeply nested data!

The API of this chart is the same as the treemap, so it should be reasonably easy for users to start using it right away! Would love comments if anyone has any

### DIFF
--- a/showcase/index.js
+++ b/showcase/index.js
@@ -55,6 +55,8 @@ import VoronoiLineChart from './plot/voronoi-line-chart';
 import SimpleTreemap from './treemap/simple-treemap';
 import TreemapExample from './treemap/dynamic-treemap';
 
+import BasicSunburst from './sunbursts/basic-sunburst';
+
 import SimpleRadialChart from './radial-chart/simple-radial-chart';
 import DonutChartExample from './radial-chart/donut-chart';
 import CustomRadiusRadialChart from './radial-chart/custom-radius-radial-chart';
@@ -106,6 +108,8 @@ export const showCase = {
 
   SimpleTreemap,
   TreemapExample,
+
+  BasicSunburst,
 
   SimpleRadialChart,
   DonutChartExample,
--- a/showcase/sunbursts/basic-sunburst.js
+++ b/showcase/sunbursts/basic-sunburst.js
@@ -21,17 +21,30 @@
 import React from 'react';
 
 import Sunburst from 'sunburst';
+import {EXTENDED_DISCRETE_COLOR_RANGE} from 'theme';
 
 import D3FlareData from '../treemap/d3-flare-example.json';
 
+/**
+ * Recursively work backwards from highlighted node to find path of valud nodes
+ * @param {Object} node - the current node being considered
+ * @returns {Array} an array of strings describing the key route to the current node
+ */
 function getKeyPath(node) {
   if (!node.parent) {
-    return 'root';
+    return ['root'];
   }
 
-  return [node.data.name].concat(getKeyPath(node.parent));
+  return [node.data && node.data.name || node.name].concat(getKeyPath(node.parent));
 }
 
+/**
+ * Recursively modify data depending on whether or not each cell has been selected by the hover/highlight
+ * @param {Object} data - the current node being considered
+ * @param {Object|Boolean} keyPath - a map of keys that are in the highlight path
+ * if this is false then all nodes are marked as selected
+ * @returns {Object} Updated tree structure
+ */
 function updateData(data, keyPath) {
   if (data.children) {
     data.children.map(child => updateData(child, keyPath));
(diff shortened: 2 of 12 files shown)


showcase/sunbursts/basic-sunburst.js
```diff
@@ -0,0 +1,98 @@
...
+  const name = node.data && node.data.name || node.name || '';
+  if (!node.parent) {
+    return 'root';
+  }
+
+  return `${getTextPath(node.parent)} > ${name}`;
+}
+
+function getKeyPath(node) {
+  // const name = node.data && node.data.name || node.name || '';
```
javidhsueh: not used?

showcase/sunbursts/basic-sunburst.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+function decorateData(data) {
+  if (!data.children) {
+    return data;
+  }
+  // add a fill to all the uncolored cells
+  if (!data.color) {
+    data.style = {
+      fill: '#223F9A'
```
javidhsueh: should make a constant for this magic color

showcase/sunbursts/basic-sunburst.js
```diff
@@ -0,0 +1,98 @@
...
+      <div className="basic-sunburst-example-wrapper">
+        <Sunburst
+          animation
+          className="basic-sunburst-example"
+          onValueMouseOver={node => {
+            const yy = getTextPath(node.parent);
+            const xx = getKeyPath(node.parent);
+            // console.log(xx)
+            // console.log(this)
+            // debugger;
```
javidhsueh: not used

showcase/sunbursts/basic-sunburst.js
```diff
@@ -0,0 +1,98 @@
...
+            this.setState({value: yy});
+          }}
+          onValueClick={node => {
+            this.setState({subState: subState ? false : node});
+          }}
+          style={{
```
javidhsueh: it should be a const 'defaultStyle'

src/plot/series/arc-series.js
```diff
@@ -142,20 +140,21 @@ class ArcSeries extends AbstractSeries {
         ref="container"
         transform={`translate(${marginLeft + xTranslate},${marginTop + yTranslate})`}>
         {data.map((row, i) => {
+          const cleanedRow = {...row};
```
javidhsueh: why we need to shallow copy the row?

javidhsueh: maybe add more comments?

src/sunburst/index.js
```diff
@@ -0,0 +1,133 @@
...
+
+    return partitionFunction(structuredInput).descendants()
+      .reduce((res, cell, index) => {
+        if (hideRootNode && index === 0) {
+          return res;
+        }
+
+        return res.concat([{
+          angle0: Math.max(0, Math.min(2 * Math.PI, x(cell.x0))),
+          angle: Math.max(0, Math.min(2 * Math.PI, x(cell.x1))),
```
javidhsueh: shouldn't it be called angle1 by following the same naming convention?

src/sunburst/index.js
```diff
@@ -0,0 +1,133 @@
...
+          radius0: Math.max(0, y(cell.y0)),
+          radius: Math.max(0, y(cell.y1)),
```
javidhsueh: radius1 instead?

src/sunburst/index.js
```diff
@@ -0,0 +1,133 @@
...
+  }
+  return {
+    ...node,
+    parent: cleanParent(node.parent)
+  };
+}
+
+class Sunburst extends React.Component {
+
+  static get propTypes() {
```
javidhsueh: colorType is missing

src/plot/xy-plot.js
```diff
@@ -46,6 +46,28 @@ const DEFAULT_MARGINS = {
   bottom: 40
 };
 
+function cleanseData(data) {
+  return data.map(series => {
+    if (!Array.isArray(series)) {
+      return series;
+    }
+    return series.map(row => ({...row, parent: null}));
```
javidhsueh: it seems to be dealing with sunburst case. Can we just have a white list of the props instead?

showcase/index.js
```diff
@@ -107,6 +109,8 @@ export const showCase = {
   SimpleTreemap,
   TreemapExample,
 
+  BasicSunburst,
```
javidhsueh: alphabetize all showcases?

src/plot/xy-plot.js
mcnuttandrew: i also want to re-write treemap to be in XY coordinates, so itll also be dealing with that.

src/sunburst/index.js
mcnuttandrew: that'd be following the d3 naming convention, not the react-vis one. There is not a notion of {attr}1, there is only {attr} and {attr}0.

showcase/index.js
mcnuttandrew: right now they are broken apart by group, but that is definately something to do in the future, specifically i think itll be addressed in #343

PR #362, opened 2017-04-07
Add additional example for arc series

mcnuttandrew:
clock

This example constructs another example of how to use the arc series, which can often be a little thorny. This pretty silly example also provides another instance of how to made non-charts using the charting lib.

### DIFF
--- a/showcase/sunbursts/clock-example.js
+++ b/showcase/sunbursts/clock-example.js
@@ -55,7 +55,7 @@ export default class ClockExample extends React.Component {
     const {time} = this.state;
     const seconds = time % 60;
     const minutes = (time / 60) % 60;
-    const hours = (time / 3600) % 24;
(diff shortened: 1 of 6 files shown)


showcase/sunbursts/clock-example.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export default class ClockExample extends React.Component {
+  state = {
+    time: 0
+  }
+
+  componentDidMount() {
+    /* eslint-disable react/no-did-mount-set-state */
+    this.setState({timerId: setInterval(() => {
```
Apercu: We don't need to put the timer in the state, simply do `this._int = setInterval(() => this.setState({ time: getSeconds() ))`, also you might want to use `requestAnimationFrame` instead

PR #364, opened 2017-04-08
Sunburst animation

mcnuttandrew:
Sunburst was previously unable to be animated due to problems with d3-interpolate's handling of deeply nested tree structures. This PR addresses this by adding a method for briefly stripping data of it's tree structure when it's about to be interpolated and then recombining with the original post animation pre-render.

animated-sunburst

### DIFF
--- a/showcase/sunbursts/animated-sunburst.js
+++ b/showcase/sunbursts/animated-sunburst.js
@@ -46,6 +46,8 @@ function updateData() {
   };
 }
 
+const DIVERGING_COLOR_SCALE = ['#00939C', '#85C4C8', '#EC9370', '#C22E00'];
+
 export default class AnimatedSunburst extends React.Component {
   state = {
     data: updateData()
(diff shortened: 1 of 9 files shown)


src/sunburst/index.js
```diff
@@ -100,14 +101,15 @@ class Sunburst extends React.Component {
         <ArcSeries {...{
           colorType,
           ...this.props,
-          animation: false,
+          animation,
           radiusDomain: [0, radialDomain],
-          data: mappedData
+          // need to present a stripped down version for interpolation
+          data: animation ? mappedData.map(row => ({...row, parent: null, children: null})) : mappedData,
```
mcnuttandrew: this same technique will most likely need to be used if we move the treemap to use XYPlot as well...

showcase/sunbursts/animated-sunburst.js
```diff
@@ -0,0 +1,76 @@
...
+    data: updateData()
+  }
+
+  render() {
+    const {data} = this.state;
+    return (
+      <div className="animated-sunburst-example-wrapper">
+        <button onClick={() => this.setState({data: updateData()})}> UPDATE DATA </button>
+        <Sunburst
+          animation={{
```
Apercu: I would just put this on one line

PR #372, opened 2017-04-11
Break apart showcase for dev legibility

mcnuttandrew:
The showcase app started sprawl pretty darn hard, this PR breaks it up. If we like we can break it up inside of the examples page so that people dont have to scroll as hard. Addresses some of #343

### DIFF
--- a/showcase/showcase-components/showcase-button.js
+++ b/showcase/showcase-components/showcase-button.js
@@ -19,25 +19,24 @@
 // THE SOFTWARE.
 
 import React from 'react';
-import LesMisData from './les-mis-data.json';
+import PropTypes from 'prop-types';
 
-import './force-directed.scss';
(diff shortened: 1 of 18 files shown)


mcnuttandrew:
I also centralized the styling of the buttons in the examples

showcase/showcase-components/showcase-button.js
```diff
@@ -0,0 +1,42 @@
...
+import PropTypes from 'prop-types';
+
+class ShowcaseButton extends React.Component {
+  render() {
+    const {buttonContent, onClick} = this.props;
+    return (
+      <button
+        className="showcase-button"
+        onClick={onClick}>
+        {buttonContent}
```
akre54: Why not use `this.props.children` instead of `buttonContent`? That way the content is arbitrary (doesn't have to be a string)

PR #376, opened 2017-04-11
Remove pre 1.0 API reference in axes.md

### LINKED ISSUE #374: Move 'breaking changes' section in README to CHANGELOG
Convention dictates that's probably the best place, and I think we should probably be using that real estate in the readme to entice users, not warn them that stuff breaks all the time.

On a similar line of thought, I'm of the opinion we should probably strip references to the pre 1.0 API in the docs (ie: line 5 of the [axes docs)[https://github.com/uber/react-vis/blob/master/docs/markdown/animation.md})

samhogg:
Fixes #374

### DIFF
--- a/docs/markdown/axes.md
+++ b/docs/markdown/axes.md
@@ -2,8 +2,6 @@
 
 <!-- INJECT:"CustomAxesOrientation" -->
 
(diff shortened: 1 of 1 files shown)


mcnuttandrew:
image

Issue #108, opened 2016-06-19
Add the better documentation on how to use the scales in react-vis


ranjithnori:
Please provide an example of how to use fill scale and innerRadius scale. Here's how I tried to do it. It didn't work as expected. I couldn't find any examples about the same.

```js
<RadialChart
  radiusDomain={[0,20]}
  innerRadiusDomain={[5,9]}
  data={[
    {angle: 2, fill: '#1a3141', radius: 10, innerRadius: 7},
    {angle: 7, fill: '#ff6f00', radius: 10, innerRadius: 7},
    {angle: 5, fill: '#1a3242', radius: 10, innerRadius: 7}
  ]}
  width={300}
  height={300} />
```

pawel2105:
Why is this issue closed? The documentation still isn't really clear.

mcnuttandrew:
Underlying this ticket is a bug in the way that the radial chart was processing the inner/outer radii. The documentation for scale still **should** be improved, however in the new 2.0.0-beta release this should be a lot better/more logical to manipulate.

PR #377, opened 2017-04-12
Rewrite Radial Chart

mcnuttandrew:
screen shot 2017-04-12 at 10 38 18 am

This PR completely re-writes the radial chart to use the new arc series approach. This rewrite address some longstanding weirdness in the old chart (such as not being able to easily specify the inner radius of wedges #108 #234) as well as adds the ability for radial charts to have hints (#46). Finally it also productionalizes the labelSeries that had been kicking around in the responsive vis (#39)

screen shot 2017-04-12 at 10 41 12 am

What's left

- [X] LabelSeries Documentation
- [X] Make the radial chart's documentation discuss the new chart

**NOTE** This is a breaking change. Specifically it does away with radial charts convention for naming it's listeners (eg onSectionMouseOver) and replaces them with the standard XYPlot ones. It also removes the "listener wedges" from the radial chart in favor of just injecting pointer-events none onto the labels.


docs/docs-app/constants/pages.js
```diff
@@ -30,8 +30,8 @@ export const examplePages = generatePath([
     children: [{
       name: 'Showcase',
       content: {
-        pageType: 'example',
         markdown: getDocUrl('examples/showcase.md'),
+        pageType: 'example',
```
btford: Why move it here but not for below? I assume you're alphabetizing?

docs/markdown/label-series.md
```diff
@@ -0,0 +1,92 @@
...
+## Data format Reference
+
+Like other series the `labelSeries` requires the data be formatted as an array of objects with several required keys and several options ones. Here's an example
+
+
+```javascript
+const myData = [
+  {x: 0, y: 0, label: 'woah!', style={fontSize: 10}},
+  {x: 1, y: 0, label: 'dope city', yOffset: 5},
+  {x: 0, y: 1, label: 'cool Dog friend', xOffset: 5}
```
btford: cool dog friend indeed! 🐕

docs/markdown/label-series.md
```diff
@@ -0,0 +1,92 @@
...
+```javascript
+<XYPlot width={300} height={300}>
+  <LabelSeries
+    animation
+    allowOffsetToBeReversed
+    data={data} />
+</XYPlot>
+```
+
+Each series corresponds to exactly **one** svg path. It is perfectly okay to many series to express many polygons!
```
btford: `to many series to express` is a bit clunky. Maybe `with many series` or `for many series,`.

docs/markdown/label-series.md
```diff
@@ -0,0 +1,92 @@
...
+
+```javascript
+const myData = [
+  {x: 0, y: 0, label: 'woah!', style={fontSize: 10}},
+  {x: 1, y: 0, label: 'dope city', yOffset: 5},
+  {x: 0, y: 1, label: 'cool Dog friend', xOffset: 5}
+]
+```
+
+Would render three points with labels as suggested!
```
btford: here's an `example... would render...` seems like a bit of a fragment.

docs/markdown/label-series.md
```diff
@@ -0,0 +1,92 @@
...
+### allowOffsetToBeReversed (optional)
+The allows the offset specified on the data rows to flipped if the label is too close to an edge. This allows you to make sure your labels never get randomly clipped by going offscreen.
+
+#### className (optional)
+Type: `string`
+Provide an additional class name for the series.
+
+#### data
+Type: `Array<Object>`
+Array of data for the series. See above data format reference.
```
btford: Is it possible to link to `see above data format`?

docs/markdown/label-series.md
```diff
@@ -0,0 +1,92 @@
...
+- `index` is the index of the data point in the array of data;
+- `event` is the event object.
+
+#### onSeriesMouseOver (optional)
+Type: `function(d, {event})`  
+`mouseover` event handler for the elements corresponding separate data points. First argument received is, `d`, the relevant data point, and second an object with the only `event` property.
+
+#### onSeriesMouseOut (optional)
+Type: `function(d, {event})`  
+`mouseout` event handler for the elements corresponding separate data points. First argument received is, `d`, the relevant data point, and second an object with the only `event` property.
```
btford: most relevant how? like nearest?

docs/markdown/radial-chart.md
```diff
@@ -1,89 +1,124 @@
 # Radial chart
 
-`RadialChart` is responsible for creating pie and donut charts.
-Note: currently radial chart _does not_ support series and it is not planned in the nearest future.
+`RadialChart` is responsible for creating pie and donut charts. While this kind of chart is easy to overlook as insignificant, intentionally confusing, or almost always replaceable with a treemap; they can be useful for quickly showing small groups. People don't understand angles very well (such is our biology), but over the last hundred years we have seen a lot of pie charts! This has caused us to become intimately familiar with them.
```
btford: Great intro! Would be really cool to link to resources or articles explaining why/how we know (biologically) that we don't understand angles well.

src/styles/radial-chart.scss
```diff
@@ -1,10 +1,16 @@
...
 
-.rv-radial-chart__series--pie {
-  > path {
-    stroke: #fff;
-    stroke-width: 1px;
+  .rv-xy-plot__series--label {
+    pointer-events: none;
   }
 }
+// .rv-radial-chart__svg {
```
btford: why is this commented out rather than removed?

src/utils/scales-utils.js
```diff
@@ -268,12 +268,14 @@ function _collectScaleObjectFromProps(props, attr) {
...
-    [`${attr}Type`]: type = LINEAR_SCALE_TYPE} = props;
+    [`${attr}Type`]: type = LINEAR_SCALE_TYPE,
+    [`${attr}NoFallBack`]: noFallBack
+  } = props;
 
   let {[`${attr}Domain`]: domain} = props;
 
   // Return value-based scale if the value is assigned.
-  if (typeof value !== 'undefined') {
+  if (!noFallBack && typeof value !== 'undefined') {
```
btford: It's better not to write identifiers with a negation in them, otherwise you end up with double negatives in the code like `not noFallBack`. Would be better to reverse the logic and rename `noFallBack` to `hasFallback` or `isFallback`. Also `fallback` is a single word, so no need to upper case the `B`.

tests/components/radial-tests.js
```diff
@@ -28,46 +28,42 @@ test('RadialChart: Basic rendering', t => {
   const pieSlices = $.find('.rv-radial-chart__series--pie__slice').length;
   t.equal(pieSlices, RADIAL_PROPS.data.length, 'should find the same number of slices as data entries');
 
-  const overlaySlices = $.find('.rv-radial-chart__series--pie__slice-overlay').length;
-  t.equal(overlaySlices, RADIAL_PROPS.data.length, 'should find the right number of overlay slices');
-
-  const labels = $.find('.rv-radial-chart__series--pie-label-inner-wrapper').length;
+  const labels = $.find('.rv-xy-plot__series--label-text').length;
```
btford: This might be out of scope (since this occurs in many tests), but I find that it's better to export constants that relate to class names from the implementation, and then import them in the test. Otherwise, you'll have to update the class name in two places if it ever changes.

btford:
(oh also the tests should probably pass before this goes in)

docs/docs-app/constants/pages.js
mcnuttandrew: i started to make a more complex change but decided to move it to a different diff, this is an artifact of me not being able to remember the state before

docs/markdown/label-series.md
mcnuttandrew: i'm not sure how to do anchor tags in the context of react-router?

PR #385, opened 2017-04-13
Add outline mode to mark-series-gl

mcnuttandrew:
Deck.gl supports outlines for circles, so it seems like something we should offer as well!

screen shot 2017-04-13 at 10 07 07 am

### DIFF
--- a/docs/markdown/markseries.md
+++ b/docs/markdown/markseries.md
@@ -112,3 +112,15 @@ Type: `function(info)`
 
 #### animation (optional)  
 See the [XYPlot](xy-plot.md)'s `animation` section for more information.
+
+## MarkSeriesGL API Additions
+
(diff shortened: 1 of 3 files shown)


docs/markdown/markseries.md
```diff
@@ -112,3 +112,15 @@ Type: `function(info)`
 
 #### animation (optional)  
 See the [XYPlot](xy-plot.md)'s `animation` section for more information.
+
+## MarkSeriesGL API Additions
+
+In addition to the above api the GL version of markSeries exposes several additional props.
+
+#### seriesId
+Type: `string` (REQUIRED)
```
Apercu: maybe put the requireds next to the prop name to match how optionals are displayed too?

PR #389, opened 2017-04-13
Canvas Mode

mcnuttandrew:
While working with the new gl layers i found that it got kinda cumbersome to try to have many layers at once. Rather than introducing some type of elaborate optimization, I figured i'd try out building up the same structure in canvas. This PR introduces three new series:

- MarkSeriesCanvas
- LineMarkSeriesCanvas
- LineSeriesCanvas

And one canvas wrapper. The canvas wrapper lives in the XYPlot, and makes use of the children of xyplot who are canvas series

canvas

There's still a couple of things to do

- [ ]  Docs
- [ ]  Better support for curves
- [ ]  Maybe animation?


akre54:
lgtm 👌 

I wonder if we could follow the example of d3-shape to use something like [d3-path](https://github.com/d3/d3-path) or [react-art](https://github.com/reactjs/react-art), essentially making the renderer draw-mode agnostic behind a buffer abstraction, instead of having to maintain multiple draw instructions across different components for each chart type.

In my head I'm thinking we'd have only one component type per series type which would return a react descriptor, and then the renderer would deal with creating the svg, canvas, webgl, etc. How feasible does this sound?

mcnuttandrew:
> I wonder if we could follow the example of d3-shape to use something like d3-path or react-art, essentially making the renderer draw-mode agnostic behind a buffer abstraction, instead of having to maintain multiple draw instructions across different components for each chart type.

I think that sounds like a really beautiful idea, but I feel like deck.gl has a different enough interface where we can't quite treat them completely homogeneously.

>In my head I'm thinking we'd have only one component type per series type which would return a react descriptor, and then the renderer would deal with creating the svg, canvas, webgl, etc. How feasible does this sound?

 I think your totally right having a single LineSeries, that accepts a renderer prop from ['svg', 'gl',  'canvas'] would be really cool. I'm not quite sure how to implement that give the way that XYPlot interfaces with it's children (in that directly checks static methods present on those children). 

if you wanna give it a go, i'd be hyped to see what you find?

src/plot/series/canvas-wrapper.js
```diff
@@ -0,0 +1,78 @@
...
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+
+class CanvasWrapper extends Component {
+  componentDidMount() {
+    this.drawChildren(this.props, this.refs.canvas.getContext('2d'));
+  }
+
+  componentWillUpdate(nextProps) {
+    this.drawChildren(nextProps, this.refs.canvas.getContext('2d'));
```
mcnuttandrew: i am frankly not sure! i feel like caching it might lead some weird sync-ing issue, but my grasp of this part of props is not the strongest

src/plot/series/line-series-canvas.js
```diff
@@ -0,0 +1,80 @@
...
+import {getAttributeFunctor, getAttributeValue} from 'utils/scales-utils';
+
+import AbstractSeries from './abstract-series';
+
+class LineSeriesCanvas extends AbstractSeries {
+  static get requiresSVG() {
+    return false;
+  }
+
+  static get isCanvas() {
```
mcnuttandrew: no, i was just following the pattern that was use in other places?

PR #390, opened 2017-04-14
Update radial-chart.md

dmitry-zaets:
- Fix typo `label` -> `subLabel`


mcnuttandrew:
Important to note that those docs refer to the new radial chart, which is available under 2.0.0-beta

PR #393, opened 2017-04-21
feat(core): allow <script> usage with browserify

Apercu:
Also fix gitattributes :smile:


jckr:
that's going to let us easily use react-vis in codepen, etc.

Issue #230, opened 2017-01-13
Rounding error on x-axis values when they are small decimals


timfpark:
Using an XYPlot with formatted tick values (ala tickFormat={v => `${v} mi.`}) and a small range of values (for example, the below where the mileage runs between 0 and 3.1 miles yielding decimal increments of 0.2 miles on the x axis causes occasionally a bug where these tick values are not rounded correctly and instead contain a very large amount of precision that wrecks the visual display.

<img width="599" alt="screen shot 2017-01-12 at 4 48 59 pm" src="https://cloud.githubusercontent.com/assets/712001/21914287/187b75bc-d8e7-11e6-8875-3523db9fb043.png">

I've dug into the code but can't seem to figure out where the rounding needs to be corrected - i am happy to investigate if someone can give me a pointer on where to look.

Thanks for a great library overall!

mcnuttandrew:
Hey @timfpark 

This is an interesting bug! Ill investigate. In the mean time you could inject the degree of precision you are looking for by modifying your custom formatter a little bit more. Specifically, if you import d3-format, and then build a formtter like

`var f = d3.format(".1f");`

which you could then use in conjunction with your tickFormat prop:

`tickFormat={v => `${f(v)} mi.`}

that would probably do what you are looking for

Apercu:
I think anton was working on this particular issue

mcnuttandrew:
It's been a long time since we've had any word on this, I'm going to close and if anyone is interested in reopening it (ie if it comes up again) we can see about it then

Issue #397, opened 2017-04-26
opacity behavior


jckr:
Right now I can't have an opacity set at the level of a series, unless I also set opacityType to be "category" which is really counter-intuitive:

ie <MarkSeries opacity="0.2" data={data}/> -> won't take opacity into account
<MarkSeries opacity="0.2" opacityType="category" data={data}/> -> all marks are now 20% opaque

And if that wasn't cumbersome enough, if I do that, I won't be able to pass opacity values through the data. (I mean I could, but the opacity scale will still be categorical, so it's impossible to pass values directly). 

That's linked to https://github.com/uber/react-vis/pull/275 and changes on the scale-utils. PR coming up

mcnuttandrew:
I think this was addressed in #398

PR #398, opened 2017-04-26
fixed opacity behavior, added tests

jckr:
this is re https://github.com/uber/react-vis/issues/397

### DIFF
--- a/showcase/plot/scatterplot.js
+++ b/showcase/plot/scatterplot.js
@@ -42,6 +42,7 @@ export default class Example extends React.Component {
         <MarkSeries
           className="mark-series-example"
           strokeWidth={2}
+          opacity="0.8"
           sizeRange={[5, 15]}
           data={[
             {x: 1, y: 10, size: 30},
--- a/src/utils/scales-utils.js
+++ b/src/utils/scales-utils.js
@@ -111,7 +111,7 @@ const SCALE_FUNCTIONS = {
  * @private
  */
 export function _getSmallestDistanceIndex(values, scaleObject) {
-  const scaleFn = _getScaleFnFromScaleObject(scaleObject);
+  const scaleFn = getScaleFnFromScaleObject(scaleObject);
   let result = 0;
   if (scaleFn) {
     let nextValue;
@@ -142,7 +142,7 @@ export function _getSmallestDistanceIndex(values, scaleObject) {
  * @returns {*} Scale function.
  * @private
  */
-export function _getScaleFnFromScaleObject(scaleObject) {
+export function getScaleFnFromScaleObject(scaleObject) {
   if (!scaleObject) {
     return null;
   }
@@ -208,9 +208,9 @@ export function getDomainByAttr(allData, attr, type) {
  * @private
  */
 function _createScaleObjectForValue(attr, value, type) {
-  if (type === 'literal') {
+  if (type === LITERAL_SCALE_TYPE) {
     return {
-      type: 'literal',
+      type: LITERAL_SCALE_TYPE,
       domain: [],
       range: [value],
       distance: 0,
@@ -223,7 +223,7 @@ function _createScaleObjectForValue(attr, value, type) {
     return null;
   }
   return {
-    type: 'category',
+    type: CATEGORY_SCALE_TYPE,
     range: [value],
     domain: [],
     distance: 0,
@@ -378,7 +378,7 @@ export function _getScaleDistanceAndAdjustedDomain(data, scaleObject) {
     adjustedDomain[0] = Math.min(domain[1] / 10, 1);
   }
 
-  const adjustedScaleFn = _getScaleFnFromScaleObject({
+  const adjustedScaleFn = getScaleFnFromScaleObject({
     ...scaleObject,
     domain: adjustedDomain
   });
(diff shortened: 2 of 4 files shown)


tests/utils/axis-utils-tests.js
```diff
@@ -32,7 +32,7 @@ test('axis-utils #getTicksTotalFromSize', t => {
 
 test('axis-utils #getTickValues', t => {
   const scale = scaleLinear().domain([0, 1]).range(['red', 'blue']);
-  t.deepEqual(getTickValues(scale, 10, false),
+  t.deepEqual(getTickValues(scale, 10, false).map(d => Math.round(d * 1000) / 1000),
```
mcnuttandrew: I know @apercu introduced changes recently that caused this test to have these values, rounding them invalidates the value of this test

tests/utils/scales-utils-tests.js
```diff
@@ -97,6 +97,10 @@ test('scales-utils #getAttributeFunctor', t => {
     result({x: Math.random()}) === _xValue,
     `No props: Fallback value ${_xValue} should be returned by the produced functor`
   );
+  t.ok(
```
mcnuttandrew: prefer t.equal, as it will provide a more informative error

src/utils/scales-utils.js
```diff
@@ -150,6 +150,9 @@ export function _getScaleFnFromScaleObject(scaleObject) {
   const modDomain = domain[0] === domain[1] ?
     (domain[0] === 0 ? [-1, 0] : [-domain[0], domain[0]]) :
     domain;
+  if (type === LITERAL_SCALE_TYPE) {
```
mcnuttandrew: add a test for _getScaleFnFromScaleObject if you are modifying it?

src/utils/scales-utils.js
```diff
@@ -656,8 +659,11 @@ export function getMissingScaleProps(props, data, attributes) {
  * Return a d3 scale that returns the literal value that was given to it
  * @returns {function} literal scale.
  */
-export function literalScale() {
+export function literalScale(defaultValue) {
```
mcnuttandrew: add test for literalScale if you are modifying it

showcase/plot/scatterplot.js
```diff
@@ -42,6 +42,7 @@ export default class Example extends React.Component {
         <MarkSeries
           className="mark-series-example"
           strokeWidth={2}
+          opacity="0.8"
```
mcnuttandrew: why "0.8" and not {0.8} ?

tests/utils/axis-utils-tests.js
jckr: I was with @apercu when he changed the values to [0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1] from [0, 0.1, 0.2, 0.30000000000000004, 0.4, 0.5, 0.6000000000000001, 0.7000000000000001, 0.8, 0.9, 1], because that's what the operation in the test returned then. Lo and behold, a few days later, the same operation yields [0, 0.1, 0.2, 0.30000000000000004, 0.4, 0.5, 0.6000000000000001, 0.7000000000000001, 0.8, 0.9, 1].

What will it be tomorrow? your guess is as good as mine because not only is this a highly unpredictable float division, but one which depends on how d3 implements it. 

for those reasons, I think that comparing the result to either [0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1] or [0, 0.1, 0.2, 0.30000000000000004, 0.4, 0.5, 0.6000000000000001, 0.7000000000000001, 0.8, 0.9, 1] is going to fail eventually. 
comparing to [0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1] with a tolerance of ±0.001 which is what this codes does ensures that the behavior we want is right, independently of the exact output of the float divisions.

showcase/plot/scatterplot.js
jckr: I guess 0.8 (number) makes sense from a functional point of view. Then again, opacity is stored in the CSSStyleDeclaration object as a string. So imo neither or them is vastly superior to the other.

tests/utils/axis-utils-tests.js
mcnuttandrew: :( I was hoping that balthazar's change was related to https://twitter.com/mbostock/status/853254045477126144 

(which in turn i was hoping had resolved https://github.com/uber/react-vis/issues/230), but alas. I think tick values should probable just accepts another argument called precision or something

src/utils/scales-utils.js
```diff
@@ -656,8 +659,11 @@ export function getMissingScaleProps(props, data, attributes) {
...
   function scale(d) {
+    if (d === undefined) {
+      return defaultValue;
```
mcnuttandrew: what if default value is undefined? it should still return undefined right? Add a test case?

tests/utils/axis-utils-tests.js
chrisirhc: Looks like this was fixed in d3-array 1.2.0 . Perhaps we need to update our dependencies to 1.2 , and update d3-scales to use d3-array 1.2 .

src/utils/scales-utils.js
jckr: yes. we can def add a test case.

mcnuttandrew:
Any update on this?

mcnuttandrew:
@jckr any update on this guy? it's been open for a while

PR #401, opened 2017-04-26
Remove all deck.gl dependencies

javidhsueh:
Deck.gl is heavy and we are not sure about having multiple GL charts at the same time will degrade the performance of not.  Also, it creates the dependency conflicts when having other modules use different version of deck.gl.


docs/markdown/line-series.md
```diff
@@ -2,15 +2,6 @@
 
 <!-- INJECT:"LineChart" -->
 
-react-vis offers two different types of LineSeries, one that renders SVG and one that renders WebGL.
```
mcnuttandrew: This can stay but be updated to discuss canvas

docs/markdown/markseries.md
```diff
@@ -31,15 +31,6 @@ const myData = [
 ]
 ```
 
-react-vis offers two different types of MarkSeries, one that renders SVG and one that renders WebGL.
```
mcnuttandrew: same comment as above about canvas

showcase/index.js
```diff
@@ -83,7 +81,6 @@ import TreemapExample from './treemap/dynamic-treemap';
 export const showCase = {
   ComplexChart,
   LineChart,
-  LineChartGL,
```
mcnuttandrew: maybe don't delete this demos? they are still good for testing the canvas stuff. Just remove the deck.gl content

Apercu:
I agree that we should keep it. The 2.0.0 beta on master was not necessarily a good idea, but we shouldn't remove it for all. I'm more thinking about a separate build of react-vis integrating it, while the default would be vanilla. And deck.gl is pretty stable now.

samhogg:
Instead of ripping deck.gl out completely, could we make the rendering API pluggable?

Something along the lines of splitting into svg, canvas, and gl implementations but keep the user facing chart API consistent. Also has the advantage that react native implementations could be added, too.

chrisirhc:
This isn't at all the reason we're removing it. @javid is revising the description in a bit.

We're removing deck.gl from the main core react-vis package because having deck.gl as a dependency creates dependency conflicts with any existing project that uses deck.gl (and deck.gl will not run when there's multiple versions of it installed in the same project). The plan is still to move it to a separate repository such as react-vis-addons-gl .

javidhsueh:
Apologize for the previous wording.
I agree to have deck.gl to power the GL charts, but it should be something pluggable by the users. 
As I mentioned, having deck.gl in react-vis introduces the dependency conflicts and not be able to solve at the user end. Also, having deck.gl in react-vis requires all the users to bump the node version to v6.  According to these reasons, I think we shouldn't have deck.gl in react-vis for now. 

It's nice to have deck.gl to power all GL charts but we should have a better way to plugin it.

mcnuttandrew:
>  According to these reasons, I think we shouldn't have deck.gl in react-vis for now.

Fair enough! It seems reasonable to me to pull out the GL stuff into a different repo given that react-vis has it's own raster (canvas) methods, which should be sufficient for now. Like i pointed out in my comment i think a number of things you deleted can be kept around and made to just highlight the difference between canvas and svg?

javidhsueh:
Agree. I'll add those back and update the doc.

ibgreen:
While the deck.gl team would of course love to see deck.gl continue to be available in react-vis, we are also very interested in feedback on problems.

>  It should be something pluggable by the users.

I can't really disagree. We have done a lot to trim the size and number of dependencies but deck.gl is still obviously a big thing to include for apps that are not using it. Also, while deck.gl itself supports tree shaking, support in bundlers is still "shaky" and a lot of folks are still using browserify. So I can understand the wish to package it as a complementary module. I'd personally recommend against splitting repositories as I feel that would fragment our tech stack.

> having deck.gl in react-vis introduces the dependency conflicts and not be able to solve at the user end

In principle, the deck.gl dependency issues are no different than React's dependency issues that you are already dealing with - they are usually handled by having React as a peer dependency and the same thing works for deck.gl. But I understand that it is undesirable to have a long list of peer dependencies for react-vis, especially for something that is not a core function.

> Also, having deck.gl in react-vis requires all the users to bump the node version to v6. 

deck.gl intentionally requires Node 6 when **building** deck.gl. But is that also true when installing and using deck.gl? Maybe something we could/should fix? Pretty sure we wouldn't want to support Node < 4 though.

@shaojingli @Pessimistress

chrisirhc:
> I'd personally recommend against splitting repositories as I feel that would fragment our tech stack.

I like the idea of moving it into a separate build/package within the same repository. That's how react creates/maintains its addons as well. However, this does add to the complexity of the build/package process.

I agree that moving the code into a separate repository is definitely sub-optimal and a makeshift solution. Perhaps we can keep the deck.gl code in a branch in the meantime until we find time to re-look into the build/package process?

docs/markdown/line-series.md
```diff
@@ -2,14 +2,11 @@
...
-`LineSeriesGL` instead of `LineSeries`. It is most effective to hover over your points using onNearestX rather
-than onValueMouseOver.
+react-vis offers two different types of LineSeries, one that renders SVG and one that renders Canvas.
+The SVG mode is accessed by using the normal `LineSeries`, just as above, while the Canvas mode is used by simply calling
+`LineSeriesCanvas` instead of `LineSeries`.
 
-<!-- INJECT:"LineChartGL" -->
-
-**NOTE**: using the GL version of this layer disables animation
+<!-- INJECT:"LineCanvasGL" -->
```
mcnuttandrew: this name seems wrong to me

docs/markdown/markseries.md
```diff
@@ -31,14 +31,11 @@ const myData = [
...
-The SVG mode is accessed by using the normal `MarkSeries`, just as above, while the WebGL mode is used by simply calling
-`MarkSeriesGL` instead of `MarkSeries`. Important tip! It is most effective to hover over your points using onNearestXY rather
-than onValueMouseOver.
+react-vis offers two different types of MarkSeries, one that renders SVG and one that renders Canvas.
+The SVG mode is accessed by using the normal `MarkSeries`, just as above, while the Canvas mode is used by simply calling
+`MarkSeriesCanvas` instead of `MarkSeries`. 
 
-<!-- INJECT:"ScatterplotGLChart" -->
-
-**NOTE**: using the GL version of this layer disables animation
```
mcnuttandrew: this note is still valid but about canvas

src/styles/examples.scss
```diff
@@ -382,14 +382,3 @@ article {
     stroke-width: 2px !important;
   }
 }
-
-.scatterplot-gl-example-controls {
```
mcnuttandrew: these classes are still used in the examples?

javidhsueh:
react-vis_example

docs/markdown/markseries.md
```diff
@@ -31,14 +31,11 @@ const myData = [
...
 ```
 
-react-vis offers two different types of MarkSeries, one that renders SVG and one that renders WebGL.
-The SVG mode is accessed by using the normal `MarkSeries`, just as above, while the WebGL mode is used by simply calling
-`MarkSeriesGL` instead of `MarkSeries`. Important tip! It is most effective to hover over your points using onNearestXY rather
-than onValueMouseOver.
+react-vis offers two different types of MarkSeries, one that renders SVG and one that renders Canvas.
+The SVG mode is accessed by using the normal `MarkSeries`, just as above, while the Canvas mode is used by simply calling `MarkSeriesCanvas` instead of `MarkSeries`.
 
 <!-- INJECT:"ScatterplotGLChart" -->
```
mcnuttandrew: I don't mean to keep you in endless revisions here, but ScatterplotGLChart should probably not be called that? It should ScatterplotCanvas or something? Same story with LineChartGl => LineChartCanvas.

mcnuttandrew:
I think this should be fine to land now, stamp

Apercu:
There seems to be a bug with the merging capability on this pull request, never saw that before.
I'll merge manually to master.

PR #403, opened 2017-04-27
Fix discrete color legend

jckr:
DiscreteColorLegend has a colors default property, which should be used to pass a pass a palette (ie an array of colors). instead, it uses DISCRETE_COLOR_RANGE regardless. This is especially annoying given that DISCRETE_COLOR_RANGE is not directly accessible by users. 
in this PR, DiscreteColorLegend can take the colors property into account. 
Added an example.

### DIFF
--- a/showcase/legends/horizontal-discrete-custom-palette.js
+++ b/showcase/legends/horizontal-discrete-custom-palette.js
@@ -1,0 +1,54 @@
+// Copyright (c) 2016 - 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the 'Software'), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


src/legends/discrete-color-legend.js
```diff
@@ -49,27 +49,28 @@ const defaultProps = {
...
         item.color :
-        DISCRETE_COLOR_RANGE[i % DISCRETE_COLOR_RANGE.length],
+        colors[i % colors.length],
       disabled: Boolean(item.disabled)
     };
   });
 }
 
 function DiscreteColorLegend({
+  colors,
```
mcnuttandrew: maybe make DISCRETE_COLOR_RANGE the default for colors?

showcase/legends/horizontal-discrete-custom-palette.js
```diff
@@ -0,0 +1,54 @@
...
+  '#6588cd',
+  '#66b046',
+  '#a361c7',
+  '#ad953f',
+  '#c75a87',
+  '#55a47b',
+  '#cb6141'
+];
+
+export default function DiscreteColorExample() {
```
mcnuttandrew: Consider adding a note about this in the documentation, and including a test for this example.

src/legends/discrete-color-legend.js
```diff
@@ -49,27 +49,28 @@ const defaultProps = {
   orientation: 'vertical'
 };
 
-function fillItemsWithDefaults(items) {
+function fillItemsWithDefaults(items, colors) {
```
mcnuttandrew: add a jsdoc?

src/legends/discrete-color-legend.js
jckr: it's in the defaultProps already

mcnuttandrew: so it is (disappearing-homer)

showcase/legends/horizontal-discrete-custom-palette.js
jckr: Funny that you mention the documentation, because I have a local branch where I'm rewriting it extensively. The reason why I'm doing PRs is when I encounter quirks or things that don't work as I thought they would/should

mcnuttandrew: that's splendid to hear! I guess just add a test then

mcnuttandrew:
Have you also confirmed that this works for the searchable version of the legend?

mcnuttandrew:
fix lint? otherwise stamp

Issue #407, opened 2017-05-03
PolygonSeries leaks over XAxis Labels


Smarq:
Hello,

I am having an issue with the PolygonSeries running over the XAxis labels. Here is my code: 
```
<XYPlot
     xType="linear"
     width={800}
     height={500}
     yDomain={[minPrice - stdDeviation, maxPrice + stdDeviation]}
     margin={{ left: 100, right: 100 }}>
     <PolygonSeries
          data={[
                { x: minMileage, y: minPrice - stdDeviation },
                { x: minMileage, y: maxPrice + stdDeviation },
                { x: maxMileage, y: maxPrice + stdDeviation },
                { x: maxMileage, y: minPrice - stdDeviation }
         ]}
         color="#f5f5f5" />
     <PolygonSeries
            data={getPriceBand(slope, yIntercept, minMileage, maxMileage, stdDeviation, 3.2, 2)}
            color="#f3b8af" />
     <PolygonSeries
            data={getPriceBand(slope, yIntercept, minMileage, maxMileage, stdDeviation, -2, -3.2)}
            color="#d5b5f2" />
     <VerticalGridLines tickTotal={5} />
     <HorizontalGridLines tickTotal={5} />
     <LineSeries
         data={ targetLine }
         color="darkGreen"/>
     <MarkSeries
             data={data}
             color="redOrange">
     </MarkSeries>
     <XAxis />
     <YAxis />
</XYPlot>
```
Which produces the following screenshot:

![screen shot 2017-05-03 at 9 54 44 am](https://cloud.githubusercontent.com/assets/2789511/25666589/966c17fa-2fe6-11e7-8804-061c258bc34d.png)

Ultimately, I believe that changing the yDomain is causing this issue. The feature I am trying to implement is a zoom and I was hoping to be able to do that using xDomain and yDomain.

LoicUV:
I have a similar problem using `xDomain` with a line & area series where the data overflows the axis.
I think applying a domain should strip the data that is outside of the domain.

![overflow](https://cloud.githubusercontent.com/assets/592542/25706492/e5b51d0c-30e0-11e7-959b-e303e65a307e.png)

mcnuttandrew:
Hey @Smarq 

You've hit upon a complex issue! While it would make sense for the chart to crop anything outside of it's innerWidth/innerHeight, that can lead to weird looking side effects like dots being cut in half etc.

[...]

PR #412, opened 2017-05-08
Add borders component

mcnuttandrew:
When domain is set on the XYPlot the series that it contains can sometimes spill over into margin. One potential solution would be to not allow overflow, however this can be limiting because of the relatively flat structure of the XYPlot. The most versatile solution (imo) is to create a Borders component that is directly correlated to the margins of the plot. That way the user can embed it into the structure of their chart however they wish. 

screen shot 2017-05-08 at 12 31 44 pm

This addresses issue #407

### DIFF
--- a/docs/markdown/borders.md
+++ b/docs/markdown/borders.md
@@ -1,0 +1,54 @@
+## Borders
+
+Sometimes when modifying the domain of the XYPlot it can be useful to enforce a border, so that some components appear, and others do not. One way to do this is to use the `Borders` component. It is a simple component that creates rectangles the directly correspond to the margins of the plot.
+
+<!-- INJECT:"GradientExample" -->
+
+For example, a set up like this:
+
+```jsx
+<XYPlot xDomain={[1.2, 3]} yDomain={[11, 26]} width={300} height={300}>
+  <AreaSeries
+    data={[
+      {x: 1, y: 10, y0: 1},
+      {x: 2, y: 25, y0: 5},
+      {x: 3, y: 15, y0: 3}
+    ]}/>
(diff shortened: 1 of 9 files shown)


jckr:
this solution requires to overlay white (?) rectangles at the precise right place to obtain the desired effect.
In both SVG and canvas however, the most natural way to get that behavior is to use clipping. 

Recently, we are moving in the <defs> space by using gradients.
I feel we should add clipping functionality to our series in general. 

Here's how it could work. 
we give our series 2 new props: id and clippedBy. 
if a series has an id, we are going to create a clipPath in the defs of our top svg element with that ID. 
then, if another series has the clippedBy attribute, we apply the right clip-path attribute. 

In canvas, if a series has a clippedBy attribute, we would create a shape corresponding to that other series, then clip() instead of closePath(). Then render the original series. 

in addition, we could create a default clipping mask with id "chartArea" that would be a rectangle that corresponds to margins.

thoughts?

mcnuttandrew:
I think the behaviour you are suggesting is already supported by the gradientDefs component! Although we maybe also need to expose an id prop on all the series(?). I think that could maybe be described in a tutorial rather than upgrading the components themselves? 

That all said I still think having the borders component approach is good because it is really really simple:
- it doesn't require the user to know about clip paths, which can get hairy 
- it doubles down on an idea they are already familiar with (the margins)
- it follows the composable children api already in place 

I think clip path stuff is really good, but is more advanced/complicated. So why not both?

mcnuttandrew:
@apercu @jckr @btford, anyone wanna throw a review on this?

Issue #342, opened 2017-04-03
Custom MarkSeries component


LoicUV:
Would you consider adding a custom component prop to `<MarkSeries />` to be used in place of its default `<circle />` when present ? Something like : 

```javascript
    return (
      <g className={`${predefinedClassName} ${className}`}
         ref="container"
         transform={`translate(${marginLeft},${marginTop})`}>
        {data.map((d, i) => {
          const attrs = {
            r: sizeFunctor ? sizeFunctor(d) : DEFAULT_SIZE,
            cx: xFunctor(d),
            cy: yFunctor(d),
            style: {
              opacity: opacityFunctor ? opacityFunctor(d) : DEFAULT_OPACITY,
              stroke: strokeFunctor && strokeFunctor(d),
              fill: fillFunctor && fillFunctor(d),
              strokeWidth: strokeWidth || DEFAULT_STROKE_WIDTH
            },
            key: i,
            onClick: e => this._valueClickHandler(d, e),
            onMouseOver: e => this._valueMouseOverHandler(d, e),
            onMouseOut: e => this._valueMouseOutHandler(d, e)
          };
          return CustomComponent ? <CustomComponent {...attrs} /> : <circle {...attrs} />;
        })}
      </g>
    );
```

mcnuttandrew:
I think this sounds like a reasonable PR! I see two ways of doing this, either by modifying the markseries, as you have above, or making a new series type, something like CustomMarkSeries. If you do a new series you won't disrupt the API of the old series, and you could add additional built in marktypes (like boxes diamonds etc).

Also just to make sure that the component is easy to use, i'd probably write the contents of your return (either the circle or custom component) that translates to the correct x/y coordinates, so that the user doesn't have to worry about functors and what not

mcnuttandrew:
This enhancement is now in! Marking as resolved.

PR #415, opened 2017-05-11
Allow a markType to be passed to MarkSeries to enable square marks

akre54:


akre54:
True, and that would allow images or compound components. My main use case was to allow this to be easily passed down on the various layered components (LineMarkSeries, e.g.).

Maybe the customMark could be a prop function that takes `attrs` and `d`. I'll whip something up.

mcnuttandrew:
Also rel, #342

PR #420, opened 2017-05-12
Upgrade docs

mcnuttandrew:


mcnuttandrew:
(Note this is just deploying already stamped docs)

PR #423, opened 2017-05-15
Allow a markType function to be passed to MarkSeries to enable custom marks

akre54:
Second try at #415. This is a smaller surface area API change but is potentially more powerful.

The interface is still a little tricky since the prop sizes and x and y positions are still tied too closely to circle coordinates (see example), but open to any suggestions.

```js
markType={({attrs: {size, x, y, ...attrs}, d}) => {
  return <rect {...attrs} height={size * 2} width={size * 2} x={x - size} y={y - size} />;
}}
```

### DIFF
--- a/src/plot/series/mark-series.js
+++ b/src/plot/series/mark-series.js
@@ -62,10 +62,11 @@ class MarkSeries extends AbstractSeries {
          ref="container"
          transform={`translate(${marginLeft},${marginTop})`}>
         {data.map((d, i) => {
+          const size = sizeFunctor ? sizeFunctor(d) : DEFAULT_SIZE;
+          const x = xFunctor(d);
+          const y = yFunctor(d);
+
           const attrs = {
-            r: sizeFunctor ? sizeFunctor(d) : DEFAULT_SIZE,
-            cx: xFunctor(d),
-            cy: yFunctor(d),
             style: {
               opacity: opacityFunctor ? opacityFunctor(d) : DEFAULT_OPACITY,
               stroke: strokeFunctor && strokeFunctor(d),
@@ -80,9 +81,9 @@ class MarkSeries extends AbstractSeries {
           };
 
           if (typeof markType === 'function') {
-            return markType({attrs, d});
+            return markType({attrs: {...attrs, size, x, y}, d});
           }
-          return <circle {...attrs} />;
+          return <circle {...attrs} cx={x} cy={y} r={size} />;
         })}
       </g>
     );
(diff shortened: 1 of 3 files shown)


src/plot/series/mark-series.js
```diff
@@ -78,7 +79,11 @@ class MarkSeries extends AbstractSeries {
             onMouseOver: e => this._valueMouseOverHandler(d, e),
             onMouseOut: e => this._valueMouseOutHandler(d, e)
           };
-          return <circle {...attrs} />;
+
+          if (typeof markType === 'function') {
```
mcnuttandrew: maybe just check for presence rather than typeof? it seems like the prop-types should type check that?

src/plot/series/mark-series.js
```diff
@@ -62,10 +62,11 @@ class MarkSeries extends AbstractSeries {
          ref="container"
          transform={`translate(${marginLeft},${marginTop})`}>
         {data.map((d, i) => {
+          const size = sizeFunctor ? sizeFunctor(d) : DEFAULT_SIZE;
+          const x = xFunctor(d);
```
mcnuttandrew: on going nit: we have been using the word functor incorrectly in react-vis (it is not a function-object), and so should seek to just call it the variable, eg x instead of xFunctor

tests/components/mark-series-tests.js
```diff
@@ -16,6 +16,14 @@ test('MarkSeries: Showcase Example - Scatterplot', t => {
   t.end();
 });
 
+test('MarkSeries: Showcase Example - Scatterplot with custom markType', t => {
```
mcnuttandrew: + 1 for tests
- 1 for no docs

mcnuttandrew:
Also, this is addressing #342

akre54:
Whups just saw this! Responding in kind

src/plot/series/mark-series.js
akre54: Agreed. Let's open an issue for this.

src/plot/series/mark-series.js
jckr: do you actually need a function? because then, you'd have to make sure the interaction handlers and all the stuff we may add in the future to scatterplot marks work.
my assumption is that your marks could all be paths. i mean even the defaut marks, circles, could be paths. and size can be dealt with with "transform: scale(x)". so the argument you'd pass would just be an SVG style path?

tests/components/mark-series-tests.js
jckr: likewise. this has to include changes to docs.

mcnuttandrew:
Closing this PR as #517 handles the task at hand

PR #425, opened 2017-05-21
feat(sankey): add labels support

Apercu:
screen shot 2017-05-21 at 12 02 02 am

Display by default the `name` property of the nodes as labels.
Also provides a `hideLabels` property, turned off by default


mcnuttandrew:
Given that this a top level chart type, consider using xyplot and label series to make this happen? 

Add tests and docs

PR #432, opened 2017-05-25
Radar chart

mcnuttandrew:
This PR adds a new basic chart type: the long awaited radar chart (#299). Couple of small things still to do, but i think it should be ready for review.

- [x] Docs
- [x] Address Todos

radar

Open to any suggestions on how to improve the API. Right now we simply ask users to provide rows of polygons for the data prop, and row of domains, like so 

```javascript
const RADAR_PROPS = {
  data: [{
    explosions: 7,
    wow: 10,
    dog: 8,
    sickMoves: 9,
    nice: 7
  }],
  domains: [
    {name: 'nice', domain: [0, 100]},
    {name: 'explosions', domain: [6.9, 7.1]},
    {name: 'wow', domain: [0, 11]},
    {name: 'dog', domain: [0, 16]},
    {name: 'sickMoves', domain: [0, 20]}
  ],
  height: 300,
  width: 400
};
```
Where the order of the domains provide the order of the axes. Here are some questions:
- It doesn't accept children but it could! I am unsure if that's a good idea
- I am not sure how tooltip-ing would work, but it's definitely something that should be considered.
- Some radars feature marks on the vertices, should we support that feature?

### DIFF
--- a/src/radar-chart/index.js
+++ b/src/radar-chart/index.js
@@ -1,0 +1,192 @@
+// Copyright (c) 2016 - 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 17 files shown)


src/radar-chart/index.js
```diff
@@ -0,0 +1,192 @@
...
+
+const predefinedClassName = 'rv-radar-chart';
+const DEFAULT_FORMAT = format('.2r');
+// TODO add jsdoc
+function getAxes({domains, animation, style}) {
+  return domains.map((domain, index) => {
+    const angle = index / domains.length * Math.PI * 2;
+    const sortedDomain = domain.domain.sort();
+
+    return (<DecorativeAxis
```
Apercu: opening and closing of the element on their own lines?

zachguo:
Thanks for implementing Radar! But how can user import Radar chart?
I tried `import {Radar} from 'react-vis';` and `import {RadarChart} from 'react-vis';`. Neither works. It seems Radar is not in `index.js`.

mcnuttandrew:
Oop my bad Ill file a PR

zachguo:
Thanks, I'm testing it out. 
Another issue is `tickFormat` is working as a root prop, `domains.tickFormat` has no effect. Documentation needs to be changed.

mcnuttandrew:
will you file an issue for the bugs your finding with radar? little easier to track than on a closed PR

zachguo:
Sorry, I filed two issues.

PR #434, opened 2017-05-30
Move react/react-dom out of devDeps to peerDeps

CharlieMartell:
It is unclear why react/react-dom are listed in devDependencies, and seems to cause errors with multiple react versions when linking react-vis for development.


CharlieMartell:
It seems react-vis lists react as a devDependency in order to run its docs which is still needed. Closing PR but perhaps docs should be moved into its own package.json

PR #435, opened 2017-05-30
Upgrade the docs & Dont Jekyll

mcnuttandrew:
Github was ignoring new updates to the docs page because it's jekyll interpreter was invalidating a bunch of jsx syntax we had (eg something like "style={{fontSize: 14}}", would be invalid because it didn't terminate as a jekyll variable correctly).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,5 @@ react-vis.gif
 npm-debug.log*
 .DS_Store
 .*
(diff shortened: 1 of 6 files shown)


mcnuttandrew:
Landing as these changes have already been stamped in other PRs

Apercu:
Interesting, thanks!

PR #438, opened 2017-05-30
Allow unanimatedProps passed through Axis.js to support not interpola…

CharlieMartell:
Previously, if you had an XYPlot, it would render its axis with hardcoded "animatedProps" meaning if you had anything in your axis-titles they would be ran through the d3-interpolator which is not always intended. 

This PR allows you to pass unanimatedProps and not interpolate some props of an axis via a XYPlot.

### DIFF
--- a/docs/markdown/animation.md
+++ b/docs/markdown/animation.md
@@ -4,11 +4,11 @@ Animation makes your charts feel physical, it makes them feel alive, shoot it ma
 
 *Booleans*: if true is present then `react-vis` will use the `no-wobble` preset (see below)
 
-*Strings*: React-motion offers several different motion presets that cover most use cases. To access them set your animation prop to one of [noWobble, gentle, wobbly, stiff].
+*Strings*: react-motion offers several different motion presets that cover most use cases. To access them set your animation prop to one of [noWobble, gentle, wobbly, stiff].
 
 <!-- INJECT:"AnimationExample" -->
 
-*Objects*: React-motion expects an object formatting like `{damping: NUMBER, stiffness: NUMBER}`, and if you want to give us an object like that, we will hand it direct to react-motion. You can also pass an object with `{nonAnimatedProps: ['foo', 'bar']}` to prevent those props from being interpolated by d3-interpolator.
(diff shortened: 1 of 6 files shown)


CharlieMartell:
That makes sense I hadn't seen the PropTypes for animation being oneOfType allowing for nonAnimatedProps to be contained within animation.

src/plot/axis/axis.js
```diff
@@ -137,9 +137,14 @@ class Axis extends PureComponent {
 
   render() {
 
-    const {animation} = this.props;
+    const {animation, unanimatedProps} = this.props;
+
 
     if (animation) {
+      const animatedProps = animation.nonAnimatedProps ? defaultAnimatedProps.filter(
+        prop => animation.nonAnimatedProps.indexOf(prop) < 0
```
mcnuttandrew: could use find instead of indexof

docs/markdown/animation.md
```diff
@@ -8,7 +8,7 @@ Animation makes your charts feel physical, it makes them feel alive, shoot it ma
 
 <!-- INJECT:"AnimationExample" -->
 
-*Objects*: react-motion expects an object formatting like {damping: NUMBER, stiffness: NUMBER}, and if you want to give us an object like that, we will hand it direct to react-motion.
+*Objects*: React-motion expects an object formatting like `{damping: NUMBER, stiffness: NUMBER}`, and if you want to give us an object like that, we will hand it direct to react-motion. You can also pass an object with `{nonAnimatedProps: ['foo', 'bar']}` to prevent those props from being interpolated by d3-interpolator.
```
mcnuttandrew: Nb I think they spell it lowercase

src/plot/xy-plot.js
```diff
@@ -305,8 +304,8 @@ class XYPlot extends React.Component {
         dataProps = {data: data[seriesIndex]};
       }
       return React.cloneElement(child, {
-        ...dimensions,
         animation,
+        ...dimensions,
```
mcnuttandrew: Why the reorder?

docs/markdown/animation.md
CharlieMartell: Ah I was just trying to make it the same as in +*String*+, I will go ahead and change both.

Issue #419, opened 2017-05-12
Contour Series


mcnuttandrew:
It would be really cool if we could add a contour series! This would allow use to do really expressive heatmaps and contours:

![faithful](https://cloud.githubusercontent.com/assets/6854312/26010076/601471c2-3700-11e7-84e5-17175e2dfbf4.png)

One way that this could be done is by bostock's wonderful d3-contour lib (https://github.com/d3/d3-contour)

mcnuttandrew:
@Apercu will publish a new version (contour + radar = minor version)?

Apercu:
Sure! ;D

PR #441, opened 2017-05-31
Contour Series

mcnuttandrew:
This PR adds a a contour-density series based on d3-contour. It acts a much improved version of the heatmap series. The example is a slightly different version of Bostock's Old faithful (https://bl.ocks.org/mbostock/e3f4376d54e02d5d43ae32a7cf0e6aa9). Open to API suggestions and comments! 

contour

In the near future it could also get a canvas mode pretty easily. Still to do

- [x] Tests
- [x] Docs

This Address #419


showcase/plot/contour-series-example.js
```diff
@@ -0,0 +1,81 @@
...
+              stroke: '#125C77',
+              strokeLinejoin: 'round'
+            }}
+            colorRange={[
+              '#79C7E3',
+              '#FF9833'
+            ]}
+            data={data}/>
+          <MarkSeries animation data={data} size={1} color={'#125C77'}/>
+          <Borders style={{
```
Apercu: Might probably be a good idea to add a global property `all` that all four side extend, to prevent redundancy in cases like this one

PR #447, opened 2017-06-08
Fix VerticalRectSeries - single time slice expands xDomain

### LINKED ISSUE #443: VerticalRectSeries - single time slice expands xDomain
Edge case with `VerticalRectSeries` - when there is a single [x0, x] pair, the x domain is expanded out far beyond the realm of the data (~1970 - ~2070, which I think is the default in d3-axis?) 

I assume this occurs with other series types too, although haven't tested for those.

The issue still occurs when you specify an `xDomain` on `XYPlot`.

## Example

```jsx
<XYPlot
  width={300}
  height={300}
  xType="time"
  stackBy="y">
  <VerticalGridLines />
  <HorizontalGridLines />
  <XAxis />
  <YAxis />
  <VerticalRectSeries
    data={[
      {x0: new Date(2015, 0, 1).valueOf(), x: new Date(2015, 1, 1).valueOf(), y: 16}
    ]}
  />
  <VerticalRectSeries
    data={[
      {x0: new Date(2015, 0, 1).valueOf(), x: new Date(2015, 1, 1).valueOf(), y: 12}
    ]}/>
</XYPlot>
```

## Output

screen shot 2017-05-31 at 9 47 35 am

sandiiarov:
Fixes #443

### DIFF
--- a/src/utils/scales-utils.js
+++ b/src/utils/scales-utils.js
@@ -355,20 +355,37 @@ function _computeScaleDistance(values, domain, bestDistIndex, scaleFn) {
 }
 
 /**
+ * Normilize array of values with a single value.
+ * @param {Array} arr Array of data.
+ * @param {Array} values Array of values.
+ * @param {string} attr Attribute.
+ * @param {string} type Type.
+ * @private
+ */
+function _normalizeValues(data, values, attr, type) {
+  if (type === TIME_SCALE_TYPE && values.length === 1) {
+    const attr0 = data[0][`${attr}0`];
+
+    return [attr0, ...values];
+  }
+
+  return values;
+}
+
+/**
  * Get the distance, the smallest and the largest value of the domain.
  * @param {Array} data Array of data for the single series.
  * @param {Object} scaleObject Scale object.
- * @returns {{domain0: number, domainN: number, distance: number}} Resuylt.
+ * @returns {{domain0: number, domainN: number, distance: number}} Result.
  * @private
  */
 export function _getScaleDistanceAndAdjustedDomain(data, scaleObject) {
-  const {attr, domain} = scaleObject;
+  const {attr, domain, type} = scaleObject;
 
-  const values = getUniquePropertyValues(data, attr);
+  const uniqueValues = getUniquePropertyValues(data, attr);
+
   // Fix time scale if a data has only one value.
-  if (scaleObject.type === TIME_SCALE_TYPE && values.length === 1) {
-    values.unshift(data[0][`${scaleObject.attr}0`]);
-  }
+  const values = _normalizeValues(data, uniqueValues, attr, type);
 
   const index = _getSmallestDistanceIndex(values, scaleObject);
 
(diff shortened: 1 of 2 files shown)


src/utils/scales-utils.js
```diff
@@ -363,7 +363,13 @@ function _computeScaleDistance(values, domain, bestDistIndex, scaleFn) {
  */
 export function _getScaleDistanceAndAdjustedDomain(data, scaleObject) {
   const {attr, domain} = scaleObject;
-  const values = getUniquePropertyValues(data, attr);
+
+  let values = getUniquePropertyValues(data, attr);
```
mcnuttandrew: Can you make this a helper function to avoid the let?

src/utils/scales-utils.js
```diff
@@ -363,7 +363,13 @@ function _computeScaleDistance(values, domain, bestDistIndex, scaleFn) {
...
+  // Fix time scale if a data has only one value.
+  if (scaleObject.type === TIME_SCALE_TYPE && values.length === 1) {
+    values = [data[0].x0, ...values];
```
mcnuttandrew: Should make sure the thing you are looking at has an x0, e.g. What if there is a y0, as is the case for horizontalRectSeries?

src/utils/scales-utils.js
```diff
@@ -367,7 +367,7 @@ export function _getScaleDistanceAndAdjustedDomain(data, scaleObject) {
   const values = getUniquePropertyValues(data, attr);
   // Fix time scale if a data has only one value.
   if (scaleObject.type === TIME_SCALE_TYPE && values.length === 1) {
-    values.unshift(data[0].y0 || data[0].x0);
+    values.unshift(data[0][`${scaleObject.attr}0`]);
```
mcnuttandrew: I still think this would be better as a helper method, it would aid in code ledgibility, would guarantee we have room to grow if there are more cases, and, most importantly, would avoid mutating data.

src/utils/scales-utils.js
```diff
@@ -363,7 +363,13 @@ function _computeScaleDistance(values, domain, bestDistIndex, scaleFn) {
  */
 export function _getScaleDistanceAndAdjustedDomain(data, scaleObject) {
   const {attr, domain} = scaleObject;
+
   const values = getUniquePropertyValues(data, attr);
+  // Fix time scale if a data has only one value.
+  if (scaleObject.type === TIME_SCALE_TYPE && values.length === 1) {
```
mcnuttandrew: I still think this should be a helper method. It'll make the code easier to read and can avoid mutating data

src/utils/scales-utils.js
```diff
@@ -363,7 +363,13 @@ function _computeScaleDistance(values, domain, bestDistIndex, scaleFn) {
...
+    values.unshift(data[0][`${scaleObject.attr}0`]);
+  }
+
```
mcnuttandrew: Aren't there cases when an att won't have a zero case?

src/utils/scales-utils.js
mcnuttandrew: Sorry about double comment, github mobile ain't great

src/utils/scales-utils.js
sandiiarov: I am not sure.

But if we try to pass a data without x0 or y0

screen shot 2017-06-09 at 3 16 01 pm

we get these errors:

screen shot 2017-06-09 at 3 14 39 pm

samhogg: @sandiiarov try using a `VerticalBarSeries` with just an `[x,y]` tuple - `VerticalRectSeries` requires an `x0`

sandiiarov: With `VerticalRectSeries` I get the same errors

src/utils/scales-utils.js
sandiiarov: Hmm. What do you think about this:

mcnuttandrew:
Sorry for the long delay, was on vacation. Looks good!

PR #449, opened 2017-06-09
Allow optional padding prop to scale axis

CharlieMartell:
This PR lets a user supply some optional scaling props in order to add padding to their axis. This solves a problem that charts with large data sets with small standard deviations causing them to be deceptive in their visualization.


examples/zoomable-chart/yarn.lock
```diff
@@ -1 +0,0 @@
-../yarn.lock
```
btford: Why delete this?

tests/utils/scales-utils-tests.js
```diff
@@ -304,3 +348,7 @@ test('scales-utils #_adjustCategoricalScale', t => {
   });
   t.end();
 });
+
+test('scales-utils #missingScaleProps_padDomain', t => {
```
Apercu: add a TODO or delete?

examples/zoomable-chart/yarn.lock
CharlieMartell: this was removed and readded automatically by running the script that removes references to unpm

mcnuttandrew:
I don't understand this pr, why can't you just set the domain via xDomaub or yDomain?

mcnuttandrew:
Couple of additional comments

-there isn't any example usage or docs?
- this API doesn't match the rest of the react-vis's API. I would suggest instead xPadding and yPaadding rather than an object of padding, as this can't be re-used for other variables (r or theta)

CharlieMartell:
You could pass in a domain via xDomain or yDomain, however the domain is calculated in scales-utils getDomainByAttr and if you are using react-vis and want to specify your own domain you have to copy the way react-vis calculates domain in order to modify it according to dynamic data. Allowing the user to scale their data for visualization seems in line with react-vis goal no?

With respect to it not matching the API that makes sense, I will go ahead and adjust to better fit with documentation.

mcnuttandrew:
K I think I get it now. Maybe this could be expanded to include left and right padding? E.g. XLeftPadding & XRightPadding for additional specificity

Also, the user should prolly get a warning if they are supplying a domain and a padding as that'll probably lead to confusion

PR #450, opened 2017-06-09
Make padding more in line with react-vis API

CharlieMartell:
This PR addresses some of the comments surrounding PR449.


docs/markdown/scales-and-data.md
```diff
@@ -22,6 +22,9 @@ All scale properties are prepended with the name of the attribute, for instance
 * `[name]Domain` (optional)  
   Type: `Array`  
   Array of values to visualize from. If domain is not passed, it will be calculated from the values which are passed to component.
+* `[name]Padding` (optional)  
+  Type: `Number`  
+  A percentage that will pad your `[name]Domain`, if you supply `[name]Domain` it cannot be padded aswell, if not supplied `[name]Domain` will not be padded.
```
mcnuttandrew: Typo

src/plot/xy-plot.js
```diff
@@ -221,14 +221,14 @@ class XYPlot extends React.Component {
   }
 
   /**
-   * Get the list of scale-related settings that are optionally passed into x-y-plot
+   * Get the list of optional scale-related settings
    * @param {Object} props Object of props.
    * @returns {Object} Optional Props.
    * @private
    */
   _getOptionalScaleProps(props) {
```
mcnuttandrew: Is this method necessary now?

CharlieMartell: Yes or else attributes following `[name]Padding` schema will not be passed to getMissingScaleProps. Not all props are passed in only ones that fall under "defaultScaleProps, optionalScaleProps, userScaleProps" It is perhaps forward looking though and could be reduced to just checking if a property ends with Padding. I feel this would be less general though than checking against OPTIONAL_SCALE_PROPS

CharlieMartell:
datapadded
Shows with padding and without on the same data set

PR #454, opened 2017-06-15
More sensible color options for series

jckr:
This PR allows users to use scales at the series level. 

For instance, let's suppose you want to color 20 line series using a characteristics of the series to color them, for instance with a linear or category scale. Well, today, you can't - you'd have to pass color values for a series as literals. But with this PR, you can pass a numeric value and use a linear scale, or pass anything and use a category scale.

Another thing this PR does is change the default color ( + fill, + stroke) scale range. Today, the range is optimized for linear scales, and it's only 2 colors. If the user specifies a category scale, the default palette doesn't make sense - it would be repeating 2 colors. By contrast, we have a default palette for legends. Now, if you specify a categorical palette for color (or fill or stroke) and you don't specify a range, you get our default _categorical_ palette.


mcnuttandrew:
Nice! (posting a screen shot for anyone who is curious about this PR)

screen shot 2017-06-15 at 5 11 37 pm

mcnuttandrew:
Looks good! Add a test for the new showcase example to test that each line is receiving appropriate colors?

jckr:
@mcnuttandrew done

mcnuttandrew:
Stamp!

PR #460, opened 2017-06-22
Color documentation

jckr:
Pretty long / hopefully comprehensive document about all the ways to use color in react-vis. 
When I started really embracing react-vis getting colors right was my main pain point. Let's hope that this will help others gain full control of color. 

This also highlights a few quirks / inconsistencies of react-vis when it comes to color, so there's more to do on that front. 

Chosen bits - 
screenshot 2017-06-21 18 42 19

--- 

screenshot 2017-06-21 18 42 13

### DIFF
--- a/docs/docs-app/constants/pages.js
+++ b/docs/docs-app/constants/pages.js
@@ -178,6 +178,13 @@ export const docPages = generatePath([
         pageType: 'documentation'
       }
     }, {
+      name: 'Colors',
+      content: {
+        markdown: getDocUrl('colors.md'),
+        filename: 'colors.md',
+        pageType: 'documentation'
+      }
+    },{
       name: 'Legends',
       content: {
         markdown: getDocUrl('legends.md'),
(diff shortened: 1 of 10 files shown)


samhogg:
This is great work, nice job @jckr 👍

jckr:
Thanks @mcnuttandrew and @samhogg . The code does work - the document embeds live examples, not screenshots. However, the live examples are slightly different from the code displayed on the page - the data is being randomly generated, not stored in variables called series1 or series2 for instance. There are also extra props passed to XYPlot for aesthetic reasons. 

my question to you @mcnuttandrew is if I merge this will the gh-pages branch / the .io page be automatically updated or how does that work?

mcnuttandrew:
> The code does work - the document embeds live examples, not screenshots.

I'm not doubting that the code currently works, I want to make sure that it always works! That said, this PR is so 🔥🔥🔥  that if writing tests for it blocks you, I can write you some (maybe make a task for it?). 

> my question to you @mcnuttandrew is if I merge this will the gh-pages branch / the .io page be automatically updated or how does that work?

no it will not, there is one more step in between. you can either follow this tutorial, or i can cut a new copy for you https://github.com/uber/react-vis/wiki/Deploying-Documentation

Finally, it looks like you need to run the remove-unpm script to beat the travis errors.

PR #486, opened 2017-06-30
Radar center label text by default

zachguo:
Before:
screen shot 2017-06-30 at 2 00 29 pm
After:
screen shot 2017-06-30 at 2 10 35 pm

### DIFF
--- a/showcase/radar-chart/animated-radar-chart.js
+++ b/showcase/radar-chart/animated-radar-chart.js
@@ -72,6 +72,9 @@ export default class AnimatedRadar extends Component {
               text: {
                 opacity: 0
               }
+            },
+            labels: {
+              textAnchor: 'middle'
             }
           }}
           margin={{
--- a/src/radar-chart/index.js
+++ b/src/radar-chart/index.js
@@ -245,7 +245,8 @@ RadarChart.defaultProps = {
       text: {}
     },
     labels: {
-      fontSize: 10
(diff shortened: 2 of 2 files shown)


showcase/radar-chart/animated-radar-chart.js
```diff
@@ -72,6 +72,9 @@ export default class AnimatedRadar extends Component {
               text: {
                 opacity: 0
               }
+            },
+            labels: {
```
mcnuttandrew: not necessary if this is the new default?

zachguo: It's just for showing the effect, or the default `labels` style will be override by an empty `{}`.

mcnuttandrew: that shouldn't be the desired behavior? the desired effect should that the labels style should read: {...DEFAULT_STYLE, ....USER_PROVIDED_STYLE};

regardless, this PR looks fine ill merge it when the tests pass

zachguo: Ideally yes, but when user passes a custom `style` prop, [`style.labels` here](https://github.com/zachguo/react-vis/blob/03cfbe36b7e7c9fa018ad9b2752c712ba6e50bcd/src/radar-chart/index.js#L190) would be undefined.

PR #490, opened 2017-07-03
Add tests for axes, develop multi-line example for axis

mcnuttandrew:


zachguo:
Can we have a new release after this PR is merged? Thanks!!

mcnuttandrew:
> Can we have a new release after this PR is merged? Thanks!!  

i think most folks are ooo for the long weekend, so itll most likely happen Wednesday (unless @apercu happens to be in today for some reason)

Issue #455, opened 2017-06-15
Add makeHeightFlexible


mcnuttandrew:
In #262 @Firenze11 suggested

"Is there a plan to also expose "makeHeightFlexible" too? I'm try to use a HorizontalBarSeries chart to function as a vertical indexing sidebar, and found it would be useful to auto-determine its height."

jckr:
<img width="1133" alt="screenshot 2017-06-29 19 45 45" src="https://user-images.githubusercontent.com/9727630/27718866-95a5128c-5d03-11e7-846c-2bb50becf003.png">
will write the PR tomorrow

PR #491, opened 2017-07-05
make height flexible and vis flexible added

jckr:
addresses https://github.com/uber/react-vis/issues/455
screenshot 2017-07-04 17 54 53

### DIFF
--- a/docs/markdown/flexible-plots.md
+++ b/docs/markdown/flexible-plots.md
@@ -12,7 +12,7 @@ const FlexibleVis = makeVisFlexible(XYPlot);
 
 makeWidthFlexible modifies XYPlot so that it no longer requires a width. It will take all the with in its container div. Likewise, makeHeightFlexible modifies XYPlot so that is no longer requires a height. Its height will be that of its container div. Finally, makeVisFlexible modifies XYPlot so that it no longer requires eiter a width and a height. Its dimensions will be that of its container.
 
-The result of that function can be used exactly as XYPloy:
+The result of that function can be used exactly as XYPlot:
 
 ```jsx
 <FlexibleWidth height={100}>
@@ -33,3 +33,6 @@ A flexible plot is useful when you don't know for sure the size of the container
 
 ### Size of parent container is not the same as "all the available space" 
 Flexible plots will inherit dimensions from their container. This is not the same thing as taking all the available space; if there are other elements in that container, a flexible plot won't deduce their dimensions from its own. For best results, a flexible plot should be the only child of its container.
+
+### Responsive visualizations
+We can go one step beyond and not simply adjust the dimensions of the chart of the available space, but change how a dataset is being represented. See the example [responsive visualizations](/#/examples/charts/responsive-vis)
--- a/docs/markdown/flexible-plots.md
+++ b/docs/markdown/flexible-plots.md
@@ -1,0 +1,35 @@
+## Flexible plots
+
+By default, XYPlot requires a width and a height. There are times, however, when you'd like your chart to take all the space it can. For these cases, React-vis provides a helper function to make one or both dimensions of your chart flexible.
+
+<!-- INJECT:"FlexibleCharts" -->
+
+```jsx
+const FlexibleWidth = makeWidthFlexible(XYPlot);
+const FlexibleHeight = makeHeightFlexible(XYPlot);
+const FlexibleVis = makeVisFlexible(XYPlot);
+```
+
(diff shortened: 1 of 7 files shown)


docs/markdown/flexible-plots.md
```diff
@@ -0,0 +1,35 @@
+## Flexible plots
+
+By default, XYPlot requires a width and a height. There are times, however, when you'd like your chart to take all the space it can. For these cases, React-vis provides a helper function to make one or both dimensions of your chart flexible.
```
mcnuttandrew: docs note: we should probably decide on a consistant spelling of react-vis, eg is React-Vis/React Vis/react-vis/etc

docs/markdown/flexible-plots.md
```diff
@@ -0,0 +1,35 @@
...
+
+```jsx
+const FlexibleWidth = makeWidthFlexible(XYPlot);
+const FlexibleHeight = makeHeightFlexible(XYPlot);
+const FlexibleVis = makeVisFlexible(XYPlot);
+```
+
+makeWidthFlexible modifies XYPlot so that it no longer requires a width. It will take all the with in its container div. Likewise, makeHeightFlexible modifies XYPlot so that is no longer requires a height. Its height will be that of its container div. Finally, makeVisFlexible modifies XYPlot so that it no longer requires eiter a width and a height. Its dimensions will be that of its container.
+
+The result of that function can be used exactly as XYPloy:
```
mcnuttandrew: typo

docs/markdown/flexible-plots.md
```diff
@@ -0,0 +1,35 @@
...
+
+<FlexibleVis>
+  <VerticalBarSeries data={data} />
+</FlexibleVis>
+```
+
+### No worries about resizing
+A flexible plot is useful when you don't know for sure the size of the container where the chart will go. On top of that, flexible plots resize themselves when the window size changes. You can try that by changing the size of this window.
+
+### Size of parent container is not the same as "all the available space"
```
mcnuttandrew: maybe add a note a about responsive vis? http://uber.github.io/react-vis/#/examples/charts/responsive-vis

tests/components/xy-plot-tests.js
```diff
@@ -148,3 +150,18 @@ test('Render a bar chart with some nonAnimatedProps', t => {
 
   t.end();
 });
+
+test('testing flexible charts', t => {
```
mcnuttandrew: +1 for tests

src/make-vis-flexible.js
```diff
@@ -112,12 +114,18 @@ export default function makeVisFlexible(Component) {
...
       const containerElement = getDOMNode(this.refs[CONTAINER_REF]);
-      const offsetWidth = containerElement.offsetWidth;
-      if (this.state.width !== offsetWidth) {
-        this.setState({
-          width: offsetWidth
-        });
-      }
+      const {offsetHeight, offsetWidth} = containerElement;
+
+      const newHeight = this.state.height === offsetHeight ? {} :
```
mcnuttandrew: maybe do a 

const {height, width} = this.state

 for better style?

showcase/flexible/flexible-examples.js
```diff
@@ -0,0 +1,79 @@
...
+  margin: {top: 10, left: 10, right: 10, bottom: 10}
+};
+
+const FlexibleWidth = makeWidthFlexible(XYPlot);
+const FlexibleHeight = makeHeightFlexible(XYPlot);
+const FlexibleVis = makeVisFlexible(XYPlot);
+
+export function FlexibleCharts({height, width}) {
+  return (<div>
+    <div style={{display: 'flex',
```
Apercu: do you mind putting each style property on a newline? having one on a line and three on the second is a bit inconsistent. same goes for the other div.

showcase/flexible/flexible-examples.js
```diff
@@ -0,0 +1,79 @@
...
+      justifyContent: 'space-between', position: 'relative', width: width || '60vw'}}>
```
Apercu: can we also use default props at the component level for the `width` and `height`? that would allow us to remove these conditionals

tests/components/xy-plot-tests.js
jckr: yeah they're not great though bc the flexible dimension(s), at test time, is going to be 0px. it's animated.

src/make-vis-flexible.js
jckr: i thought we were generally against state/props deconstructing?

sandiiarov:
Hi. This is an awesome feature. Thank you! But I got an error

```
ERROR in ./~/react-vis/es/index.js
Module not found: Error: Can't resolve 'make-vis-flexible' in '/Users/alexs/Projects/111/node_modules/react-vis/es'
```

I am pretty sure that a problem is here  https://github.com/uber/react-vis/blob/master/src/index.js#L69

need to add './' like this one

`
export {makeHeightFlexible, makeVisFlexible, makeWidthFlexible} from './make-vis-flexible';
`

Issue #492, opened 2017-07-05
xDomain not respected for VerticalRectSeries


sqren:
Hi there,

I'm having trouble setting the range for a histogram made with `VerticalRectSeries`. 

```js
  <XYPlot xDomain={[0, 4]} width={500} height={100}>
    <XAxis marginRight={10} />
    <VerticalRectSeries
      data={[{ x0: 0, x: 1, y: 10 }, { x0: 2, x: 3, y: 20 }]}
    />
  </XYPlot>
```

I would expect the above to have x-axis between 0 and 4, but instead the axis goes from -1 to 5.
Should I specify xDomain as `[1,3]` or am I doing something wrong?

Example: https://codepen.io/sqren/pen/weXQmx?editors=0010

jckr:
Great question. Fix incoming 
![image](https://user-images.githubusercontent.com/9727630/27875115-f446736e-6166-11e7-8d0d-97b5d8df1f4f.png)

jckr:
Fixed in 1.6.6

sqren:
Awesome, thanks for the quick resolution!

PR #493, opened 2017-07-05
fixes issue #492

jckr:
For rect series one of the scales can get adjusted, just like for bar series, even though the user has control on both sides of the rectangle.
For instance (as highlighted in issue 492) for vertical bar series, x scale gets adjusted, even though user can control the position of the left and right edge of each mark. 
(likewise for horizontal rect series, y scale could get adjusted.)


mcnuttandrew:
Wait i take it back, it looks like i already wrote tests to that effect. make the tests pass and itll be fine

Issue #417, opened 2017-05-12
Canvas Components Should Animate


mcnuttandrew:
Right now the the canvas series do not get animated! They should be animatable like all the rest of components.

mcnuttandrew:
resolved

PR #496, opened 2017-07-05
Add a naive canvas animation system

mcnuttandrew:
This PR adds an animated canvas drawing system for the canvas components. Similarly to the regular animation system it runs on d3-interpolate, however we have to do the animation work ourself, because we are directly manipulating the canvas. This addresses #417 

canvas-animation

canvas-animation2

It runs a little clumsily when there are multiple animations running at once. Would be delighted to hear what someone who knows this stuff better has to say. That said, when it runs well, it runs significantly more smoothly than the same svg animations for medium volume data

(This PR Also update canvas so that there is a margin on the canvas itself.)

### DIFF
--- a/src/plot/series/canvas-wrapper.js
+++ b/src/plot/series/canvas-wrapper.js
@@ -151,7 +151,6 @@ class CanvasWrapper extends Component {
 
   render() {
     const {
-      children,
       marginBottom,
       marginLeft,
       marginRight,
@@ -167,7 +166,6 @@ class CanvasWrapper extends Component {
           height={innerHeight + marginTop + marginBottom}
           width={innerWidth + marginLeft + marginRight}
           ref="canvas" />
-        {children}
       </div>
     );
   }
--- a/src/plot/series/canvas-wrapper.js
+++ b/src/plot/series/canvas-wrapper.js
@@ -21,48 +21,153 @@
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 
+import {interpolate} from 'd3-interpolate';
+import {extractAnimatedPropValues} from 'animation';
+import {ANIMATED_SERIES_PROPS} from 'utils/series-utils';
+
+const MAX_DRAWS = 30;
+
+/**
+ * Draw loop draws each of the layers until it should draw more
+ * @param {CanvasContext} ctx - the context where the drawing will take place
+ * @param {Number} height - height of the canvas
+ * @param {Number} width - width of the canvas
+ * @param {Array} layers - the layer objects to render
+ */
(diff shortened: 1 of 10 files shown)


mcnuttandrew:
review bump?

src/plot/series/canvas-wrapper.js
```diff
@@ -21,47 +21,150 @@
...
+    };
+  });
+}
 class CanvasWrapper extends Component {
   componentDidMount() {
-    this.drawChildren(this.props, this.refs.canvas.getContext('2d'));
+    this.drawChildren(this.props, null, this.refs.canvas.getContext('2d'));
   }
 
   componentWillUpdate(nextProps) {
```
chrisirhc: Since you're doing render operations in here, `componentDidUpdate` might be more appropriate according to the lifecycle docs:
https://facebook.github.io/react/docs/react-component.html#componentdidupdate

src/plot/series/canvas-wrapper.js
```diff
@@ -21,47 +21,150 @@
...
+    const height = innerHeight + marginTop + marginBottom;
+    const width = innerWidth + marginLeft + marginRight;
+    const layers = buildLayers(newProps.children, oldProps ? oldProps.children : []);
+    // if we don't need to be animating, dont! cut short
+    if (!childrenShouldAnimate) {
+      drawLayers(ctx, height, width, layers);
+      return;
+    }
+
+    engageDrawLoop(ctx, height, width, layers);
```
chrisirhc: Looks like we're missing cleanup of the drawCycle intervals. All previous drawCycle interval timers need to be cleared every time a new drawCycle is created.

They also need to be cleaned up on unmount.

src/plot/series/canvas-wrapper.js
```diff
@@ -21,47 +21,150 @@
...
       return;
     }
-    ctx.clearRect(0, 0, innerWidth, innerHeight);
-    children.forEach(layer => layer.type.renderLayer(layer.props, ctx));
+
+    const childrenShouldAnimate = checkChildrenForAnimation(children);
+
+    const height = innerHeight + marginTop + marginBottom;
+    const width = innerWidth + marginLeft + marginRight;
+    const layers = buildLayers(newProps.children, oldProps ? oldProps.children : []);
```
akre54: is the `oldProps` just for `children`? In general I think it's better to not try and add extra stateful logic. Let's keep this in the lifecycle methods like @chrisirhc said

mcnuttandrew: I'm not sure i get what you are suggesting? oldProps is necessary for interpolation against the previous state?

Issue #498, opened 2017-07-06
broken link in documentation


alaindresse:
The link to mark series in
    http://uber.github.io/react-vis/#/documentation/xy-plot/series
is broken.

It should be
    http://uber.github.io/react-vis/#/documentation/xy-plot-series/mark-series
instead of
    http://uber.github.io/react-vis/mark-series.md.

mcnuttandrew:
Should be resolved, marking as closed

alaindresse:
Super impressed: Super quick !

PR #499, opened 2017-07-06
Fix broken mark series link, touch up series explanation

mcnuttandrew:
Addresses #498


jckr:
do you mind changing it in interaction.md and scales-and-data.md which had the old (but then correct) markseries.md spelling?
thanks

mcnuttandrew:
Whoops, i missed that in my grep, fix coming up

PR #517, opened 2017-07-22
Custom SVG series

mcnuttandrew:
This PR adds a custom svg series. This series adds a ```<g>``` element based on an xy coordinate, and the fills it with a custom component. 

screen shot 2017-07-22 at 2 23 53 pm

Usage is like this:

```jsx
<CustomSVGSeries
          className="custom-marking"
          customComponent="square"
          data={[
            {x: 1, y: 10, customComponent: 'circle', size: 10},
            {x: 1.7, y: 12, size: 20, style: {stroke: 'red', fill: 'orange'}},
            {x: 2, y: 5},
            {x: 3, y: 15},
            {x: 2.5, y: 7, customComponent: (customComponent, positionInPixels) => {
              return (
                <g>
                  <circle cx="0" cy="0" r={10} fill="green"/>
                  <text x={0} y={0}>
                    <tspan x="0" y="0">{`x: ${positionInPixels.x}`}</tspan>
                    <tspan x="0" y="1em">{`y: ${positionInPixels.y}`}</tspan>
                  </text>
                </g>
              );
            }}
          ]}/>
```

Where the contents of the custom component can be declared on the series level or on the row level. They can be declared either through a string call to a predefined component (circle/square/etc) or from a function. In #423, @jckr was suggesting that user just be able to define a path element, however I feel this is already covered pretty well with the polygon series. This solution allows for a pretty total flexibility, and doesn't make the API of markSeries hard to maintain across render types. When complete this will address the suggestion raised in #342, as well as my own comments in #423.

Still to go 
- [x] Docs
- [x] Tests
- [x] Two additional string generated components (cross and diamond)
- [x] JSDocs on the functions in the new series

Questions
- Are there any additional shapes I should have available by default?
- Are the names I'm using for the custom inner component clear?

### DIFF
--- a/docs/markdown/custom-svg-series.md
+++ b/docs/markdown/custom-svg-series.md
@@ -1,0 +1,93 @@
+## CustomSVGSeries
+
+When creating visualizations, it is sometimes necessary to get your hands dirty and completely take control over what SVG components will be shown. This could be necessary in situations where you have predefined SVG code that you just want to appear the way you drew it in Sketch (but positioned using coordinates), or maybe you have multiline text annotations that you want to formatted in a particular way, or maybe even you just want to use an alternative type of mark instead of the usual scatterplot mark to differentiate series in a set. To serve these and many other tasks, we use the CustomSVGSeries.
+
+<!-- INJECT:"CustomSVGExample" -->
+
+The premise of the series is that it simply puts a ```<g>``` element at a desired x,y location, and the offers you a variety of ways to fill in the contents of that ```<g>``` element. Here's an example of the data format:
+
+```javascript
+const myData = [
+  {x: 1, y: 10, customComponent: 'circle', size: 10},
+  {x: 1.7, y: 12, size: 20, style: {stroke: 'red', fill: 'orange'}},
+  {x: 2, y: 5},
+  {x: 3, y: 15},
+  {x: 2.5, y: 7, customComponent: (customComponent, positionInPixels) => {
+    return (
+      <g>
+        <circle cx="0" cy="0" r={10} fill="green"/>
+        <text x={0} y={0}>
+          <tspan x="0" y="0">{`x: ${positionInPixels.x}`}</tspan>
+          <tspan x="0" y="1em">{`y: ${positionInPixels.y}`}</tspan>
+        </text>
+      </g>
+    );
+  }}
+]
+```
+
+Just like other series, x and y are used to position the group. The customComponent key word is used to determine how to fill in the svg (see below), and then size is used modify the size of the contents when using a string. Used in context of the series:
+
+```javascript
+<XYPlot width={300} height={300}>
+  <CustomSVGSeries customComponent="square" data={myData} />
+</XYPlot>
+```
+
+### Defining your marks
+
+The type of custom svg marks can be determined in one of several ways:
+
+- As a string on a series level
+- As a function on a series level
+- As a string on row level
+- As a function on a row level
+
+There are currently four types of string accessible custom marks: **star**, **square**, **circle**, and **diamond**. If using a string, it can be useful to specify a size for the mark. This is done on a row level (see above data api example), with the size prop. Size is expressed in pixels, and is NOT scaled with the normal react-vis size keyword. They look like this:
+
+<!-- INJECT:"CustomSVGAllTheMarks" -->
+
+If using a function to defined your mark, it is important to note that the function receives two arguments (customComponent, positionInPixels), where customComponent is the row of data as you have defined it. Thus if you are defining a function for the series as a whole you can make modifications based on the individual row as you go!
+
+
+
+## API reference
+
+#### animation (optional)  
+See the [XYPlot](xy-plot.md)'s `animation` section for more information.
+
+#### className (optional)
+Type: `string`
+Provide an additional class name for the series.
+
+#### customComponent (optional)
+Type: `string|function`
+Provides the mark type for the entire series. Defaults to a 'circle'. See `Defining You Marks` above.
+
+#### data
+Type: `Array<Object>`
+Array of data for the series. See above data format reference.
+
+### style
+Type: `object`
+A list of CSS properties to style the series outside of the explicitly set properties. Note that it will override all other properties (ie fill, stroke, opacity, color). See [style](style.md)
+
+## Interaction handlers
+#### onNearestX (optional)
+Type: `function(value, {event, innerX, index})`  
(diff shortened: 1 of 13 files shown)


atlas5d-matt:
As an outsider perspective I'm super excited about this. I need to make graphs with a circle mark and variance "whisker" lines, and had started trying to adapt https://github.com/uber/react-vis/pull/423 as a way to handle this. This approach seems even better though.

One piece of initial feedback I'd give as I'm just starting to play with this: it appears that to use a function for a customComponent it has to be defined on the data point. Maybe I'm mistaken and I was just doing something wrong, but I'd find it very useful to be able to use that approach on the series itself as well. I'll try to contribute towards that once I get my usage demo fleshed out.

mcnuttandrew:
hey @atlas5d-matt,

> I need to make graphs with a circle mark and variance "whisker" lines, and had started trying to adapt #423 as a way to handle this.  

That's an interesting use case! One word of caution: I'd worry about recasting your variance lines in a new coordinate system (which is what inserting a g would do). In order to make use of react-vis's built in I would maybe suggest doing a custom series, ala the candle stick example [source](https://github.com/uber/react-vis/tree/master/examples/candlestick)/[demo](http://uber.github.io/react-vis/#/examples/charts/candlestick). That way your whisker lines will be related to the same spatial variable as the marks! 

(btw, if you wanted to be super helpful and cool, you could productionalize the series in that example, so that everyone could do that)

atlas5d-matt:
@mcnuttandrew Thanks! I hadn't spotted it yet, but the Candlestick example is exactly what I need. I had already come to the realization that for the CustomSVGSeries approach I was going to need to have the yFunctor passed in to the customComponent function (maybe still useful for some other use?), but this custom series approach does seem better for my current purpose.

By "productionalize" do you mean making example code that calls the CandlestickSeries, or adding the CandlestickSeries to the react-vis codebase as a series to choose from, or something else?

mcnuttandrew:
>By "productionalize" do you mean making example code that calls the CandlestickSeries, or adding the CandlestickSeries to the react-vis codebase as a series to choose from, or something else?

I mean move the CandlestickSeries out of the examples and into the main library. A complete PR  would mean writing tests (and therein adding examples), adding documentation, and making the series more robust (from what I recall it wasn't perfect). I added a task about it (#521) which has some notes on what i think would be a strong implementation.

mcnuttandrew:
screen shot 2017-07-28 at 1 21 34 pm

src/plot/series/custom-svg-series.js
```diff
@@ -0,0 +1,159 @@
...
+
+    const x = this._getAttributeFunctor('x');
+    const y = this._getAttributeFunctor('y');
+
+    return (
+      <g className={`${predefinedClassName} ${className}`}
+         ref="container"
+         transform={`translate(${marginLeft},${marginTop})`}
+         style={style}>
+        {data.map((seriesComponent, index) => {
```
Apercu: maybe separate the function for lisibility

src/plot/series/label-series.js
```diff
@@ -80,7 +80,7 @@ class LabelSeries extends AbstractSeries {
             textAnchor: leftOfMiddle ? 'start' : 'end',
             x,
             y,
-            transform: `rotate(${rotation},${x},${y})`,
+            transform: `rotate(${rotation || 0},${x},${y})`,
```
Apercu: use defaultProps?

src/plot/series/custom-svg-series.js
```diff
@@ -0,0 +1,159 @@
...
+    } = this.props;
+
+    if (!data || !innerWidth || !innerHeight) {
+      return null;
+    }
+
+    if (animation) {
+      return (
+        <Animation {...this.props} animatedProps={ANIMATED_SERIES_PROPS}>
+          <CustomSVGSeries {...this.props} animation={null}/>
```
mcnuttandrew: animation is more than just a boolean! it can also be a string or an object. I'm 99% sure that if i put animation={false} as jsx animation will still be read as true (while if i were to spread an object of props that contain animation: false, that would be read correctly)

docs/markdown/custom-svg-series.md
```diff
@@ -0,0 +1,93 @@
...
+const myData = [
+  {x: 1, y: 10, customComponent: 'circle', size: 10},
+  {x: 1.7, y: 12, size: 20, style: {stroke: 'red', fill: 'orange'}},
+  {x: 2, y: 5},
+  {x: 3, y: 15},
+  {x: 2.5, y: 7, customComponent: (customComponent, positionInPixels) => {
+    return (
+      <g>
+        <circle cx="0" cy="0" r={10} fill="green"/>
+        <text x={0} y={0}>
```
mcnuttandrew: I'm not sure I get what you mean? can you say more?

Apercu: in this case, what would `console.log(customeComponent)` output?

src/plot/series/custom-svg-series.js
Apercu: then the proptype probably needs to be updated ;) and maybe not provide it here?

docs/markdown/custom-svg-series.md
mcnuttandrew: gotcha! i see what you mean now, fixing

src/plot/series/custom-svg-series.js
mcnuttandrew: wait i am totally wrong! animation={false} is valid, i guess i may have learned that lesson incorrectly from doing animation="false" or somethign

src/plot/series/custom-svg-series.js
```diff
@@ -99,7 +99,7 @@ class CustomSVGSeries extends AbstractSeries {
     if (animation) {
       return (
         <Animation {...this.props} animatedProps={ANIMATED_SERIES_PROPS}>
-          <CustomSVGSeries {...this.props} animation={null}/>
+          <CustomSVGSeries {...this.props} animation={false}/>
```
Apercu: Do we pass this just to show this prop exists?

mcnuttandrew: If animation wasn't false then we would get an infinite stack of animations? I'm not sure if i catch your meaning?

Apercu: not providing the prop will do as if it was `undefined` and the condition would not be met anyway right?

mcnuttandrew: I guess i don't understand how we would remove it from the parent to the child with out doing it this way?

Apercu: oh...................... oups... hehe nothing to see here 😹 🃏

Issue #520, opened 2017-07-27
Doc Typos


headfox23:
I found some typos in the docs:
https://github.com/uber/react-vis/blob/master/docs/markdown/line-mark-series.md#onvalueclick
The title says `onvalueclick` but the example shows `onClick`
The same applys to:
- onValueMouseOut -> onMouseOut
- onValueMouseOver -> onMouseOver

Thanks :)

PS: For lazy copy-paste-victims like me its misleading like hell because it simply doesnt work :D

mcnuttandrew:
resolved!

PR #522, opened 2017-07-28
correct docs typos

mcnuttandrew:
Addresses #520


mcnuttandrew:
Man, I wish we had a way for automatically maintaining the code that we show in these contexts. Eg it would be rad if we could someone write tests against these code snippets, but that's a story for another day.

Issue #413, opened 2017-05-08
Docs page using development version of react


mcnuttandrew:
![screen shot 2017-05-08 at 12 59 36 pm](https://cloud.githubusercontent.com/assets/6854312/25822705/4706a2e6-33ee-11e7-8859-d54142eed3a5.png)

https://facebook.github.io/react/docs/optimizing-performance.html#use-the-production-build

mcnuttandrew:
The docs also need to be upgraded to react-router v4 in order to clear some errors

PR #547, opened 2017-08-12
Use prod version of react and minify bundle

mcnuttandrew:
Addresses #413


docs/package.json
```diff
@@ -18,6 +18,7 @@
     "autoprefixer-loader": "^3.2.0",
     "babel-core": "^6.21.0",
     "babel-loader": "^6.2.10",
+    "babel-preset-es2015": "^6.24.1",
     "babel-preset-es2017": "^6.22.0",
```
Apercu: can probably remove this one if we're not using it anymore

Issue #541, opened 2017-08-11
Treemap support for 'literal' colorType


jchenjc:
Hi, 

I was able to customize the Sunburst chart by passing in a tree with style built into the data:
`
{
    title: 'John',
    size: 1,
    style: {
      fill: '#FF991F', 
      fillOpacity: 0.533
    }
}
`.

However, Treemap didn't respect the same tree data, the chart showed up with a default color (whatever I specified in the 'style' section has no effect). I don't see the 'colorType' property in the Treemap source. Is it supposed to be supported? If not, does that mean I need create a separate function to build a different tree for Treemap?

Thanks!

mcnuttandrew:
Hey @jchenjc,

ColorType is definately supported in Treemap! Each of the nodes checks to see if it has a color value, which is then interpreted via colorType. To wit, copy and pasted from the documentation: 

```javascripr
const myData = {
 "title": "analytics",
 "color": "#12939A",
 "children": [
  {
   "title": "cluster",
   "children": [
    {"title": "AgglomerativeCluster", "color": "#12939A", "size": 3938},
    {"title": "CommunityStructure", "color": "#12939A", "size": 3812},
    {"title": "HierarchicalCluster", "color": "#12939A", "size": 6714},
    {"title": "MergeEdge", "color": "#12939A", "size": 743}
   ]
  },
  {
   "title": "graph",
   "children": [
    {"title": "BetweennessCentrality", "color": "#12939A", "size": 3534},
    {"title": "LinkDistance", "color": "#12939A", "size": 5731},
    {"title": "MaxFlowMinCut", "color": "#12939A", "size": 7840},
    {"title": "ShortestPaths", "color": "#12939A", "size": 5914},
    {"title": "SpanningTree", "color": "#12939A", "size": 3416}
   ]
  },
  {
   "title": "optimization",
   "children": [
    {"title": "AspectRatioBanker", "color": "#12939A", "size": 7074}
   ]
  }
 ]
}
```

That said, you do highlight a valuable insight, we do not support a prop style on each of the nodes, and don't mention colorType directly in the docs. These things can certainly be fixed

mcnuttandrew:
Oh also, fill & fillOpacity are not directly relavant here bc the treemap uses DOM elements and not svg

jchenjc:
Thanks for the quick response again @mcnuttandrew!

[...]

Issue #431, opened 2017-05-24
Refactor Treemap


mcnuttandrew:
Treemap should make use of XYPlot! This will allow us to offer "rendering modes", namely divs/canvas/svg. Pretty cool!

PR #551, opened 2017-08-15
Enable SVG mode on treemap

mcnuttandrew:
An occasionally frustrating thing about the treemap is that it is the only chart type in react-vis that doesn't support SVG. This PR adds a SVG mode to tree map (#431) which could also be trivially extended to be canvas. 

Additionally this PR enables some inline style options which hadn't made their way to Treemap yet (#541)

### DIFF
--- a/src/treemap/index.js
+++ b/src/treemap/index.js
@@ -163,11 +163,8 @@ class Treemap extends React.Component {
     const {renderMode} = this.props;
     const {scales} = this.state;
     const nodes = this._getNodesToRender();
-
-    if (renderMode === 'SVG') {
-      return <TreemapSVG {...this.props} nodes={nodes} scales={scales}/>;
-    }
-    return <TreemapDOM {...this.props} nodes={nodes} scales={scales}/>;
+    const TreemapElement = renderMode === 'SVG' ? TreemapSVG : TreemapDOM;
+    return <TreemapElement {...this.props} nodes={nodes} scales={scales}/>;
   }
 
 }
--- a/src/treemap/index.js
+++ b/src/treemap/index.js
@@ -36,8 +36,10 @@ import {
 import {CONTINUOUS_COLOR_RANGE, DEFAULT_COLOR, DEFAULT_OPACITY, OPACITY_TYPE} from 'theme';
 import {AnimationPropType} from 'animation';
 import {getAttributeFunctor, getMissingScaleProps} from 'utils/scales-utils';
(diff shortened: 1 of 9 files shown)


src/treemap/index.js
```diff
@@ -153,66 +136,79 @@ class Treemap extends React.Component {
...
-            y1: useCirclePacking ? node.y : node.y1,
-            r: useCirclePacking ? node.r : 1,
-            scales: this.state.scales
-          };
-          return (<TreemapLeaf {...nodeProps} key={`leaf-${index}`} />);
-        })}
-      </div>
-    );
+
+    if (renderMode === 'SVG') {
```
Apercu: maybe do something like

```js
const TreemapElement = renderMode === 'SVG' ? TreemapSVG : TreemapDOM;
return (
  <TreemapElement {...this.props} nodes={nodes} scales={scales} />
);
```

to prevent duplication?

src/treemap/treemap-dom.js
```diff
@@ -0,0 +1,62 @@
...
+
+class TreemapDOM extends React.Component {
+  render() {
+    const {animation, className, height, mode, nodes, width, scales, style} = this.props;
+    const useCirclePacking = mode === 'circlePack';
+    return (
+      <div
+        className={`rv-treemap ${useCirclePacking ? 'rv-treemap-circle-packed' : ''} ${className}`}
+        style={{
+          width: `${width}px`,
```
Apercu: px by default, no need for the template string

```js
style={{ width, height }}
```

src/treemap/treemap-svg.js
```diff
@@ -0,0 +1,188 @@
...
+        {x, y: y + nodeHeight},
+        {x: x + nodeWidth, y: y + nodeHeight},
+        {x: x + nodeWidth, y}
+      ];
+
+      acc.updatedNodes = acc.updatedNodes.concat([
+        (<PolygonSeries
+          animation={animation}
+          className="rv-treemap__leaf"
+          key={index}
```
Apercu: More of a generic thing than something we need to fix before merging this diff, but we might need to look into another thing to provide as our keys, since that's not the preferred way to do it. React will be slow if we remove a node and the index changes but not the data, or if there is a reorder.

mcnuttandrew: i added an issue #553

Apercu:
👍

Issue #510, opened 2017-07-12
Sunburst - onValueClick not triggered when displaying Hint on onValueMouseOver


subhamagr:
The `onValueClick` is triggered when `Hint` is not displayed but not when `Hint` is displayed. I checked the css as well, everything seems fine.

I have worked on the showcase examples project.  Here's my code.

```javascript
import React from 'react';
import { Hint, Sunburst } from 'index';

const DIVERGING_COLOR_SCALE = ['#00939C', '#85C4C8', '#EC9370', '#C22E00'];

export default class AnimatedSunburst extends React.Component {
  state = {
    data: {},
    hover: null
  };

  render() {
    return (
      <div className="animated-sunburst-example-wrapper">
        <Sunburst
          onValueClick={e => console.log(e)}
          onValueMouseOver={v => this.setState({hover: v})}
          onSeriesMouseOut={v => this.setState({hover: null})}
          animation={{damping: 20, stiffness: 300}}
          data={this.state.data}
          colorType={'category'}
          colorRange={DIVERGING_COLOR_SCALE}
          style={{stroke: '#fff'}}
          height={300}
          width={350}
        >
        {this.state.hover &&
          <Hint value={this.state.hover} />
        }
        </Sunburst>
      </div>
    );
  }
```

Please help me knowing where I'm doing wrong.

Harjot1Singh:
Try removing the `animation` prop, I had an issue where the listeners don't really fire properly when it's present

mcnuttandrew:
Hey @subhamagr this issue should be fixed now! Sorry about the long delay

Issue #381, opened 2017-04-13
Listeners get clobbered


mcnuttandrew:
There is a reasonably frequent bug in which onValueMouseOut/onSeriesMouseOut listeners are not called due to the presence of other listeners. This often occurs in conjunction with the presence of animation. (To wit, check out the animated sunburst for more)

samhogg:
This is what I was running into in my other PR - in my experience `onMouseLeave` is more reliable

ajainarayanan:
@mcnuttandrew @samhogg  Are there any updates on this issue? Is there a resolution that was reached that you can share. Let me know if I can be of any help. I'm facing this issue on all the MarkSeries graphs that I'm rendering and as a workaround I right now bind the `esc` key and have an additional `x` on the `Hint` to close the hint. 

The `onValueMouseOver` gets triggered but `onValueMouseOut` is not getting triggered for me.

ajainarayanan:
For reference I have created this webpackbin if someone wants to try to reproduce the issue: https://www.webpackbin.com/bins/-Kok65xQnKmQZ5sp6xvL. The onmouseover triggers all the time but not the onmouseout event handler. Not sure if anything can be done on the react-vis layer. Probably an issue with chrome may be (https://github.com/facebook/react/issues/4492)? 🤔

vittvolt:
I'm also getting this problem on BarSeries, where onValueMouseOut / onSeriesMouseOut is not triggered.

mcnuttandrew:
Anyone following this issue might be interested in the discussion on #518

Vednus:
I got this working by turning animation off when onValueMouseOver is called and then turning it back on when onValueMouseOut is called

jckr:
I'm going to close this one and keep discussion on #518

PR #554, opened 2017-08-16
Make arc listeners respect mouse events during animation

mcnuttandrew:
This PR Addresses #510 and provides a technique for addressing #381. 

It is often the case that the animation component clobbers listeners. This can be super confusing! One solution is to slightly modify the way that we call animation to lay a shadow "listener layer" on top of the render SVG. This allows us to separate the concerns between calling events and making good animation.


mcnuttandrew:
@Apercu will you cut a patch version when you get a chance?

Issue #555, opened 2017-08-16
The link to source code for the Candlestick example is broken


arkadyan:
On the [Candlestick example page](http://uber.github.io/react-vis/#/examples/charts/candlestick) the source code link is broken. It currently points to `https://github.com/uber/react-vis/blob/master/examples/candlestick/candlestick.js`, but the correct address is `https://github.com/uber/react-vis/blob/master/showcase/examples/candlestick/candlestick.js`.

PR #556, opened 2017-08-16
Fix the Candlestick example source code link

arkadyan:
Fixes https://github.com/uber/react-vis/issues/555


mcnuttandrew:
Thanks!

PR #559, opened 2017-08-18
Use component name as fallback for display name

sqren:
When using the HOC `makeWidthFlexible` the resulting component will look like this:

```html
<Flexibleundefined ...>...</Flexibleundefined>
```

It can be fixed by explicitly setting a displayName:

```js
MyComponent.displayName = 'MyComponent';
makeWidthFlexible(MyComponent);
```

I think the `makeWidthFlexible` HOC should fallback to use the component name, if no display name is given. This will always be more useful than `undefined`

### DIFF
--- a/src/make-vis-flexible.js
+++ b/src/make-vis-flexible.js
@@ -86,6 +86,15 @@ function removeSubscriber(cb) {
 }
 
 /**
+ * Helper for getting a display name for the child component
+ * @param {*} Component React class for the child component.
+ * @returns {String} The child components name
+ */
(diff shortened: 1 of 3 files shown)


mcnuttandrew:
This seems reasonable! Mind writing a test to prevent regression?

sqren:
Sure thing. I'll get back to this on Monday.

sqren:
I didn't see any tests for the higher order components, so created a new file. Not sure if it fits in. Let me know what you think.

PR #562, opened 2017-08-23
Voronoi: Add mouseUp and mouseDown events

sqren:
This PR adds mouseUp and mouseDown events, which make it much easier to implement dragging on the Voronoi componont.

I didn't see any tests for the Voronoi component, so not sure what style of test this enhancement would benefit from (if any).

### DIFF
--- a/showcase/examples/dragable-chart/drag-marker.js
+++ b/showcase/examples/dragable-chart/drag-marker.js
@@ -6,6 +6,7 @@ function DragMarker({innerHeight, marginTop, marginLeft, x, x2}) {
   const start = x2 > x ? x : x2;
   return (
     <rect
+      className="drag-marker"
       pointerEvents="none"
       fill="black"
       opacity="0.2"
--- a/showcase/examples/dragable-chart/drag-marker.js
+++ b/showcase/examples/dragable-chart/drag-marker.js
@@ -1,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function DragMarker({innerHeight, marginTop, marginLeft, x, x2}) {
+  const width = Math.abs(x2 - x);
+  const start = x2 > x ? x : x2;
+  return (
+    <rect
+      pointerEvents="none"
+      fill="black"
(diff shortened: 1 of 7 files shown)


mcnuttandrew:
I like what you've got going here! I'm really curious to see an implementation of dragging using react-vis, it would make a great addition to the examples folder + docs.

As for tests, there are some tests for the voronoi component but they are embedded in the mark series tests. One testing approach we use quite a bit is to write tests against components in the showcase folder. I think starting a new voronoi-tests file and writing a test against one of the existing examples would be a great move.

If you wanted to also gain a million points, you could also start a voronoi doc file, but I'd understand if that's a lot to ask.

sqren:
I've added an example for dragging. It's similar to the [zoomable chart example](https://github.com/uber/react-vis/blob/master/showcase/examples/zoomable-chart/zoomable-chart-example.js) but uses Voronoi to get more control over the dragable area.

Do it fit into the codebase? Let me know what you think.

mcnuttandrew:
I'm only on mobile for another week or so, can someone else give this a full review? (@apercu @jckr ?)

jckr:
I'm generally ok with this, however:
Voronoi today has a special status in the sense that it's not a full-fledged series like LineSeries, MarkSeries etc. and as such it's not an extension of AbstractSeries and it doesn't have an API similar to these Series. 
I feel that eventually Voronoi should graduate to Series status; maybe become VoronoiSeries and be deprecated. When that's the case, all interaction methods would have to follow the series syntax - onValueClick, onSeriesClick etc. and then, it would make more sense to add interaction handlers at the AbstractSeries level than reimplement this for specific series.

sqren:
@jckr I saw you published 1.7.3 but the Voronoi component has not been updated in the npm module. Is there a build step missing?

Apercu:
Hey @sqren Can you try `1.7.4`?

sqren:
@Apercu Works like a charm. Thanks!

Apercu:
Sure, have a nice day ;D

PR #563, opened 2017-08-25
handle right clicks to series and marks

jckr:
this PR allows users to handle right clicks to either marks or series. Each time a vis can handle an onValueClick, there is now an onValueRightClick; likewise, every series that supported onSeriesClick now supports onSeriesRightClick. 

documentation updated. 
no new tests are provided, since onValueClick wasn't tested per se, but it doesn't break existing tests.

### DIFF
--- a/docs/markdown/line-series.md
+++ b/docs/markdown/line-series.md
@@ -78,6 +78,11 @@ An object which holds CSS properties that will be applied to the SVG element(s) 
 
 ### Interaction handlers
 
+Note - interacting with a line may be difficult especially with the standard width. To address that, consider:
+- the proximity handlers - onNearestX, onNearestXY;
+- increasing the width of your line to make it easier to reach with the mouse,
+- creating a near-transparent line series with extra width to catch mouse events. 
+
 #### onNearestX (optional)
 Type: `function(value, {event, innerX, index})`  
 A callback function which is triggered each time the mouse pointer moves. It can access the datapoint of the mark whose x position is the closest to that of the cursor. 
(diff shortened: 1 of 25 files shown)


docs/markdown/line-series.md
```diff
@@ -78,6 +78,11 @@ An object which holds CSS properties that will be applied to the SVG element(s)
 
 ### Interaction handlers
 
+Note - interacting with a line may be difficult especially with the standard width. To address that, consider:
```
mcnuttandrew: Plus one

jckr: 1) do i need right click interaction for sankey and treemap - maybe, but they don't follow the onValueClick / onSeriesClick model to begin with. So that suggests this should be a separate effort. 
2) opportunity for more tests, I agree but again I think there's need for something more strategic than bundled with this specific PR. also at the time of this PR there were 999 tests for the record
3) documentation super non dry - what i'm thinking is moving documentation content from flat markdown file to a more database like structure so we could render doc pages based on props and have a finer control of look and feel for what's a series prop description vs what's explanatory text. right now, everything is a title or not, bold or not, italics or not and that's it. ironic for front end people.

mcnuttandrew:
> So that suggests this should be a separate effort.

I'm fine with that, just make an issue and assign yourself so there isn't a forgotten hole in the library

> I agree but again I think there's need for something more strategic than bundled with this specific PR.

Another good issue, I have some thoughts on the matter, feel free to make an issue and assign me.


> right now, everything is a title or not, bold or not, italics or not and that's it.

Actually that's not the case, you can use class injection (see docs app) as well as inline HTML! You can do anything! Sky is the limit!


> documentation content from flat markdown file to a more database like structure so we could render doc pages based on props and have a finer control of look and feel for what's a series prop description vs what's explanatory text.

I am pretty strongly against this. Part of the reason we have the current setup is bc we don't want to maintain a bunch of custom HTML and other mechanisms to render the docs. Another part is that a number of people still appear to read and reference the docs from the markdown itself. i feel like most libraries write markdown for docs, so we should keep step with the commonly held practice, both for discovery and for ease of contribution.

Sorry for weird formatting, mobile reviewing is hard

Issue #565, opened 2017-08-25
React-vis requiresSVG  doesn't work properly with Preact


mrawdon:
When using Preact stateless components are rendered outside of the SVG element

![image](https://user-images.githubusercontent.com/1678756/29731804-e4c2b7fc-89b3-11e7-9ce9-108084a71bbb.png)

![image](https://user-images.githubusercontent.com/1678756/29731844-190fb23a-89b4-11e7-8c16-3b6bb249a0c3.png)

Preact makes a copy of any stateless component.  When doing that it only copies the documented React class properties https://facebook.github.io/react/docs/react-component.html#class-properties
This occurrs in Preact at https://github.com/developit/preact-compat/blob/master/src/index.js#L264

A simple solution for this is moving requiresSVG  and isSVG into defaultProps.  I've implemented this and will create a Pull Request.  Not sure if this will result in confusion though, as these aren't properties that should ever be changed from the default.

chrisirhc:
Could we use the `statics` API instead? See https://github.com/developit/preact-compat/issues/47 .

Relevant lines:
https://github.com/developit/preact-compat/blob/e25d0607d8a00a3e2f9229f33de4dc0f63c8e931/src/index.js#L417-L419

mrawdon:
I wasn't going to use defaultProps to do the check in xy-plot I use the props object(just needed a way to set the default)  https://github.com/Mitrend/react-vis/blob/master/src/plot/xy-plot.js#L99

So it can still be controlled by the user

developit/preact-compat#47 seems pretty specific to the children property so not really sure that its relevant here.

chrisirhc:
> So it can still be controlled by the user

That's the thing, it shouldn't be controlled by the user per usage of the component. It should be controlled by the developer of the component.

mrawdon:
Sorry I misunderstood your point earlier. I agree a user shouldn't be modifying these  but it doesn't seem like the functional components api gives any other options. 

We opened #569 to work out the documentation.

@mcnuttandrew any thought here?

mcnuttandrew:
Generally I agree with @chrisirhc that the user shouldn't be able to modify this prop. I'm not sure what the right solution is! 

As i noted above, I feel like stashing the props of interest under a underscore generally reads to users that they shouldn't mess with them? Eg _isCanvas is probably not something a user will attempt to use. We could also throw warnings if the default props are different than the recieved props, however this usage feels less than clean.

mrawdon:
[...]

PR #566, opened 2017-08-25
Moved requiresSVG to defaultProps for preact support

mrawdon:
Addresses issue #565

### DIFF
--- a/src/plot/axis/axis-ticks.js
+++ b/src/plot/axis/axis-ticks.js
@@ -33,11 +33,13 @@ const propTypes = {
     LEFT, RIGHT, TOP, BOTTOM
   ]).isRequired,
   style: PropTypes.object,
-  width: PropTypes.number.isRequired
+  width: PropTypes.number.isRequired,
+  requiresSVG: PropTypes.bool
 };
 
 const defaultProps = {
-  style: {}
+  style: {},
+  requiresSVG: true
 };
 
 function _getTickFormatFn(scale, tickTotal, tickFormat) {
@@ -209,6 +211,5 @@ class AxisTicks extends React.Component {
 AxisTicks.defaultProps = defaultProps;
 AxisTicks.displayName = 'AxisTicks';
 AxisTicks.propTypes = propTypes;
-AxisTicks.requiresSVG = true;
 
 export default AxisTicks;
--- a/src/plot/axis/axis.js
+++ b/src/plot/axis/axis.js
@@ -69,7 +69,8 @@ const propTypes = {
   marginLeft: PropTypes.number,
   marginRight: PropTypes.number,
   innerWidth: PropTypes.number,
-  innerHeight: PropTypes.number
+  innerHeight: PropTypes.number,
+  requiresSVG: PropTypes.bool
 };
 
 const defaultProps = {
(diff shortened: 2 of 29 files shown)


mcnuttandrew:
Hey @mrawdon,

This is quite a PR! I had no idea anyone was trying to use react vis with preact, very cool. I support this change with a couple of suggestions

- currently this PR is a major breaking change, I think both formats could be supported until the next major version jump by checking to see if .type is defined while filtering

- in order to reduce confusion and potential hazard for users, maybe add an underscore in front of the prop name, i.e. _requiresSVG in order to suggest that users not use it.

I also have some nits

- make sure the new props are alphabetized correctly
- update the documention to reflect this change

Overall, super cool work

mrawdon:
Makes sense to me.  I can make those changes.  

- For alphabetizing if I add and underscore to the property name(_requiresSVG).  Would you still alphabetize it based on the r of require?  Or just move it to the end?
- For documentation are there any other examples where you have documented internal properties?  I found some internal props in Axis (https://github.com/uber/react-vis/blob/master/src/plot/axis/axis.js#L66)  that don't appear to be documented anywhere.  Not trying to object to documentation, just not sure where/how to document this.

mcnuttandrew:
I'd say _requiresSVG is categorized under r. 

I thought there was a docs page on custom components but on second look, I guess not? Will you open an issue about docs for how to write custom components?

mcnuttandrew:
Additional thought: it seems like it would be good to write an integration test proving that react-vis works well with preact, so that we don't accidentally cause regressions later.

mrawdon:
I got the tests to run against preact.  There's a bunch of them failing right now, I'm working through those now.

mrawdon:
So I just committed the changes we discussed. Around the prop types.  

I opened issue #569 around the documentation 

I also added a way to run both the tests and showcase with preact. Right now these are pretty much all failing due mostly to this bug I opened with preact-compat developit/preact-compat#393
There also seems to be a problem with the preact-compat-enzyme stuff, where it doesn't seem to be rendering the tests the same way as react with enzyme.

Let me know what you think, especially around adding preact into the package.json, wasn't really sure if that belonged here.

mcnuttandrew:
Hey @mrawdon, I'm trying to checkout your PR locally, and i can't get the tests to pass? Also it's pretty cumbersome to get the examples to run in preact? 

Relatedly, would doing as is suggested in #589 and moving all of the isCanvas/requiresSVG methods to be static gets fix the situation?

mrawdon:
Answers to your 3 points below

1.  Are tests not running all for you?  Or do they fail with the exception below?  That issue can be fixed by aliasing react-motion to preact-motion.  There's also another change that needs to be made to work around developit/preact-compat#419
Even with these fixes tests will eventually blow up with differences between enzyme and preact-compat-enzyme. 
```
/home/mike/react-vis2/node_modules/react-motion/lib/Motion.js:241
    var renderedChildren = this.props.children(this.state.currentStyle);
                                      ^

TypeError: this.props.children is not a function
    at render (/home/mike/react-vis2/node_modules/react-motion/lib/Motion.js:241:39)
    at renderComponent (/home/mike/react-vis2/node_modules/preact/src/vdom/component.js:89:3)
    at renderComponent (/home/mike/react-vis2/node_modules/preact/src/vdom/component.js:114:21)
    at renderComponent (/home/mike/react-vis2/node_modules/preact/src/vdom/component.js:114:21)
```
2. The showcase won't render correctly without a fix for developit/preact-compat#419 and the react-motion issue(I have them working locally I can push those into the PR if you want, just not sure how to break this stuff up)

3.  The issue is only with functional components, so I don't see how adding statics will work there.

mrawdon:
Any thoughts here?

PR #571, opened 2017-08-30
RadarChart: support colorRange prop

targos:
It is already defined in the default props but the rendering
unconditionally uses the default color range.


mcnuttandrew:
👍  looks big reasonable, i'll merge this in as soon as the CLA passes. If you've already signed it and it's stuck, push up a bump commit to re-initiate the check.

targos:
I've signed the CLA.

Issue #528, opened 2017-08-06
Documentation Search


mcnuttandrew:
As we gradually add a larger and larger corpus of documents, it would be convenient if there was a mechanism for search across the entire documentation. To wit, it is really annoying to try to find 
where the legends documentation is. 

I am not sure what the right technique for adding this functionality would be, but an initial pass would enable search across article names.

PR #572, opened 2017-08-30
Docs search

mcnuttandrew:
This PR adds search functionality to the documentation. The basic architecture of this feature works as follows.

1) Before the app arrives in the browser all of the markdown files are compiled into a single index.json file
2) On load of app in the browser a webWorker is launched who grabs the index.json file
3) Individual queries are sent to the web worker for search.

search

This addresses #528


docs/markdown/getting-started/getting-started.md
```diff
@@ -1,120 +0,0 @@
-### Getting started
```
mcnuttandrew: this file wasn't being used

Apercu:
Ah this is so cool! I'm in the process of creating some sort of project allowing us to bootstrap our documentation website, this will definitely belong in there.

I'll review and accept, dunno how much time it will take to port react-vis to it

mcnuttandrew:
> I'm in the process of creating some sort of project allowing us to bootstrap our documentation website, this will definitely belong in there.

I've been thinking about a project like that! I think a bootstrapping thing could be cool, but I feel like this who structure could be extracted into a gitbook like framework. I imagine such a structure being something analogous to a react component that accepts a JSON list of pages (delimited by section) and style props, and requires the user to keep all of their markdown in a single place. Were i to name to this something, it would be DocsBuilder.

Apercu:
Pardon my wording, it's actually not only a scaffolding tool but also more like a create-react-app sort of thing, abstracting all the configs and base stuff like the redux provider and routes. I'm gonna add you to the repo

PR #578, opened 2017-08-31
RadarChart: fix passing animation prop to label series

targos:


targos:
I found this while debugging a strange behavior of my chart. See https://github.com/uber/react-vis/issues/579

targos:
My pleasure. Thanks for the excellent library!

Issue #573, opened 2017-08-30
Scaling large yRange values


Falieson:
I'm using react-vis 1.7.4
My yRange is [199740, 200400] and I'm just getting a wall of bars w/ no relative difference because everything is being scaled compared to a yRange of [0, 200400]

https://www.webpackbin.com/bins/-Ksoencaj2nvtlGMdKtA 

How to fix this? I tried declaring the yRange but that just made the entire graph disappear.

Thanks for your help...

mcnuttandrew:
Hey @Falieson 

You need to modify your example just a little bit to make it work. First you need to be using yDomain, not yRange. domain refers to the scope of the values that are coming in, and range refers to the pixels being output. Secondly there is a weird quirk with modifying where the barChart "starts", in that you need to provide a "baseValue". So all told your chart should look like this


```javascript
 return (
      <XYPlot
        height={480}
        xType='time'
        width={500}
yDomain={[yRange.min, yRange.max]}
yBaseValue={yRange.min}
      >
        <HorizontalGridLines />
        <VerticalBarSeries data={xyData} />
        <XAxis />
        <YAxis />
      </XYPlot>
    )
```

That said, I'm a frequent contributor of this library and that took me quite a while to figure out. I going to open a PR to remove the baseValue property because it seems like it is redundant to domain.

mcnuttandrew:
Should now be resolved in 1.7.5!

Falieson:
Thank you!

PR #581, opened 2017-08-31
Make baseValue copy domain

mcnuttandrew:
As highlighted in #573 it is really confusing that baseValue takes over an inserts itself into different series behaviour. This PR modifies baseValue so that if the user specifies a domain, the base value is equal to the bottom of that domain. 

There is some mess, in that this PR also has #580 in it, though as soon as that lands this can be clean up too.


Apercu:
Oh btw is there something special with the sankey docs? I have a d3 interpolate issue

mcnuttandrew:
> Oh btw is there something special with the sankey docs? I have a d3 interpolate issue 

not that I know of? what are you seeing?

Issue #430, opened 2017-05-24
Refactor Sankey


Apercu:
The Sankey component should use `XYPlot`, `PolygonSeries`, `RectSeries` and a `LabelSeries`

mcnuttandrew:
The refactor is in!

PR #587, opened 2017-09-04
Rewrite sankey

mcnuttandrew:
This PR rewrites and restructure the sankey diagram (resolving #430). 

- Importantly: this change to follow the pattern of the rest of the library will allow for the easier development of new features in the sankey diagram. (Eg tool tips should now be trivial to add)
- Amusingly: this also allows for animation

sankey

The main prompting for this re-write is that the package we were using for sankeying (@apercu's temporary d3-sankey-align) has a version mismatch with d3-interpolate that causes it crash.


docs/markdown/sankey.md
```diff
@@ -1,8 +1,8 @@
 # Sankey
 
-Note: This component is in alpha.
+Sankey diagrams are a form of graph that allows for the easy communication of flows and other transferal processes.
 
-<!-- INJECT:"BasicSankeyExample" -->
+<!-- INJECT:"EnergySankey" -->
```
Apercu: (Sankeyku)

docs/markdown/sankey.md
```diff
@@ -65,57 +68,94 @@ match the height of the path link.
...
+Type: `Boolean`  
+Defaults: `false`.
 Hide the display of the node names if specified to true.
 
-Defaults to `false`.
 
-##### onClick
+##### onValueClick (optional)
+Type: `function`  
+Default: none
```
Apercu: maybe `noop` instead of `none`?

package.json
```diff
@@ -44,7 +44,7 @@
     "d3-geo": "^1.6.4",
     "d3-hierarchy": "^1.1.4",
     "d3-interpolate": "^1.1.4",
-    "d3-sankey-align": "^0.1.0",
```
Apercu: cool I can delete this

src/sankey/index.js
```diff
@@ -1,153 +1,198 @@
...
-      onBlur,
-      onClick,
-      onHover,
+      onValueClick,
+      onValueMouseOver,
+      onValueMouseOut,
+      style,
       width
     } = this.props;
+    const nodesCopy = [...new Array(nodes.length)].map((e, i) => ({...nodes[i]}));
```
Apercu: Not sure why we need to force copying the nodes and links? Shouldn't that be handled by the app itself rather than the component?

src/sankey/index.js
```diff
@@ -1,153 +1,198 @@
...
+    const linksCopy = [...new Array(links.length)].map((e, i) => ({...links[i]}));
+
+    const formattedMargin = {
```
Apercu: nit: Maybe we should create an helper out of this, I'm guessing we are using it somewhere else right?

src/sankey/index.js
```diff
@@ -1,153 +1,198 @@
...
+            data={path(link)}
+            opacity={link.opacity || linkOpacity}
+            color={link.color}
+
+            strokeWidth={Math.max(link.width, 1)}
+            node={link}
+            nWidth={nWidth}
+            key={`link-${i}`}/>
+        ))
+        }
```
Apercu: don't you think the bracket should be on the same line as the `))`? Also I don't get the purpose of the following bracket

src/sankey/index.js
```diff
@@ -1,153 +1,198 @@
...
+        {
+          <VerticalRectSeries
+            animation={animation}
+            className={`${className} rv-sankey__node`}
+            data={nodesCopy.map(node => {
```
Apercu: returning the object directly would avoid one level of nesting

src/sankey/index.js
mcnuttandrew: I think the idea is that we want the users to have to do as little configuration as possible, if we require them to make copies just for us, that will most likely be sort of error prone?

PR #588, opened 2017-09-05
Crosshair: allow to specify styles in props

targos:
I can do the same for the line width, if that makes sense.


targos:
@mcnuttandrew I updated the PR to use style objects. Do you think I should move the default styles to JS? It may be best to keep them in CSS, otherwise I think it breaks the possibility for the users to override them in their own stylesheets.

PR #591, opened 2017-09-06
XYPlot: add support for various mouse events

targos:
Adds onClick, onDoubleClick and onWheel event handlers.


targos:
I need the double-click and mouse wheel events to control the zoom of my component:
- double click to reset the zoom (full zoom out)
- wheel to zoom in (extend or reduce the scale of the Y axis)

PR #596, opened 2017-09-07
Ocular docs

Apercu:
I removed the showcase app too since we can easily debug a component with the ocular docs, and was a bit redundant in my opinion


Apercu:
Also I was wondering why we had two different showcase indexes?

website/src/config.js
```diff
@@ -0,0 +1,54 @@
...
+export const PROJECT_DESC = 'A composable charting library';
+
+export const PROJECTS = {};
+
+export const HOME_HEADING = 'A composable charting library';
+
+export const HOME_RIGHT = (
+  <div>
+    <div className="m-bottom">
+      <ComplexChart forFrontPage />
```
mcnuttandrew: maybe its time to put something better in here than complex chart?

website/src/styles/_variables.scss
```diff
@@ -0,0 +1,13 @@
...
+$black-20: #213746;
+$black-40: #5A666D;
+$white: #EDEDED;
+$white-40: #8D9BA3;
+
+$footer-height: 13rem;
+$topbar-height: 4rem;
+$topbar-maxheight: 12rem;
+$toc-width: 17rem;
+$mobile: 576px;
```
mcnuttandrew: I think it would be better if components were centered on the page? Like, 

screen shot 2017-09-07 at 10 57 07 am

Looks weird to me

Apercu:
If you think the showcase is better for developing I can add it back :)

Yes @jckr is coming up with an image, this is only temporary.

As for the testing, not really sure if that's the priority, but yes could be something to add.

Apercu:
@mcnuttandrew Just re-added the showcase app

Issue #593, opened 2017-09-07
cleanseData is a performance bottleneck


targos:
With relatively big series (in my case: 100k points), about half of the (re)render time is spent in the `cleanseData` function to prepare for the deep equality check. This also creates a lot of temporary objects that have to be garbage collected.
My first suggestion would be to stop comparing deeply the data arrays, but maybe I'm missing a valid use case?
The problem is that this check happens everytime the component is going to be rerendered, even if the series haven't changed. For example when the domain of an axis is changed to zoom in or out.
I would be happy to work on optimizing this if it can be done without breaking changes.

https://github.com/uber/react-vis/blob/6c5f7841fccaa0214630472c3137853161340b99/src/plot/xy-plot.js#L65-L72

![image](https://user-images.githubusercontent.com/2352663/30165883-f57871b2-93e1-11e7-8f7d-9547db9b41ef.png)

targos:
Fixed in https://github.com/uber/react-vis/pull/597. Thanks.

PR #597, opened 2017-09-07
Only try to clean data when asked

mcnuttandrew:
When building the sunburst I added a function called cleanseData which makes a copy of every element under consideration in the xy-plot. This is useful for sunburst and treemap as it allows us to remove cyclic structures in the json, however as noted in #593, this can cause some serious performance bottle neck. This PR delegates that functionality under a flag to keep things moving quickly unless necessary.

### DIFF
--- a/docs/xy-plot.md
+++ b/docs/xy-plot.md
@@ -119,6 +119,10 @@ Height of the component. The height should be passed.
 Type: `string`  
 DOM classNames to be added to the wrapper component.
 
+#### hasTreeStructure (optional)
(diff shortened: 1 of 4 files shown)


Apercu:
Since I moved the markdown docs, we might have a small conflict to resolve here

mcnuttandrew:
> Since I moved the markdown docs, we might have a small conflict to resolve here

This PR can just wait to land until the ocular one does, ill rebase once that happens

Issue #360, opened 2017-04-07
Accessors


mcnuttandrew:
It is annoying to have to use react-vis built in language of xs, ys, and sizes! I want to be able to provide accessor functions that will allow me to keep my objects formatted in way that is ledgible to my data context. In addition this move would allow us to accept immutable objects, which could speed up rendering by causing the equality check that occurs in XYPlot to happen more easily

mcnuttandrew:
Apparently this is one out big deficits: https://www.reddit.com/r/reactjs/comments/6bbj2g/data_visualization_in_react/

PR #601, opened 2017-09-08
Accessors

mcnuttandrew:
This sprawling PR adds support for accessors throughout the library, thus addressing #360. It modifies the way that scales are built, by replacing a default notion of attr with a comprable notation of accessor, while adding default accessors that emulate the original behaviour. (Hopefully) This change is non-breaking, in that all old charts should still work, with the new accessor functionality enabled. 

Note: one of the original goals of adding accessors was to enable immutable manipulations, and while were close, there is probably one or two more PRS worth of clean up around the library that needs to be done first. (i'm looking at you, several for loops still left in the library)

There still needs some documentation, but should be ready to review.

### DIFF
--- a/src/sunburst/index.js
+++ b/src/sunburst/index.js
@@ -57,12 +57,12 @@ const LISTENERS_TO_OVERWRITE = [
    props.height {number} - the height of the graphic to be rendered
    props.hideRootNode {boolean} - whether or not to hide the root node
    props.width {number} - the width of the graphic to be rendered
-   props.sizeAccessor {function} - accessor for the size
+   props.getSize {function} - accessor for the size
  * @returns {Array} Array of nodes.
  */
-function getNodesToRender({data, height, hideRootNode, width, sizeAccessor}) {
+function getNodesToRender({data, height, hideRootNode, width, getSize}) {
   const partitionFunction = partition();
-  const structuredInput = hierarchy(data).sum(sizeAccessor);
+  const structuredInput = hierarchy(data).sum(getSize);
   const radius = (Math.min(width, height) / 2) - 10;
   const x = scaleLinear().range([0, 2 * Math.PI]);
   const y = scaleSqrt().range([0, radius]);
@@ -94,17 +94,17 @@ function getNodesToRender({data, height, hideRootNode, width, sizeAccessor}) {
  */
 function buildLabels(mappedData, accessors) {
   const {
-    angleAccessor,
-    angle0Accessor,
-    labelAccessor,
-    radius0Accessor
+    getAngle,
+    getAngle0,
+    getLabel,
+    getRadius0
   } = accessors;
 
   return mappedData
-  .filter(labelAccessor)
+  .filter(getLabel)
   .map(row => {
-    const truedAngle = -1 * angleAccessor(row) + Math.PI / 2;
-    const truedAngle0 = -1 * angle0Accessor(row) + Math.PI / 2;
+    const truedAngle = -1 * getAngle(row) + Math.PI / 2;
+    const truedAngle0 = -1 * getAngle0(row) + Math.PI / 2;
     const angle = (truedAngle0 + truedAngle) / 2;
     const rotateLabels = !row.dontRotateLabel;
     const rotAngle = -angle / (2 * Math.PI) * 360;
@@ -114,8 +114,8 @@ function buildLabels(mappedData, accessors) {
       children: null,
       angle: null,
       radius: null,
-      x: radius0Accessor(row) * Math.cos(angle),
-      y: radius0Accessor(row) * Math.sin(angle),
+      x: getRadius0(row) * Math.cos(angle),
+      y: getRadius0(row) * Math.sin(angle),
       style: {
         textAnchor: rotAngle > 90 ? 'end' : 'start',
         ...row.labelStyle
@@ -132,28 +132,28 @@ const NOOP = () => {};
 class Sunburst extends React.Component {
   render() {
     const {
-      angleAccessor,
-      angle0Accessor,
+      getAngle,
+      getAngle0,
       animation,
       className,
       children,
       data,
       height,
       hideRootNode,
-      labelAccessor,
+      getLabel,
       width,
-      sizeAccessor,
+      getSize,
       colorType
     } = this.props;
-    const mappedData = getNodesToRender({data, height, hideRootNode, width, sizeAccessor});
+    const mappedData = getNodesToRender({data, height, hideRootNode, width, getSize});
     const radialDomain = getRadialDomain(mappedData);
     const margin = getRadialLayoutMargin(width, height, radialDomain);
 
     const labelData = buildLabels(mappedData, {
-      angleAccessor,
-      angle0Accessor,
-      labelAccessor,
-      radius0Accessor: d => d.radius0
+      getAngle,
+      getAngle0,
+      getLabel,
+      getRadius0: d => d.radius0
     });
 
     const hofBuilder = f => (e, i) => f ? f(mappedData[e.index], i) : NOOP;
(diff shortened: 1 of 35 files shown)


mcnuttandrew:
> In deck.gl accessors are consistently called get...: getSize, getAngle etc. (although getX sounds a little strange?)

I guess react-vis has a slightly different API pattern? Generally when specifying a property about a dimension (eg the type of scale to be used or the domain) we specify that as xType or xDomain. I'm not opposed to copying deck.gl's api style, I just felt like the right move was to be consistant within the library first, and change later?

ibgreen:
> the right move was to be consistant within the library first, and change later?

That does make sense. Although given today's conversation I think we'd want to decide and possible change before we publish an official version. 

I am not sure how the branching strategy in react-vis works, how long master can live without being published etc. If we land this, are we creating an urgency to decide and update?

mcnuttandrew:
> I am not sure how the branching strategy in react-vis works, how long master can live without being published etc. If we land this, are we creating an urgency to decide and update?

Generally our strategy has been to develop on master, however if we want this to be part of some sort v2 release, we could hold all new pieces on a separate branch. I'm in no huge rush to get this is, so whatever is fine.

mcnuttandrew:
Bump for review?

ibgreen:
@mcnuttandrew Based on conversations with @jckr I came away with the sense that we want accessor names to be aligned with the suite (i.e. deck.gl) from the start, rather than introduce one set of names one month and then introduce a completely different set next month. At the same time, the deck.gl alignment might (at least initially) only cover a small part of the charts.

mcnuttandrew:
While I appreciate that it is good to have things consistent across the suite, I think it is, perhaps, more important to have them be consistent within each library. The react-vis pattern is xType, xDomain, etc, so it seems like it makes sense to continue that into xAccessor? I guess if you really wanted to be consistent across the suite, react-vis would need to move to a naming convention like getXType getXDomain etc? But that seems kinda awkward to me?

I don't have a strong opinion, I'd mostly just like to get this merged in so that it doesn't to be defended against merge conflicts.

ibgreen:
>  I guess if you really wanted to be consistent across the suite, react-vis would need to move to a naming convention like getXType getXDomain etc? But that seems kinda awkward to me?

Yes that is what is being proposed. In effect, we are trying to understand whether react-vis should be part of the suite or live alone. If part of the suite, there needs to be a story and a plan that gradually ties things closer together. No decisions made, just discussions at this point. I don't have specific goal in mind, I am mostly getting involved because I think react-vis deserves a strong vision and roadmap, and suite alignment could provide that.

jckr:
Yeah that's the plan. if we do add accessors which your PR allows us to do then it's a great opportunity to move towards suite alignment and to revise the whole syntax of the API.

mrawdon:
From a user standpoint this PR(in what ever form) would be a big help, and save a lot of extra data manipulation.

ibgreen:
> From a user standpoint this PR(in what ever form) would be a big help, and save a lot of extra data manipulation.

> if we do add accessors which your PR allows us to do then it's a great opportunity to move towards suite alignment and to revise the whole syntax of the API.

I took a second look. My take is that if we just update this PR to follow the `getX` instead of `xAccessor` naming pattern, this would be a major move towards suite alignment and we could do additional syntax revision in a second step.

We could get the accessors into react-vis today in a backwards compatible minor version with minimal effort, score a big alignment win, and save a bigger API overhaul for a later major version.

That said, if I was reviewing another framework in our suite, I would ask that the component doc pages and a "What's New" doc page to be updated in this PR :)

mcnuttandrew:
@ibgreen Those are reasonable notes! I can take a shot at getting this wrapped up this weekend.

mcnuttandrew:
Alright, I have moved all of the accessors over to the getX pattern. Currently this PR is non-breaking, but so it can be slid in without a fuss. I didn't update the documentation in a more sprawling way, because I figured we'd only want to do that once every property had been moved to the getXType pattern. That said, I still want to argue against using getXType: getX makes sense you are providing a function to get something, while getXType does not, XType is a static string that won't change over time.

- There's still a couple of things still to do, like moving nullAccessor to be getNull, and I've open an issue to that effect (#661).
- Following @ibgreen's suggestion i've also open an issue to make a what's new page (#662)

mcnuttandrew:
bump for review (@balthazar or @ibgreen or ...?)

src/utils/scales-utils.js
```diff
@@ -177,19 +186,19 @@ export function getScaleFnFromScaleObject(scaleObject) {
 /**
  * Get the domain from the array of data.
  * @param {Array} allData All data.
- * @param {string} attr Property name.
+ * @param {function} accessor - accessor for main value.
+ * @param {function} accessor0 - accessor for the naught value.
```
ibgreen: Something about these names bother me, especially `accessor0`. Maybe call the params, `mainAccessor`, `zeroAccessor`

Haven't read carefully, but why do we have an accessor for zero as opposed to just supplying a value? Assuming this is a single value?

+1 for dropping `attr`, by the way.

src/sunburst/index.js
```diff
@@ -175,20 +193,29 @@ class Sunburst extends React.Component {
 Sunburst.displayName = 'Sunburst';
 Sunburst.propTypes = {
   animation: AnimationPropType,
+  getAngle: PropTypes.func,
+  getAngle0: PropTypes.func,
```
ibgreen: See my comment below about  names ending with `0`.

src/sunburst/index.js
```diff
@@ -88,14 +89,22 @@ function getNodesToRender({data, height, hideRootNode, width}) {
...
  * Important to use mappedData rather than regular data, bc it is already unrolled
  * @param {Array} mappedData - Array of nodes.
+ * @param {Object} accessors - object of accessors
  * @returns {Array} array of node for rendering as labels
  */
-function buildLabels(mappedData) {
+function buildLabels(mappedData, accessors) {
+  const {
+    getAngle,
+    getAngle0,
```
ibgreen: See comment below about names ending with `0`.

src/utils/scales-utils.js
mcnuttandrew: So to clarify you are suggesting a naught accessor like zeroX and zeroAngle?

ibgreen: > To to clarify you are suggesting a naught accessor like zeroX and zeroAngle?

To be honest, I am not completely sure what I am proposing as I don't fully understand what the zero accessors are doing. Mainly reacting to the name which looks unusual and not immediately obvious to a new comer.

First of all, I am asking if `accessor0` needs to be an accessor, e.g.
* Is `accessor0` called on every item, then yes,
* but if `accessor0` is called just once, to get a single "zero" value as reference point for the series, could we just supply the zero value directly rather than ask for another accessor?

If only once, then yes, (value) props could be called ` zeroX` and `zeroAngle`

If it needs to be an accessor then my case is weaker. It could still be worth at least looking at naming options, like `getZeroX` and `getZeroAngle`.

mcnuttandrew: The zero accessor tend to come up in a few particular types of cases. For instance if you have rectangle expressed through RectSeries, you'd express the bottom of the rectangle through getY0 and the top through getY. Similarly, if you were working with an ArcSeries, getAngle0 would express the start angle of the wedge and getAngle would express the end angle. Each of these refers to values found on the data eg {angle0: Pi, angle: Pi / 2} or {y0: 2, y: 4}, so it'd be a pretty big change to alter this naming convention too.

ibgreen: Got it. Starting fresh, I would personally favor changing the expected data fields to `startAngle` etc. But agree that staying consistent now makes the most sense.

Perhaps something to consider in the API audit @jckr is planning.

PR #605, opened 2017-09-11
Use window.devicePixelRatio to sharpen canvas images

stronglink:
Using `devicePixelRatio` allows us to improve the visual quality of canvas renderings by accounting for the ratio between physical and virtual pixels. Most notably, canvas charts will look much better on Apple's Retina Displays.

**Before**
screen shot 2017-09-11 at 3 00 58 pm

**After**
screen shot 2017-09-11 at 3 00 51 pm

### DIFF
--- a/src/plot/series/canvas-wrapper.js
+++ b/src/plot/series/canvas-wrapper.js
@@ -105,6 +105,12 @@ function buildLayers(newChildren, oldChildren) {
 }
 class CanvasWrapper extends Component {
 
+  static get defaultProps() {
+    return {
+      pixelRatio: window && window.devicePixelRatio || 1,
+    };
+  }
+
   componentDidMount() {
     const ctx = this.refs.canvas.getContext('2d');
     const {pixelRatio} = this.props;
--- a/src/plot/series/canvas-wrapper.js
+++ b/src/plot/series/canvas-wrapper.js
@@ -107,7 +107,7 @@ class CanvasWrapper extends Component {
 
   static get defaultProps() {
     return {
-      pixelRatio: window && window.devicePixelRatio || 1,
+      pixelRatio: window && window.devicePixelRatio || 1
     };
   }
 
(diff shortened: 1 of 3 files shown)


src/plot/series/canvas-wrapper.js
```diff
@@ -104,8 +104,13 @@ function buildLayers(newChildren, oldChildren) {
   });
 }
 class CanvasWrapper extends Component {
+
   componentDidMount() {
-    this.drawChildren(this.props, null, this.refs.canvas.getContext('2d'));
+    const ctx = this.refs.canvas.getContext('2d');
+    const { pixelRatio } = this.props;
```
mcnuttandrew: I think your build is failing bc it expects `const {pixelRatio} = this.props;`

src/plot/xy-plot.js
```diff
@@ -413,7 +413,8 @@ class XYPlot extends React.Component {
       marginBottom,
       marginRight,
       innerHeight,
-      innerWidth
+      innerWidth,
+      pixelRatio: window.devicePixelRatio
```
mcnuttandrew: might be good to generalize this for node env testing, eg make a default at the top that is something like

```javascript
const DEFAULT_PIXEL_RATIO = window && window.devicePixelRatio || 1;
```

also, would you mind alphabetizing this list?

stronglink: Sounds like a good idea. I've also alphabetized the list.

src/plot/series/canvas-wrapper.js
stronglink: Yea, that's the cause. I've modified the README to point out `npm test` doesn't lint the code.

src/plot/xy-plot.js
```diff
@@ -56,6 +56,8 @@ const DEFAULT_MARGINS = {
   bottom: 40
 };
 
+const DEFAULT_PIXEL_RATIO = window && window.devicePixelRatio || 1;
```
mcnuttandrew: hmm it'd probably be better if this was a default prop in the canvas-wrapper component? sorry to flip around on this

stronglink: No problem! I've changed it.

mcnuttandrew:
👍👍👍

PR #607, opened 2017-09-14
Fix treemap sorting 

mrawdon:
The node size property does not exist  when sorting is performed in the current code.  From the [d3-hierarchy docs](https://github.com/d3/d3-hierarchy#node_sum) sum should be called first to set the value property.  Then sort can be called to sort the node on value.

I also used the reccomended sort functions from that doc. We may want to move sortFunction to a prop allowing a user to override these.

Adjusting the sorts, created treemaps that are more organized, flowing from largest to smallest

Now
image

vs
Before

image

### DIFF
--- a/src/treemap/index.js
+++ b/src/treemap/index.js
@@ -141,8 +141,8 @@ class Treemap extends React.Component {
           .size([innerWidth, innerHeight])
           .padding(padding);
       const structuredInput = hierarchy(data)
-        .sort((a, b) => a.size - b.size)
-        .sum(d => d.size);
+        .sum(d => d.size)
+        .sort((a, b) => a.value - b.value);
       return packingFunction(structuredInput).descendants();
     }
 
@@ -152,8 +152,8 @@ class Treemap extends React.Component {
       .size([innerWidth, innerHeight])
       .padding(padding);
     const structuredInput = hierarchy(data)
-      .sort((a, b) => a.size - b.size)
-      .sum(d => d.size);
+      .sum(d => d.size)
+      .sort((a, b) => b.value - a.value);
 
     return treemapingFunction(structuredInput).descendants();
 
--- a/src/treemap/index.js
+++ b/src/treemap/index.js
@@ -113,7 +113,7 @@ class Treemap extends React.Component {
    */
   _getNodesToRender() {
     const {innerWidth, innerHeight} = this.state;
-    const {data, mode, padding} = this.props;
+    const {data, mode, padding, sortFunction} = this.props;
     if (!data) {
       return [];
     }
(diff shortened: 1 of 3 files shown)


mrawdon:
I updated the treemap test I broke. I replaced the hard coded strings with functions to sanity check things.  I can move it back to a hard coded string if that is preferred.

src/treemap/index.js
```diff
@@ -141,8 +141,8 @@ class Treemap extends React.Component {
           .size([innerWidth, innerHeight])
           .padding(padding);
       const structuredInput = hierarchy(data)
-        .sort((a, b) => a.size - b.size)
```
mcnuttandrew: it's weird that we weren't getting an error if it didn't exist?

tests/components/treemap-tests.js
```diff
@@ -46,19 +46,30 @@ const TREEMAP_PROPS = {
...
+}
+
 // make sure that the components render at all
 testRenderWithProps(Treemap, TREEMAP_PROPS);
 
 test('Treemap: Basic rendering', t => {
   const $ = mount(<Treemap {...TREEMAP_PROPS}/>);
   t.equal($.find('.rv-treemap__leaf').length, 21, 'should find the right number of children');
-  const expectedText = 'EasingNeonateinterpolateISchedulableParallelPauseFunctionSequenceSequenceTransitionTransitionerTransitionEventSchedulerArrayInterpolatorColorInterpolatorDateInterpolatorInterpolatorMatrixInterpolatorNumberInterpolatorObjectInterpolatorPointInterpolatorRectangleInterpolator';
+  const expectedText = toString(TREEMAP_PROPS.data.children, 0);
```
mcnuttandrew: looks like this 0 is unused?

src/treemap/index.js
```diff
@@ -152,8 +152,8 @@ class Treemap extends React.Component {
       .size([innerWidth, innerHeight])
       .padding(padding);
     const structuredInput = hierarchy(data)
-      .sort((a, b) => a.size - b.size)
-      .sum(d => d.size);
+      .sum(d => d.size)
+      .sort((a, b) => b.height - a.height || b.value - a.value);
```
mcnuttandrew: parens please, eg `(b.height - a.height) || (b.value - a.value)`

tests/components/treemap-tests.js
mcnuttandrew: also i'd maybe advise against having these functions generate test results? It makes it harder to update the tests?

src/treemap/index.js
mrawdon: It would just end up with undefined-undefined=NaN which I think just leaves things in its original order

tests/components/treemap-tests.js
mrawdon: Yeah you're right about the 0, I'll just move those back to strings instead of the functions anyway.

mrawdon:
I think I've addressed your concerns here.  The original test keeps the same behavior, while allowing a user to override the sortFunction which I added a test for.

Issue #534, opened 2017-08-07
Moving away from SCSS


NameFILIP:
Are there any plans to move away from SCSS to avoid the `@import ...` styles step? Having SCSS as a dependency only to use the charting library is not ideal for projects which don't use SCSS.

Inline styles or `styled-components` would be a good alternative.

mcnuttandrew:
We've talked about this before, and maybe now is the time to completely switch over to inline styles as we've been in a mixed context (of some inline/some stylesheet) for a while. I think in order to complete this task the styling system would need to be added to the legends, refined in the treemap and radial chart.

All that is to say, i think this should be totally possible. I would push against use styled-components as we already have quite a bit of investment in inline styling, but I would hear other opinions.

jckr:
I'm personally happy with the current situation which doesn't mean I can't be convinced otherwise. 

Right now, react-vis components look and feel can be controlled in many ways:
- through the props which are exposed (ie color for a series),
- by passing a style prop - for added flexibility, and this is pretty close to having inline styles,
- through external style sheets / scss. 

So it's correct that today to have the default react-vis styles some kind of linking to the react-vis stylesheet must take place. that's one extra step and while in certain build systems it's relatively painless in others it's more of a hassle. 

However, if we were to move to inline styles, then users would no longer have the possibility to change look and feel via external style sheets. or it would be extra complicated and hacky. 

imo styling through style sheets, if not the only way to control look and feel of charts, is still a pretty valid approach. 

so that's where i stand on this

alee8:
How about considering CSS Level 3 to replace the features using SCSS?

mcnuttandrew:
@alee8 can you say more about that? the react-vis style sheet is already exported as a plain css file? I guess note of pedantry, that style sheet could be imported via html (thus preventing the need for adding scss compilation to a project that wouldn't have it by default)?

borisrorsvort:
Inline styles library are legions, what about people who are already constrained by one? 

For reference, Material-ui makes a good case for styling solution interop https://material-ui-1dab0.firebaseapp.com/guides/interoperability/

PR #608, opened 2017-09-14
Allow user to alter the input field class name on the Searchable Legend

mrawdon:
We have a default input style in our app.  By putting an unconfigurable class name on the input I can't easily use both the default css from react-vis and the default input style from our app.

### DIFF
--- a/showcase/legends/searchable-discrete-color.js
+++ b/showcase/legends/searchable-discrete-color.js
@@ -22,6 +22,8 @@ import React from 'react';
 
 import SearchableDiscreteColorLegend from 'legends/searchable-discrete-color-legend';
 
+import ShowcaseButton from '../showcase-components/showcase-button';
+
 export default class Example extends React.Component {
 
   constructor(props) {
@@ -36,7 +38,8 @@ export default class Example extends React.Component {
         {title: 'Limes', color: '#cf3'},
         {title: 'Potatoes', color: '#766'}
       ],
-      searchText: ''
+      searchText: '',
+      useDefaultInputClass: true
     };
     this._clickHandler = this._clickHandler.bind(this);
     this._searchChangeHandler = this._searchChangeHandler.bind(this);
@@ -45,7 +48,9 @@ export default class Example extends React.Component {
   _clickHandler(item) {
     const {items} = this.state;
     item.disabled = !item.disabled;
-    this.setState({items});
+    this.setState({
+      items
+    });
   }
 
   _searchChangeHandler(searchText) {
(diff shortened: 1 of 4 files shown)


src/legends/searchable-discrete-color-legend.js
```diff
@@ -64,7 +67,7 @@ function SearchableDiscreteColorLegend(props) {
         <input
           type="search"
           placeholder={searchPlaceholder}
-          className="rv-search-wrapper__form__input"
+          className={inputClassName}
```
mcnuttandrew: i think better solution here would be to still keep that default class name available and always accept what the user gives, eg

```javascript
className={`rv-search-wrapper__form__input ${inputClassName}`}
```

and then the defaultProps would just be `inputClassName: ''`. I believe that this solution would be more in line with the way that we do class naming throughout the rest of the library.

showcase/legends/searchable-discrete-color.js
```diff
@@ -45,24 +48,35 @@ export default class Example extends React.Component {
...
     return (
-      <SearchableDiscreteColorLegend
-        height={200}
-        width={300}
-        onSearchChange={this._searchChangeHandler}
-        searchText={searchText}
-        onItemClick={this._clickHandler}
-        items={items}
-      />
+      <div>
```
mcnuttandrew: I don't know if you need to go to the trouble of adding class toggling here, i think a more effective testing method would be to use an example inputClassName here, and then write a unit test against it.

mrawdon: Ok that makes sense.  I wasn't sure if this was a property that needed to be demonstrated in the show case or not

src/legends/searchable-discrete-color-legend.js
mrawdon: The issue with that approach is it still results in the styles from rv-search-wrapper__form__input being applied.  In my approach I am easily able to suppress that class and use my default input style.

An alternative would be to split the styles from rv-search-wrapper__form__input  into its own css file, making it easier to selectively include that styling.

showcase/legends/searchable-discrete-color.js
mcnuttandrew: generally, i only add buttons for properties that are visually impacted

src/legends/searchable-discrete-color-legend.js
mcnuttandrew: it seems like it will be pretty easy to override when there a multiple class names. to wit, if you had something like 
```scss
.my-custom-class.rv-search-wrapper__form__input {
//style style style
}
```
that would easily supercede the default styles. I'm guess I'm hesitant to start enabling the complete over-writing of styles? It makes sense here, but few other places.


An additional technique that we could employ here, (that we also employ other places) is offering a style prop at the top SearchableDiscreteColorLegend, that looks like

```
style: {
input: {},
colorLegend: {}
}
```

mrawdon: With either of those techniques I'd be concerned about a new property I hadn't overriden on .rv-search-wrapper__form__input being added. Like if font-family the property was added to the css file in the future, then the search field would be using a different font from the rest of my app.

If css changes like that were considered breaking and only made on major releases, then I think your solution would work for me.

Apercu:
@mrawdon concerns regarding overriding are justified since I don't see myself writing properties I don't want to force using the css defaults.

I think the idea itself is good, but implementing it only on this form input doesn't make sense to me. But that's apparently not what @mcnuttandrew is thinking, so maybe you can expand why you are reluctant to implement full style overriding?

I was thinking of adding either a top level wrapper making using of `context` (yeah I know I know) or adding a `classPrefix` property on each of the components using a `className` that would result in things looking like `.yolo-search-wrapper__form__input`.

This way people wanting to override completely every component class could do it if they wanted to (although I think it might be quite some work), or do it only at the component level.

mrawdon:
I think a change like what @Apercu is suggesting, is more in line with the issue #534 since its a larger change to how react-vis styles components.

I think using @mcnuttandrew suggestion is probably the easiest and most in line with how things currently work in react-vis

```
className={`rv-search-wrapper__form__input ${inputClassName}`}
```

If that makes sense, I'll update my pull request accordingly

mrawdon:
I ended up using my own input field separate from the component here.  So I'm going to close this

PR #632, opened 2017-10-04
Treemap margin property not used

mrawdon:
The margin prop wasn't being used by the treemap.  Instead the const DEFAULT_MARGINS was being used.

This fixes that behavior allowing the user to set the margin prop


mcnuttandrew:
Doh, good catch

Issue #631, opened 2017-10-04
Sankey Link Tooltips


awjreynolds:
The sankey chart is useful however it would be good to be able to hover over a link and see a tooltip with some sort of display of the data. Being able to pass in a link tooltip function would be extremely useful.

mcnuttandrew:
Hey @awjreynolds, It should be relatively easy to add a call backs for link. I would encourage you to open a PR to that effect! In order to maintain consistency across the library you'd probably want to add three onLinkClick/onLinkMouseOver/onLinkMouseOut.

stu-harvey:
@mcnuttandrew I went ahead and added those event handlers because I need them for a similar purpose. Would appreciate a review on the PR

awjreynolds:
@mcnuttandrew I was trying to get Hints working with the Sankey but I'm unsure if I'm approaching this right. However I do not understand how the hint is not working giving that the underlying component is XYPlot.
`
const { hoveredValue } = this.state;

<Sankey
        nodes={this.state.data.nodes}
        links={this.state.data.links}
        width={1500}
        height={this.state.data.nodes.length * 35}
        onValueClick={this.showDataPoint}
        onValueMouseOver={this.showDataPoint}
        onValueMouseOut={this.showDataPoint}
        hasVoronoi={true}
      >
        {hoveredValue ? 
          <Hint value={hoveredValue}>
            <div style={{background: 'black'}}>
              <h3>Value of hint</h3>
              <p>{hoveredValue.value}</p>
            </div>
          </Hint> : null}
      </Sankey>`

mcnuttandrew:
@awjreynolds  I'm not sure, but i'm betting some of your issue is being cause by

        onValueMouseOver={this.showDataPoint}
        onValueMouseOut={this.showDataPoint}

If you wanna spin up a code pen i'd be happy to take a look

chlumper:
@awjreynolds I extended the Sankey diagram to allow rendering of children within the XYPlot in my PR #648. The PR also adds a showcase example for the usage of a Hint in combination with the Sankey diagram (It shows the Hint on hover of a link).

See: https://github.com/chlumper/react-vis/blob/10e4a4f0cb06e81990b0c06136964f790daf1c76/showcase/sankey/link-hint.js

I hope the PR will find its way into the next release and it helps solving your problem.

mcnuttandrew:
Now that #648 is merged in, i think this issue is now addressed. @awjreynolds has this addressed your issues?

[...]

PR #634, opened 2017-10-05
Add event handlers to Sankey links

stu-harvey:
This commit adds onClick, onMouseOver, and onMouseOut event handlers to
the link SVG paths on Sankey charts. I've also added a quick example to
the showcase.

Will be useful for #631

### DIFF
--- a/docs/sankey.md
+++ b/docs/sankey.md
@@ -146,6 +146,45 @@ This handler is triggered either when the users mouse leaves a node. Callback wh
 ```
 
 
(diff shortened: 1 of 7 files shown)


mcnuttandrew:
This is great! This is way more than I was expecting tbh! Would you mind writing a test against your new example, just so there are no regressions in the future?

mcnuttandrew:
i don't specifically! @Apercu or @jckr might be able to help

PR #635, opened 2017-10-05
Move react-test-renderer to devDependencies

chollier:
Moves react-test-renderer from `dependencies` to `devDependencies` since I doubt it's actually needed to use the library

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,8 +48,7 @@
     "deep-equal": "^1.0.1",
     "global": "^4.3.1",
     "prop-types": "^15.5.8",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -48,8 +48,7 @@
     "deep-equal": "^1.0.1",
     "global": "^4.3.1",
     "prop-types": "^15.5.8",
-    "react-motion": "^0.4.8",
```
mcnuttandrew: I'm pretty sure we need react motion to animation?

PR #637, opened 2017-10-11
Swap height and width in partition pivot treemaps

mrawdon:
The height and width need to be swapped when using a partition pivot.  Otherwise the treemap will flow out of the container.
Old behavior
image

Fixed behavior
image

### DIFF
--- a/src/treemap/index.js
+++ b/src/treemap/index.js
@@ -120,7 +120,7 @@ class Treemap extends React.Component {
 
     if ((mode === 'partition' || mode === 'partition-pivot')) {
       const partitionFunction = partition()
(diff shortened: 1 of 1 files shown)


mcnuttandrew:
Did you check the effect on the SVG mode too? Otherwise looks good

mrawdon:
The SVG one was already all set, I believe the PolygonSeries ends up getting scaled properly inside of its plot.

Existing
image

My branch
image

mcnuttandrew:
Dope! Wish we had a visual regressions testing system so that we could have tests on this stuff, but what're you gonna do.

PR #638, opened 2017-10-11
Added sort to partition treemap

mrawdon:
There was no sort for the partition treemap before, added that in to match the other modes.


mcnuttandrew:
Given that this could be a breaking change, would you mind writing a test?

mrawdon:
Just pushed a change to the tests.  The custom sort test now checks all treemap modes to confirm correctness.

Issue #639, opened 2017-10-11
Space left at the top of Partition Treemaps


mrawdon:
There is always space left at the top of the treemap of a partition treemap for the root node.

![image](https://user-images.githubusercontent.com/1678756/31464708-300d8b68-aea0-11e7-8c02-987eb6fd9086.png)

This is caused by skipping the root node in the following code.
https://github.com/uber/react-vis/blob/master/src/treemap/treemap-dom.js#L34
and
https://github.com/uber/react-vis/blob/master/src/treemap/treemap-svg.js#L95

The easiest solution here would just be to display the root node like for circle packing.
To keep this from altering existing functionality exposing a prop would work.

Another option is to hide the root when its just a container for children.  That is discussed in this stackoverflow question, https://stackoverflow.com/questions/28700591/is-it-possible-to-have-multiple-parent-root-nodes-in-a-d3-icicle-plot

I can put a PR together for either option, just wanted to discuss first

mrawdon:
@mcnuttandrew any thoughts on this one?  Thanks

mcnuttandrew:
I guess i'd be push for displaying the root node by default, and adding a prop to turn it off (ala the opposite of how it is on sunburst). I think such a change would have a minimal impact on the rest of the code base.

mrawdon:
Makes sense, I'll put a PR together for this soon

PR #641, opened 2017-10-13
Added treemap hideRootNode property

mrawdon:
Added a  property to treemap hideRootNode with a default of false as discussed in #639 with @mcnuttandrew 

This did cause some changes in the number of nodes for existing treemaps.  I updated the tests and added a new test for this property.

This did result in visual changes for some treemaps.  

Partitions now show the root node(the desired result)
New
image

vs Old

image

For squarify there is a slight change in the space between the node padding(where the root node becomes visible)
New
image

vs old

image

### DIFF
--- a/tests/components/treemap-tests.js
+++ b/tests/components/treemap-tests.js
@@ -99,7 +99,56 @@ test('Treemap: Custom Sorting', t => {
 
 test('Treemap: Empty treemap', t => {
   const $ = mount(<Treemap {...{...TREEMAP_PROPS, data: {}}}/>);
-  //1 is the empty root node
+  // 1 is the empty root node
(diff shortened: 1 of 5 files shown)


tests/components/treemap-tests.js
```diff
@@ -123,7 +171,7 @@ test('Treemap: SimpleTreemap', t => {
   ].forEach(mode => {
     const selector = mode === 'circlePack' ? '.rv-treemap__leaf circle' : '.rv-treemap__leaf';
     // circle pack includes the root node, while the other modes do not
-    const numberOfElements = mode === 'circlePack' ? 252 : 251;
+    const numberOfElements = 252;
```
mcnuttandrew: that's way way better

PR #646, opened 2017-10-24
merge latest version from react-vis into forked repo

chlumper:


chlumper:
added PR by accident, sorry!

Issue #651, opened 2017-11-02
Axis Title Style Bug


btthomas:
When I try to style an axis label. It does not apply the styles:

<img width="318" alt="screen shot 2017-11-02 at 4 52 51 pm" src="https://user-images.githubusercontent.com/9452950/32350729-8e403354-bff1-11e7-8b98-3f1fb3496293.png">

<img width="599" alt="screen shot 2017-11-02 at 4 52 58 pm" src="https://user-images.githubusercontent.com/9452950/32350748-9ff8f090-bff1-11e7-977a-a27b69633b02.png">

PR: #652

PR #652, opened 2017-11-02
Fix axis title style bug

btthomas:
issue: #651 

I am now applying the style to the text inside of the axis title.

I changed the showcase so that the custom axis chart shows that the user can change the font. 

I added a test to test if the style is being applied to the correct title.

screen shot 2017-11-02 at 5 17 57 pm


jckr:
Thanks Blake. lgtm

btthomas:
@jckr , What is the npm release schedule?

PR #657, opened 2017-11-16
Improve performance of MarkSeries using shouldComponentUpdate

jonsadka:
## Prompt
Improve rerendering performance of `MarkSeries` component by using `shouldComponentUpdate`

## Problem
The `MarkSeries` component receives several props of type function (`onNearestX`, `onNearestXY`, etc.). If these functions are called, the `MarkSeries` component will automatically rerender all datapoints by mapping `this._renderCircle` on them, regardless if the change is necessary.

## Proposal
Only rerender when props affecting the view are changed. 

### Before
kapture 2017-11-13 at 11 29 13-min

### After
kapture 2017-11-13 at 10 48 02-min

### DIFF
--- a/showcase/plot/scatterplot-large.js
+++ b/showcase/plot/scatterplot-large.js
@@ -1,0 +1,84 @@
+// Copyright (c) 2016 - 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import React from 'react';
+
+import {
+  Hint,
+  HorizontalGridLines,
+  makeWidthFlexible,
+  MarkSeries,
+  VerticalGridLines,
+  XYPlot,
+  YAxis
+} from 'index';
+
+const FlexibleXYPlot = makeWidthFlexible(XYPlot);
+
+export default class Example extends React.Component {
+  componentWillMount() {
+    this.setState({
+      data: randomDatapoints(randomInt(700)),
+      nextPointCount: randomInt(700)
+    });
+  }
+
+  render() {
+    const {data, nextPointCount, value} = this.state || {};
+    return (
+      <div style={{width: '100%'}}>
+        Displaying {data.length} points per chart <button onClick={() => this.setState({
+          data: randomDatapoints(nextPointCount),
+          nextPointCount: randomInt(700)
+        })}>
+          Generate {nextPointCount} New Points
+        </button>
+        {new Array(4).fill('_').map((d, i) =>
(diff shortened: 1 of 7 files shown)


src/plot/series/mark-series.js
```diff
@@ -32,6 +33,35 @@ const predefinedClassName = 'rv-xy-plot__series rv-xy-plot__series--mark';
...
+      [this.props.colorDomain, nextProps.colorDomain],
+      [this.props.colorRange, nextProps.colorRange],
+      [this.props.data, nextProps.data],
+      [this.props.fillDomain, nextProps.fillDomain],
+      [this.props.fillRange, nextProps.fillRange],
+      [this.props.innerHeight, nextProps.innerHeight],
+      [this.props.innerWidth, nextProps.innerWidth],
+      [this.props.marginLeft, nextProps.marginLeft],
+      [this.props.marginTop, nextProps.marginTop],
+      [this.props.nullAccessor, nextProps.nullAccessor],
```
jckr: what is this?

jonsadka: javascripts

jckr: specifically - nullAccessor? where does that come from?

jonsadka: Oh. That's an optional prop which returns a boolean but for `MarkSeries` its defined at the very bottom of the file 
```javascript
MarkSeries.defaultProps = {
  nullAccessor: () => true
};
```
and from the docs
```markdown
#### nullAccessor (optional)
Type: `function`  
Default: `null`  
A function that will be invoked for each data element that will return a boolean that specifies if the data point should be drawn or not. For more information see [the D3 documentation](https://github.com/d3/d3-shape#line_defined).
```

showcase/plot/scatterplot-large.js
```diff
@@ -0,0 +1,84 @@
...
+  XYPlot,
+  YAxis
+} from 'index';
+
+const FlexibleXYPlot = makeWidthFlexible(XYPlot);
+
+export default class Example extends React.Component {
+  componentWillMount() {
+    this.setState({
+      data: randomDatapoints(randomInt(700)),
```
ericsoco: Why are you randomizing the number of points to render? Why not let that be a constant (but still randomize each datapoint)? Also, seems like a "scatterplot-large" would be significantly more than 700 points...I would expect a viz framework to be able to gracefully handle at least a couple thousand.

src/plot/series/mark-series.js
```diff
@@ -32,6 +33,44 @@ const predefinedClassName = 'rv-xy-plot__series rv-xy-plot__series--mark';
 const DEFAULT_STROKE_WIDTH = 1;
 
 class MarkSeries extends AbstractSeries {
+  shouldComponentUpdate(nextProps, nextState) {
+    const isMissingDataProp = !this.props.data || !nextProps.data;
+    const isMissingNullAccessorProp = !this.props.nullAccessor || !nextProps.nullAccessor;
+    return [
```
ericsoco: There's no reason to recreate this array every time this func is called. Make a const array outside the class declaration with just the string prop names, and loop over that and pull values out of `this.props` and `nextProps` for each string name. More legible and a slight optimization.

src/plot/series/mark-series.js
```diff
@@ -32,6 +33,44 @@ const predefinedClassName = 'rv-xy-plot__series rv-xy-plot__series--mark';
...
+      [this.props.strokeDomain, nextProps.strokeDomain],
+      [this.props.strokeRange, nextProps.strokeRange],
+      [this.props.strokeWidth, nextProps.strokeWidth],
+      [this.props.style, nextProps.style],
+      [this.props.xDomain, nextProps.xDomain],
+      [this.props.xRange, nextProps.xRange],
+      [this.props.yDomain, nextProps.yDomain],
+      [this.props.yRange, nextProps.yRange]
+    ].some(([a, b]) => !shallowequal(a, b)) || (
+      isMissingDataProp ?
```
ericsoco: I'd pull this out into its own function for two reasons:
1. legibility: you can name + comment the function to describe this comparison you're doing
2. reusability: seems like a util that could be useful elsewhere.

Also, this diff should definitely include unit tests for this logic.

ericsoco:
Is this optimization useful in other series?

And can you provide the benchmarks you used to validate it? (I know you did some testing, that should be in this PR for future reference.)

I'm generally a bit wary of `shouldComponentUpdate` optimizations that are not super simple, they often seem to backfire on you...and React DOM diffing often is all the optimization you need given that rendering is usually the most expensive part.

mcnuttandrew:
I would strongly caution against any work that reproduces every single one of the props that a series might take in a single place. This effectively means that anything a new prop is introduced it will need to be duplicated into this one specific series. 

I agree with @ericsoco's sentiment that this could be moved into abstract series so that components are not updated when their listeners are called. Given that we name listeners in a standard way (onNearestX onMouseOver onBlanketyBlank) it could probably be inferred which props need to be checked from a simple heuristic regex. Adding this to AbstractSeries would certainly be a big win and would probably lighten up the load of some of the other more complex series, such as contour etc.

Also plus 1 for some bench marks, proving you're right is good.

Issue #650, opened 2017-11-02
Stylesheet inclusion check fails on Firefox in dev environment


stropitek:
I get a security error in Firefox when using react-vis.
![screen shot 2017-11-02 at 2 36 58 pm](https://user-images.githubusercontent.com/4118690/32328778-5551232a-bfdb-11e7-847e-49b6bff87778.png)

The error occurs when I develop using webpack-dev-server. Firefox throws a security error because react-vis (loaded from localhost:8080) is trying to access a stylesheet that was loaded from a different origin (I'm including semantic-ui css from a cdn). The line that throws is [that one](https://github.com/uber/react-vis/blob/3be2bec66d14ce3079c9fc0d181d3542234ee352/src/utils/react-utils.js#L88). The error does not exist in the production bundle because that check is removed.

I'm not sure what the fix should be. Try-catching that line would not solve the underlying issue but would at least prevent the page to crash in those circumstances.

mcnuttandrew:
Huh, that's annoying. I'm not quite sure what a good workaround would be, maybe a minimal solution would involve adding a prop to XYPlot like "skipStylesheetCheck" that would skip the stylesheet check? If you wanted to file a PR to that effect I'd gladly accept it.

tux4:
Hi,
I am facing this exact same issue. It is preventing me from working on `react-vis` in firefox. A quick way of reproducing this would be to try this:

```
document.styleSheets[0].cssRules
```

Which throws:

```
SecurityError: The operation is insecure.
```

@stropitek Have you managed to fix this? If not, I can make a PR, as @mcnuttandrew suggested.

stropitek:
@tux4 I haven't started to work on this. Feel free to do the PR

mcnuttandrew:
I'm going to close this as #658 has landed and appears to address this issue. Feel free to re-open if you feel otherwise.

PR #658, opened 2017-11-16
Fix firefox SecurityError in development

tux4:
This PR fixes the `SecurityError` that is thrown in development mode for Firefox. 

See: https://github.com/uber/react-vis/issues/650 for more context.

I tested it in Firefox 58 and Chrome 60.

### DIFF
--- a/src/utils/react-utils.js
+++ b/src/utils/react-utils.js
@@ -23,8 +23,7 @@ import React from 'react';
 const [major, minor] = React.version.split('.');
 const versionHigherThanThirteen = Number(minor) > 13 || Number(major) > 13;
 
-export const isReactDOMSupported = () =>
-  versionHigherThanThirteen;
+export const isReactDOMSupported = () => versionHigherThanThirteen;
 
 /**
  * Support React 0.13 and greater where refs are React components, not DOM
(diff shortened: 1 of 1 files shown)


src/utils/react-utils.js
```diff
@@ -85,7 +103,7 @@ export function checkIfStyleSheetIsImported() {
 
   const foundImportTag = [...new Array(document.styleSheets.length)].some((e, i) => {
     const styleSheet = document.styleSheets[i];
-    const CSSRulesList = styleSheet.rules || styleSheet.cssRules;
+    const CSSRulesList = getCSSRules(styleSheet);
```
balthazar: maybe you could do

```js
const CSSRulesList = getCSSRules(styleSheet) || [];
```

and then `[...new Array(CSSRulesList.length)]`

tux4: I was wondering along similar line. Why can't I do:

```
  const foundImportTag = [...document.styleSheets].some(styleSheet => {
    const CSSRulesList = getCSSRules(styleSheet);
    return [...CSSRulesList].some(selector => {
      return selector.selectorText === MAGIC_CSS_RULE;
    });
  });

```

And `getCSSRules(styleSheet)` return `[ ]` when unsafe.

tux4: @balthazar I updated the PR with some changes to how we are iterating over the css rules and selectors. Any thoughts?

balthazar: That does look better but it looks like you remove the `styleSheet.rules || styleSheet.cssRules` logic, so I'm not sure if that would work, did you tried locally by importing and not importing? Also I'm not sure the stylesheets can be iterated like that and I guess that's why @mcnuttandrew did this using the index but I'm honestly not sure. If he could shim-in that would be great 😄

tux4: > did you tried locally by importing and not importing? 

I did try importing and not importing the magic rule. Things work correctly, and the error message is not thrown in FF.

mcnuttandrew:
I used an index for iteration across the stylesheets because i initially tried to iterate across them directly, eg document.styleSheets.forEach, which did not work, as they live in a special array type. However it appears that doing a [...X] approach forces them to behave like a normal array, so seems fine.

Tested on lastest chrome + latest Safari, works good +1. I stamp this, but will wait for @apercu to merge, as he has been reading on this as well.

balthazar:
I see that makes sense 😃 
Looking good then

ps: @mcnuttandrew I changed my nickname!

PR #660, opened 2017-11-17
fix for refs being null with JEST

jonattfin:
I'm using snapshot testing with JEST and because refs are not being recognized by JEST => containerElement is undefined, and I can't use the components in snapshot testing because on line 127, it will try to access offsetHeight & offsetWidth on an undefined element. So this was my fix for this problem.

If i do it like this, snapshot works!


jonattfin:
i will commit another

Issue #663, opened 2017-11-21
XYPlot fails on IE11 - endsWith is not supported


housseindjirdeh:
First and foremost, thank you all for open sourcing such an awesome library :)

One thing I just noticed - using an XYPlot fails on IE11 due to the [`_getOptionalScaleProps` method](https://github.com/uber/react-vis/blob/9b7c2007b86090a634df08f6514e791a9fa228ed/src/plot/xy-plot.js#L291):

```
  _getOptionalScaleProps(props) {
    return Object.keys(props)
      .filter(prop => OPTIONAL_SCALE_PROPS.map(option => prop.endsWith(option)).every(x => x))
      .reduce((obj, key) => {
        obj[key] = props[key];
        return obj;
      }, {});
  }
```

`prop.endsWith` fails since `String.endsWith` [isn't supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith#Browser_compatibility) in IE.

balthazar:
Should be fixed since Andrew's fix. Also, not every library is going refrain using newest methods for browser compatibility, especially IE. You should add polyfills on your side if that's the case 🙂

balthazar:
Published in `1.8.0`

housseindjirdeh:
Yep you're totally right - will most likely have to add polyfills sooner or later :O

Thanks a million however 🎉 🎉 🎉

PR #664, opened 2017-11-24
Refactor getOptionalScaleProps, add tests, IE11 support

mcnuttandrew:
This PR addresses #663 by refactoring getOptioanlScaleProps into a little less cutting edge state, and pulling it into a utils file. In it's current state this function seems kinda unnesscary, like it seems like a lot of overkill just to test for the presence of a padding prop, but that's not the point of this PR. The point of this PR is to unbreak this function on IE11.


balthazar:
IE11 support weew we back in the game!

PR #669, opened 2017-11-29
Warn for nullAccessor renaming

balthazar:


balthazar:
Yeah, just warn, you think I should also assign `getNull` to the old prop? Seems meh

mcnuttandrew:
I mean without reassigning it's still a breaking change, but I also don't feel that strongly?

mcnuttandrew:
<3

Issue #677, opened 2017-12-03
tickFormat doesn't render arbitrary svg react elements


krabbypattified:
According to [tickFormat](https://uber.github.io/react-vis/documentation/api-reference/axes#tickvalues-optional-), I should be able to return arbitrary svg from the formatter. However, since all of this gets wrapped in a `text` element, it won't render unless it's [permitted](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text) by the `text` element.

For instance, a `circle` won't render, but a `tspan` will.

My data visualization requires image ticks (see below), but that is not possible under the current API.

I'd like to propose a simple change: If the `tickFormat` function returns a non-`text` compliant React element, don't wrap it in a `text` element.

Thoughts?

<img width="649" alt="screen shot 2017-12-03 at 6 17 38 pm" src="https://user-images.githubusercontent.com/13423815/33531066-42f165b2-d856-11e7-9604-b91a3dc66f6c.png">

krabbypattified:
The issue is [here](https://github.com/uber/react-vis/blob/da2308d203a912b373e04d30a97ef78811500d09/src/plot/axis/axis-ticks.js#L177-L197)

krabbypattified:
const ticks = values.map((v, i) => {
      const pos = scale(v);
      const text = tickFormatFn(v, i);


      return (
        <g key={i} {...translateFn(pos, 0)}
        className="rv-xy-plot__axis__tick"
        style={style}>
          <line {...pathProps}
            className="rv-xy-plot__axis__tick__line"
            style={{...style, ...style.line}}
          />
          <text {...textProps}
            className="rv-xy-plot__axis__tick__text"
            style={{...style, ...style.text}}
          >
            {text}
          </text>
        </g>
      );
    });

PR #678, opened 2017-12-03
Allows for arbitrary svg to be returned from `tickFormat` function. #677

krabbypattified:
Right now, `tickFormat` won't render a `circle` (for instance). See issue #677.

### DIFF
--- a/src/plot/axis/axis-ticks.js
+++ b/src/plot/axis/axis-ticks.js
@@ -177,6 +177,15 @@ class AxisTicks extends React.Component {
     const ticks = values.map((v, i) => {
       const pos = scale(v);
       const text = tickFormatFn(v, i);
+      
+      const tick = React.isValidElement(text) && !['tspan','a'].includes(text.type)
+      ? text
+      : <text {...textProps}
+          className="rv-xy-plot__axis__tick__text"
+          style={{...style, ...style.text}}
+        >
+          {text}
+        </text>
 
       return (
         <g key={i} {...translateFn(pos, 0)}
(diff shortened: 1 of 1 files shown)


src/plot/axis/axis-ticks.js
```diff
@@ -177,6 +177,15 @@ class AxisTicks extends React.Component {
     const ticks = values.map((v, i) => {
       const pos = scale(v);
       const text = tickFormatFn(v, i);
+      
+      const tick = React.isValidElement(text) && !['tspan','a'].includes(text.type)
```
mcnuttandrew: You've got quite a number of lint errors coming off this chunk of code. I think this could be cleaned up by making a boolean variable like,

```
const isNotText = React.isValidElement(text) && !['tspan','a'].includes(text.type);
const tick = isNotText ? text : (
  <text {...textProps}
    className="rv-xy-plot__axis__tick__text"
    style={{...style, ...style.text}}
    >
    {text}
  </text>)
```

mcnuttandrew (commented):
Hey this is an interesting idea! I like what you've got so far, but there are a couple more steps still to go.

This functionality needs to be documented in the documentation (probably the axis.md specifically) and there should be some tests so we can verify the functionality in an automated way. Typically that way we write tests in the library is by making an example in the showcase (i'd probably make one in the axes section) that highlights new functionality, and then write a test against that example. Let me know if you have any questions.

src/plot/axis/axis-ticks.js
krabbypattified: Thanks! I'm not really experienced with pull requests, so bear with me :).
I'll update the PR soon
Or, feel free to update it yourself.

mcnuttandrew: Take your time. It's important to learn how to be comfortable doing PRs +1 for learning

mcnuttandrew:
This PR seems to have stalled, I'm going to close it, but if you want to work on the issue some more feel free to re-open

PR #681, opened 2017-12-05
Modified getStackData function to also work with partial data

sradevski:
I just created a PR so we can further discuss the code. I am not sure if I misunderstand something, but the tests fail while the changes render the example stacked charts nicely. It also works well with a bunch of data on my project (gives the result I have shown in the Issue). 

I am not sure if there is a difference between VerticalBarSeries and HorizontalBarSeries regarding the stackBy value, but from my understanding setting stackBy='y' on HorizontalBarSeries doesn't make much sense, thus the following test code will not work correctly:
```
 let children = [
    (<HorizontalBarSeries
      data={yData[0]}
    />),
    (<HorizontalBarSeries
      data={yData[1]}/>),
    (<div> i think i will by that lamp </div>)
  ];
  let results = getStackedData(children, 'y');
  let expectedResults = [[
    {x: 10, y: 2},
    {x: 5, y: 4},
    {x: 15, y: 5}
  ], [
    {x: 12, y: 4, y0: 2},
    {x: 2, y: 8, y0: 4},
    {x: 11, y: 10, y0: 5}
  ], undefined];
```
It is also weird how the HorizontalBarSeries is imported.

```
import HorizontalBarSeries from 'plot/series/vertical-rect-series';
```

As nothing is rendered and basically HorizontalBarSeries is used just as a plain object, we can ignore that. Still, to me it seems that the tests are wrong, and the above test should instead be: 

```
   let children = [
    (<HorizontalBarSeries
      data={xData[0]}
    />),
    (<HorizontalBarSeries
      data={xData[1]}/>),
    (<div> i think i will by that lamp </div>)
  ];
  let results = getStackedData(children, 'y');
  let expectedResults = [[
   {x: 2, y: 10},
   {x: 4, y: 5},
   {x: 5, y: 15}
  ], [
   {x: 2, y: 22, y0: 10},
   {x: 4, y: 7, y0: 5},
   {x: 5, y: 26, y0: 15}
  ], undefined];
```

This applies to all the tests for getStackedData. It is possible that I haven't understood the use case well, so I will be waiting for any of the maintainers' discussion.

### DIFF
--- a/src/utils/series-utils.js
+++ b/src/utils/series-utils.js
@@ -130,7 +130,7 @@ export function getStackedData(children, attr) {
     const baseAttr = attr === 'y' ? 'x' : 'y';
 
     accumulator.push(preppedData.map((d, dIndex) => {
-      if (!(cluster in latestAttrPositions)) {
+      if (!latestAttrPositions[cluster]) {
         latestAttrPositions[cluster] = {};
       }
 
--- a/src/utils/series-utils.js
+++ b/src/utils/series-utils.js
@@ -108,10 +108,13 @@ function prepareData(data) {
  * @returns {Array} New array of children for the series.
  */
 export function getStackedData(children, attr) {
+  // It stores the last segment position added to each bar, separated by cluster.
+  const latestAttrPositions = {};
+
   return children.reduce((accumulator, series, seriesIndex) => {
     // Skip the children that are not series (e.g. don't have any data).
     if (!series) {
-      accumulator.result.push(null);
+      accumulator.push(null);
       return accumulator;
     }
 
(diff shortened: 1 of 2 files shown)


src/utils/series-utils.js
```diff
@@ -108,44 +108,60 @@ function prepareData(data) {
...
-      accumulator.result.push(preppedData);
+      accumulator.push(preppedData);
       return accumulator;
     }
 
     const attr0 = `${attr}0`;
+    const baseAttr = attr === 'y' ? 'x' : 'y';
+
+    accumulator.push(preppedData.map((d, dIndex) => {
+      if (!(cluster in latestAttrPositions)) {
```
mcnuttandrew: that's a pretty python-y way to say that, maybe better to say !latestAttrPositions[cluster]

sradevski: Thanks for the response. I just wanted to confirm that is not me misunderstanding something before putting in more effort in it. I will add some tests and fix some of the issues these days and will commit again.

sradevski:
After looking at the tests I realized that they were in fact wrong based on the definition of what **stackedBy** means as written in the documentation. More specifically, 

**stackBy (optional)
Type: string
Stack the chart by the given attribute. If the attribute is y, the chart is stacked vertically; if the attribute is x then it's stacked horizontally.**


Let's take an example from before and analyze it.
```
 const yData = [
    [
      {y: 2, x: 10},
      {y: 4, x: 5},
      {y: 5, x: 15}
    ],
    [
      {y: 2, x: 12},
      {y: 4, x: 2},
      {y: 5, x: 11}
    ]
  ];

  let children = [
    (<HorizontalBarSeries
      data={yData[0]}
    />),
    (<HorizontalBarSeries
      data={yData[1]}/>),
    (<div> i think i will by that lamp </div>)
  ];
  let results = getStackedData(children, 'y');
  let expectedResults = [[
    {x: 10, y: 2},
    {x: 5, y: 4},
    {x: 15, y: 5}
  ], [
    {x: 12, y: 4, y0: 2},
    {x: 2, y: 8, y0: 4},
    {x: 11, y: 10, y0: 5}
  ], undefined];
  t.deepEqual(results, expectedResults, 'should find the correct results for stacking by y');

```
Stacking by y-axis means the segments are stacked vertically. Since all of the **x** values are different, there should be no stacking happening. What the expected result should be is 
```
let expectedResults = [[
    {x: 10, y: 2},
    {x: 5, y: 4},
    {x: 15, y: 5}
  ], [
    {x: 12, y: 2},
    {x: 2, y: 4},
    {x: 11, y: 5}
  ], undefined];
```

Although I know it is not the best of ideas to change a "working" tests, since the tests were wrong in the first place they didn't test anything, at least not correctly.

sradevski:
@mcnuttandrew I am sorry for tagging you, but I would really appreciate it if we can close this ASAP. If there are any additional issues or worries, please let me know and I will resolve them immediately. Cheers!

mcnuttandrew:
Sorry for the long delay, ~ holidays ~. I think this looks good, it's a hard function to review because the various implementations of it have, as you point out, been kind of messy. But, i think the state you have brought the whole thing into is for the best. 

Thanks for doing this!

mcnuttandrew:
hey @balthazar would you mind cutting a patch?

balthazar:
Sure thing

sradevski:
Thanks!

PR #685, opened 2017-12-11
Add classname and styles to Voronoi cells

### LINKED ISSUE #684: Changing mouse cursor while using Voronoi
I've made an interactive histogram (using `VerticalRectSeries` and `Voronoi`), and have attached `onHover` and `onClick` handlers to the Voronoi. This is how it currently looks:
without

As can be seen from the gif above, the cursor is the default. I'd like to change it to the `pointer` cursor, when hovering over buckets with data, like:
with

Normally, this is trivial to achieve with CSS but since the Voronoi is positioned on top of the other svg elements, I can't just add `cursor:pointer;` to them.

I've created a small example but for some reason it doesn't run on Codepen (I've experienced this before with react-vis. Not sure if I'm holding it wrong, or there is a "trick"):
https://codepen.io/sqren/pen/vWqPBK?editors=0010

sqren:
Closes #684

This change makes it possible to conditionally add classnames and styles to Voronoi cell, depending on the data points.

This PR also adds initial docs for the Voronoi diagram.

### DIFF
--- a/src/plot/voronoi.js
+++ b/src/plot/voronoi.js
@@ -28,7 +28,7 @@ function Voronoi({
     <g className={`${className} rv-voronoi`} style={style}>
       {voronoiInstance.polygons(nodes).map((d, i) => (
         <path
-          className={`rv-voronoi__cell ${d.data.className || ''}`}
+          className={`rv-voronoi__cell ${d.data && d.data.className || ''}`}
           d={`M${d.join('L')}Z`}
           onClick={() => onClick(d.data)}
           onMouseUp={() => onMouseUp(d.data)}
@@ -39,7 +39,7 @@ function Voronoi({
           style={{
             pointerEvents: 'all',
             ...polygonStyle,
-            ...d.data.style
+            ...(d.data && d.data.style)
           }}
           key={i} />
       ))}
--- a/src/plot/voronoi.js
+++ b/src/plot/voronoi.js
@@ -38,7 +38,8 @@ function Voronoi({
           fill="none"
           style={{
             pointerEvents: 'all',
-            ...polygonStyle
+            ...polygonStyle,
+            ...d.data.style
           }}
           key={i} />
       ))}
--- a/src/plot/voronoi.js
+++ b/src/plot/voronoi.js
@@ -16,8 +16,7 @@ function Voronoi({
   polygonStyle,
   style,
   x,
-  y,
-  cellClassName
+  y
 }) {
   // Create a voronoi with each node center points
   const voronoiInstance = voronoi()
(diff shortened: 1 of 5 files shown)


src/plot/voronoi.js
```diff
@@ -28,7 +29,7 @@ function Voronoi({
     <g className={`${className} rv-voronoi`} style={style}>
       {voronoiInstance.polygons(nodes).map((d, i) => (
         <path
-          className="rv-voronoi__cell"
+          className={`rv-voronoi__cell ${cellClassName(d.data)}`}
```
sqren: I'm not a huge fan of the name `cellClassName` and perhaps it should also allow strings (not just functions). But I didn't want to make it more complex.

src/plot/voronoi.js
```diff
@@ -56,7 +57,8 @@ Voronoi.defaultProps = {
   onMouseDown: NOOP,
   onMouseUp: NOOP,
   x: d => d.x,
-  y: d => d.y
+  y: d => d.y,
```
mcnuttandrew: might be better to have the default be something like:

```
cellClassName: d => d.className || ''
```

src/plot/voronoi.js
```diff
@@ -71,7 +73,8 @@ Voronoi.propTypes = {
   onMouseUp: PropTypes.func,
   onHover: PropTypes.func,
   x: PropTypes.func,
-  y: PropTypes.func
+  y: PropTypes.func,
+  cellClassName: PropTypes.func
```
mcnuttandrew: alphabetize here and above

src/plot/voronoi.js
sqren: Where would `d.className` come from? I can see that `d` is a polygon node from the voronoi instance but not clear on how classnames are attached to those.

sqren: I see now that it comes from `d.data.className`, which is passed to the Voronoi as `nodes`.That makes me think, that `cellClassName` is unnecessary. I'll update this PR.

sqren:
@mcnuttandrew I've updated the code and added a test for the Voronoi diagram. 
Starting from scratch I found it a little hard to add docs (not my strong suit). But I added a base, that will hopefully encourage others to improve upon it. I also wasn't clear on how to test the docs. `npm start` only gives me the show-case items. So there might be some glitches that I didn't catch.

Let me know what you think.

src/plot/voronoi.js
```diff
@@ -28,7 +28,7 @@ function Voronoi({
     <g className={`${className} rv-voronoi`} style={style}>
       {voronoiInstance.polygons(nodes).map((d, i) => (
         <path
-          className="rv-voronoi__cell"
+          className={`rv-voronoi__cell ${d.data.className || ''}`}
```
mcnuttandrew: just for paranoia maybe do d.data && d.data.className

sqren:
@mcnuttandrew Good to hear :) Commited the last fixes.

sqren:
@mcnuttandrew Any idea when this will be published on npm? Thanks!

mcnuttandrew:
@balthazar ?

balthazar:
`1.8.1`

Issue #613, opened 2017-09-18
Restrict stackBy when multiple series are present


LoicUV:
I have a plot with 2 stacked vertical bar series and a line series. I'd like to be able to use the `stackBy` prop on the `y` axis only for the two bar series and not the line series, but since it's a prop of the plot it gets applied to the line series too. Is there a way to achieve this ?
See [this codepen](https://codepen.io/anon/pen/gGPOza)

mcnuttandrew:
Hey @LoicUV,

I don't think there is a way to do this right now, however I think there easily could be! I imagine a prop added to each series called stack, that is a Boolean which is false by default. The behavior of the stacking algorithm would be such that, 

- if stack was present on any series it would only stack the series with the prop selected
- if stack is not present on any series it will try to stack all the series

I'm not available to do this work for a while, so I would encourage a PR!

craigbeck:
+1 for this issue. I have something similar where I'm using a `LineSeries` to draw a cursor along the Y axis when mousing over the graph, but with `stacked` rendering of graph this line is also stacked.

line: 
![screen shot 2018-03-08 at 11 21 03 am](https://user-images.githubusercontent.com/1620605/37171463-d5a3d5c8-22c2-11e8-885f-a06ee62940fb.png)

stacked line showing hint cursor also being stacked:
![screen shot 2018-03-08 at 11 20 31 am](https://user-images.githubusercontent.com/1620605/37171482-e482df58-22c2-11e8-8049-af524eca6247.png)

LoicUV:
@craigbeck It has been implemented, however it looks like [my changes on the documentation](https://github.com/uber/react-vis/pull/688/commits/71b972c8d78b635dfadce4438ea8972b591a301a) is missing in master. Can you look into this @mcnuttandrew ?

craigbeck:
excellent thanks @LoicUV! updated to latest and it works great

PR #688, opened 2017-12-13
Series stack opt in

LoicUV:
See #613


src/utils/series-utils.js
```diff
@@ -108,17 +108,25 @@ function prepareData(data) {
  * @returns {Array} New array of children for the series.
  */
 export function getStackedData(children, attr) {
+  let areSomeSeriesStacked = false;
```
mcnuttandrew: It would be better to execute this as using some, eg

```
const someSeriesStacked = children.some(series => series && series.props.stack);
```

src/plot/series/abstract-series.js
```diff
@@ -51,7 +51,8 @@ const propTypes = {
   onNearestX: PropTypes.func,
   onNearestXY: PropTypes.func,
   style: PropTypes.object,
-  animation: AnimationPropType
+  animation: AnimationPropType,
+  stack: PropTypes.bool
 };
 
 const defaultProps = {
```
mcnuttandrew: probably better to set the default stack prop on abstract-series's defaultProps rather than on every other series

showcase/plot/mixed-stacked-chart.js
```diff
@@ -0,0 +1,82 @@
...
+  XAxis,
+  YAxis,
+  VerticalGridLines,
+  HorizontalGridLines,
+  VerticalBarSeries,
+  VerticalBarSeriesCanvas,
+  LineSeries
+} from 'index';
+
+export default class Example extends React.Component {
```
mcnuttandrew: This is a good example! I like that we are checking that stacking works in canvas and svg. Would you mind writing a test against this example so that we don't back slide?

src/utils/series-utils.js
```diff
@@ -108,17 +108,25 @@ function prepareData(data) {
  * @returns {Array} New array of children for the series.
  */
 export function getStackedData(children, attr) {
```
mcnuttandrew: it would be good to write a unit test against getStackedData directly

src/plot/series/abstract-series.js
LoicUV: That's what I did at first but I wasn't sure it would be inherited as I saw some series defining their own `defaultProps`. Will do !

tests/utils/series-utils-tests.js
```diff
@@ -100,7 +100,7 @@ test('series-utils #seriesClusterProps', t => {
   t.end();
 });
 
-test('series-utils #getStackedData', t => {
+test('series-utils #getStackedData', t => { // eslint-disable-line max-statements
```
LoicUV: Wrong commit message here, sorry about that. 
For the eslint comment, I hope it's ok to disable only that rule here...

mcnuttandrew: Very reasonable, probably would also be reasonable to set that for all of the test suite, but that doesn't need to happen here

tests/components/xy-plot-tests.js
```diff
@@ -165,3 +166,57 @@ test('testing flexible charts', t => {
   t.notEqual(v.height, '100px', 'flexible vis - height is not 100px');
   t.end();
 });
+
+test('Render two stacked bar series with a non-stacked line series chart', t => {
+  const wrapper = shallow(
```
mcnuttandrew: Probably be a little better to use your mixed-bar-chart directly instead of creating a new component

LoicUV:
👋

showcase/plot/mixed-stacked-chart.js
mcnuttandrew: your example exposes a kinda funky ambiguity 
ex

The type of component that you are varying correctly switches between canvas and svg, but the other component (the line) remains an svg node which is weird! react-vis is clearly behaving as it should, however it is not spelled out in the docs the why and way this behaviour should be! 

nice! good edge pushing.

mcnuttandrew:
This is great work, thanks for being so patient and diligent with it

LoicUV:
Thank you !

LoicUV:
@balthazar Sorry for the mention, do you think you can push a new version with this any time soon? Thanks 🙏

PR #696, opened 2017-12-15
Upgrade docs

mcnuttandrew:


mcnuttandrew:
This PR upgrades the docs, normally i'd just upgrade them directly but i'm pretty unfamiliar with ocular so i'd appreciate a second pairs of eyes

balthazar:
Looks like it created a sub directory `dist` inside `dist`, which I don't think is right. Did you managed to get the showcases with link to work?

balthazar:
Publishing with your fix

mcnuttandrew:
The showcases are kinda weird and I don't think they need to be withLinked, like all of the components within them already have withLinks so it seemed redundant?

balthazar:
Yeah don't need to

PR #698, opened 2017-12-15
Fix bug #647 - Legend Items are always "clickable"

### LINKED ISSUE #647: Legend Items are always "clickable"
https://github.com/uber/react-vis/blob/ec1ed9f9b9dc140c1ee1d53292e51f31f99161a1/src/legends/discrete-color-legend-item.js#L38

I think you've got a bug here.  `onClick` is a function, so it's always "truthy".  Thus this `clickable` class always gets added.

I suggest fixing this over here:
https://github.com/uber/react-vis/blob/ec1ed9f9b9dc140c1ee1d53292e51f31f99161a1/src/legends/discrete-color-legend.js#L50

By only conditionally passing in a function, otherwise path in something "falsy"

sandiiarov:
closes #647

### DIFF
--- a/src/legends/discrete-color-legend.js
+++ b/src/legends/discrete-color-legend.js
@@ -37,20 +37,19 @@ function DiscreteColorLegend({
   width
 }) {
   return (
-    <div
-      className={`rv-discrete-color-legend ${orientation} ${className}`}
-      style={{width, height}}>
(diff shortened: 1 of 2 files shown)


mcnuttandrew:
Nice work! I like these PRs your doing, they're great.

PR #726, opened 2018-02-21
Avoid process reference error

angus-c:
In browser render this was erroring: 

```
VM469 client-vendor.js:68616 Uncaught ReferenceError: process is not defined
    at checkIfStyleSheetIsImported (VM469 client-vendor.js:68616)
    at XYPlot.componentDidMount (VM469 client-vendor.js:65809)
    at commitLifeCycles (VM469 client-vendor.js:46266)
```


angus-c:
thanks for landing!

PR #728, opened 2018-02-22
[Documentation] Fix reference to CSS styles

CadiChris:
The scss files are located under `styles`, not `style`


cadichris:
This PR is totally wrong..the CSS file is indeed `style.css` so README is correct. Went a little too fast on this one.

Closing this.

PR #729, opened 2018-02-22
Added a 'Localization' section in 'Scales and data' documentation

LoicUV:
Describes how one would add d3 based localization for time series charts.


LoicUV:
You are right. I think I will also describe how to localize numbers (decimal, thousands, etc.. which relies on d3-format).

PR #732, opened 2018-03-01
Expose `className` prop on Crosshair component and update docs

jonsadka:
For apps that use inline styles, the cross does not get positioned correctly because the crosshair is positioned absolutely. Expose `className` on this component so that applications using inline styles can hook into this component

### DIFF
--- a/docs/crosshair.md
+++ b/docs/crosshair.md
@@ -15,22 +15,26 @@ In case if custom representation of crosshair is needed, the component is able t
 </Crosshair>
 ```
 
+#### className (optional)
+Type: `string`
+Provide an additional class name for the series.
+
 #### values
-Type: `Array<Object>`  
+Type: `Array<Object>`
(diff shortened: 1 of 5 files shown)


mcnuttandrew:
Looks good! Maybe add a test?

jonsadka:
@mcnuttandrew tests added

mcnuttandrew:
I don't know why coveralls would have failed like that, maybe re-run the build?

jonsadka:
Yeah weird. Rebuilding travis

jonsadka:
@mcnuttandrew do you know what I need to do to get a new coverage report? I tried rebuilding both builds in travis but no change

mcnuttandrew:
that's super weird, i don't know why that would be? maybe ask @balthazar  i think he set that system up and might have some insight?

jonsadka:
@mcnuttandrew all fixed. There was a `.only(` hanging out. Stamped?

jonsadka:
# 😊

PR #734, opened 2018-03-02
Add Publishing section to the project README.md

jonsadka:


jonsadka:
Ah, fair enough. See https://github.com/uber/react-vis/wiki/Publishing

chrisirhc:
By the way, is this what the `prepublish` script is for on `npm`?

Ref
https://docs.npmjs.com/misc/scripts

Issue #725, opened 2018-02-18
DOMException SecurityError (code 18) checkIfStyleSheetIsImported


vasymus:
react-vis/es/utils/react-utils@checkIfStyleSheetIsImported

checking of importing of react-vis styles
via
var CSSRulesList = styleSheet.rules || styleSheet.cssRules

it's better to check via
document.styleSheets[i].hasOwnProperty('rules')
and
document.styleSheets[i].hasOwnProperty('cssRules')

otherwise,
the DOMException https://developer.mozilla.org/en-US/docs/Web/API/DOMException SecurityError (code 18) would be in CHROME (windows 10, Version 64.0.3282.167 (Official Build) (64-bit))
if there will be any of <link/> or <style/> tags appended to <head> by other plugins do not have this properties
(like, in my case, was https://github.com/tomchentw/react-google-maps, that append to <head> google-fonts <link/>)

for now if fixed it via changing prototype of component where it was used

`
XYPlot.prototype.componentDidMount = function () {
    try {
        checkIfStyleSheetIsImported()
    } catch (err) {
        console.warn(err)
        console.warn('most likely the problem is because of accessing document.styleSheets[i].rules or document.styleSheets[i].cssRules')
        console.warn('the right way is, e.g., document.styleSheets[i].hasOwnProperty("rules")')
    }
}
`

but it would be better if you change the react-vis/es/utils/react-utils@checkIfStyleSheetIsImported

mcnuttandrew:
Hey @vasymus 
Thanks for digging into this. I know some other folks have seen similar bugs and this sounds like you have a hold on a solution. I would gladly accept a PR that executed this change.

vasymus:
Hi!

Year, sure!

but very busy right now :-(

I think i'll make the PR at this holidays


>Воскресенье, 18 февраля 2018, 23:06 +02:00 от Andrew McNutt <notifications@github.com>:
>
>Hey  @vasymus
>Thanks for digging into this. I know some other folks have seen similar bugs and this sounds like you have a hold on a solution. I would gladly accept a PR that executed this change.
>—
>You are receiving this because you were mentioned.
>Reply to this email directly,  view it on GitHub , or  mute the thread .


-- 
Vasil Zchdankin6

jckr:
I'll write that PR. thanks

JFernandoGomez:
What's the status on this PR ?

jckr:
@JFernandoGomez I'm going to publish a new version today.

Issue #683, opened 2017-12-07
Custom SCSS import: Warning about missing stylesheets


texttechne:
While it's a cool feature to warn new users about missing stylesheets, it bites me a 2nd time (see #682). I import only the needed stylesheets in my main.scss file. Although everything works fine and is setup correctly I get the warning about missing stylesheets, because of these "magic lines":
```
.react-vis-magic-css-import-rule {
  display: inherit;
}
```
This code is buried in react-vis' main scss file, which I don't want to import.
So, I'm left to live with this incorrect warning (bad habit: devs soon start to ignore all kinds of warnings) or reproduce these lines in my code. However, this is internal code which shouldn't leak  into user land: You wouldn't be free to change this code then without breaking client code....

jckr:
will try to tackle this in response to #725

PR #740, opened 2018-03-11
remove the check stylesheets warnings

jckr:
There's just many issues with the check stylesheets feature, such as #725 or #683. There are cases where the warning is displayed even though stylesheets are actually imported, and cases when the checking process creates an error (even though, again, stylesheets are actually imported). 
So, I'm removing that function for now.


JFernandoGomez:
@jckr @mcnuttandrew when is this gonna be released?

jckr:
@JFernandoGomez the plan is to cut a new release every 2 weeks. There's other minor changes that I'm going to include on a release tomorrow.

Issue #705, opened 2018-01-11
Customize labels on RadialChart


micnyk:
Following `RadialChart` component usage

  ```
  const dataClicks = [
      { label: 'Views', angle: medium.views },
      { label: 'Clicks', angle: medium.clicks }
    ];

    <RadialChart
          data={dataClicks}
          width={300}
          height={300}
          showLabels={true}
        />
```

results with
![vis](https://user-images.githubusercontent.com/13575087/34827209-bb820396-f6d9-11e7-99d0-5bfcf446671e.PNG)

Is there any way to customize the labels? I want to make them bigger and move them out of chart.

As I see https://github.com/uber/react-vis/blob/master/src/radial-chart/index.js
There's no way to pass params into `LabelSeries` displayed on `RadialChart`.

PR #745, opened 2018-03-15
changes to the radial chart api to pass styles for labels

jckr:
A couple of changes to radial-charts for consistency. Now labels won't HAVE to be 12px. 
Addresses #705

### DIFF
--- a/src/radial-chart/index.js
+++ b/src/radial-chart/index.js
@@ -196,7 +196,7 @@ RadialChart.propTypes = {
   getLabel: PropTypes.func,
   height: PropTypes.number.isRequired,
   labelsAboveChildren: PropTypes.bool,
-  labelStle: PropTypes.object,
+  labelsStyle: PropTypes.object,
   margin: MarginPropType,
   onValueClick: PropTypes.func,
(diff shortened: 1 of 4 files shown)


src/radial-chart/index.js
```diff
@@ -173,27 +180,31 @@ RadialChart.propTypes = {
...
+    })
+  ).isRequired,
   getAngle: PropTypes.func,
   getAngle0: PropTypes.func,
   getRadius: PropTypes.func,
   getRadius0: PropTypes.func,
   getLabel: PropTypes.func,
   height: PropTypes.number.isRequired,
   labelsAboveChildren: PropTypes.bool,
+  labelStle: PropTypes.object,
```
mcnuttandrew: Typo

jckr: good catch thanks!

jckr:
yes the label situation is not great and this doesn't really solve it long turn. Weird that pie charts didn't get all the attention to begin with. Another direction is to have labels rotated along w/ wedges as opposed to always horizontal. I agree that having force directed labels would be neat.

Issue #751, opened 2018-03-18
Implement onTouchMove handling


kanzelm3:
This library is great! One feature we really want to implement is `touchmove` handling on mobile devices. We would really love it if when a user moved their finger across the chart, it would execute `onNearestX` again so that the behavior is more in line with desktop. It looks like it may be as simple as adding the `onTouchMove` handler here:  https://github.com/uber/react-vis/blob/master/src/plot/xy-plot.js#L447

I would be happy to implement it and submit a PR, can you see any potential issues with the idea though?

jckr:
We'll be happy to look at a PR. Thanks!

kanzelm3:
@jckr let me know what you think!

PR #754, opened 2018-03-22
Implement touch behaviors

kanzelm3:
This PR resolves https://github.com/uber/react-vis/issues/751 and adds support for touch devices for the following features:

- Crosshair functionality
- Draggable charts
- Zoomable charts

It does so in a manor that is unobtrusive and allows for some flexibility. The main gotcha with touch events is that `touchmove` is very different from the way `mouseover` and `mouseout` work, in the sense of you have to listen for `touchmove` on a larger container of elements, and use DOM APIs at our disposal to determine which child nodes the user is currently touching. We are unable to bind listeners for `touchmove` on the child components, as the target originates from the node `touchstart` was fired on.

I have added some documentation around new props on the XY Plot and I have also added tests. Please let me know if there is anything else you'd like me to add, or any concerns you may have with my implementation, and I would be happy to make changes.

### DIFF
--- a/docs/interaction.md
+++ b/docs/interaction.md
@@ -2,7 +2,7 @@
 
 Interaction in react-vis happens through _event handlers_ which are triggered by certain interactive events, such as mouse movement or clicks.
 
-These events can be implemented either at the XYPlot level or at the plot level:  
+These events can be implemented either at the XYPlot level or at the plot level:
 * At the plot level: this is for events that affect the whole chart. The mouse events that can be captured are: down, enter, leave, move. For instance, you can use `onMouseLeave` to reset the visualization when the user's mouse cursor is no longer on it.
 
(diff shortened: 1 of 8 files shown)


jckr:
Looking good, please sign the CLA so we can move ahead

kanzelm3:
@jckr I signed it, but the check isn't updating. When I go to the link it shows that I have already signed it.

kanzelm3:
Nevermind, had to sign in other account (since it had made the commits).

kanzelm3:
@jckr sorry about that, done.

adammartiska:
Is this really implemented? Because in documentation there is only onTouchEnd event, which isn't enough. Isn't there need of onTouchStart event that triggers startBrushing function?

PR #765, opened 2018-03-27
Better PropType validation for data in AbstractSeries; Tests for using Accessors

### LINKED ISSUE #724: PropType warning of data format supplied to LineSeries
I am able to supply a 2d array as `data` to LineSeries and render it with accessors `getX: d => d[0]` and `getY: d => d[1]`. However I still get an error in the console: Failed prop type: Invalid prop `data[0]` of type `array` supplied to `LineSeries`, expected `object`.

manassra:
Fixes #724. This change will prevent prop-validation for throwing a warning when users use `data` accessors (for using arrays, for example).

### DIFF
--- a/docs/xy-plot.md
+++ b/docs/xy-plot.md
@@ -54,6 +54,24 @@ XYPlot is a wrapper for series, hints, axes and other components. Most of these 
 * `size` (optional)
 * `style` (optional) - css properties as an object.
 
+Accessors can also be used to retreieve the properties above from the `data` object. For instance, the `getX` and `getY` accessors can be passed to the XYPlot object to access the `x` and `y` properties from `data` for each series. 
+
+```jsx
+<XYPlot
+  width={300}
+  height={300}
+  getX={d => d[0]}
+  getY={d => d[1]>
+  <LineSeries
+    color="red"
+    data={[
+      [1, 0],
+      [2, 1],
+      [3, 2]
+    ]}/>
+</XYPlot>
+```
+
 If the property is not passed in any of the objects, the property is not visualized. The user can override the way how properties are visualized by passing custom range, domain or type of scales to the series or the entire chart (please see [Series](series.md) for more info).
 
 Not all properties can be visualized in each series. Here's a short comparison of them:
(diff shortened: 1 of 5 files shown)


src/plot/series/abstract-series.js
```diff
@@ -39,7 +39,10 @@ const propTypes = {
   ...getScalePropTypesByAttribute('color'),
   width: PropTypes.number,
   height: PropTypes.number,
-  data: PropTypes.arrayOf(PropTypes.object),
+  data: PropTypes.arrayOf(PropTypes.oneOfType([
```
mcnuttandrew: add some documentation/testing to show that this case is supported?

manassra: you mean testing that the warning is not triggered anymore? I did test this manually, but I am not sure how to achieve this through automated tests.

mcnuttandrew: more like add a test case demonstrating that this is a valid way to input data (I don't remember writing any but I might have?) and add a note in the documentation showing that this is a supported case (again I may have already done that but i don't think i did??). If it's the case that I am being dumb and these things already exist than this gets a stamp

docs/xy-plot.md
```diff
@@ -54,6 +54,24 @@ XYPlot is a wrapper for series, hints, axes and other components. Most of these
...
 * `style` (optional) - css properties as an object.
 
+Accessors can also be used to retreieve the properties above from the `data` object. For instance, the `getX` and `getY` accessors can be passed to the XYPlot object to access the `x` and `y` properties from `data` for each series. 
+
+```jsx
+<XYPlot
+  width={300}
+  height={300}
+  getX={d => d[0]}
+  getY={d => d[1]>
```
schnerd: I know that this PR isn't implementing accessors, but now I'm curious–do we also support accessors on series components? That seems like a more common use case that people might have (and if we support it we should document it).

```
const data = [
  {fruit: 'apple', bought: 23, sold: 18},
  {fruit: 'banana', bought: 28, sold: 27}
];

<XYPlot
   ...
   getX{d => d.fruit}
/>
   <LineSeries getY={d => d.bought} />
   <LineSeries getY={d => d.sold} />
</XYPlot>
```

@jckr @mcnuttandrew

mcnuttandrew: Series level accessors *might* be supported, though if they are the functionality is untested. The only guaranteed support that we offer is on a plot level. I'd support adding accessors on series, seems like a reasonable idea.

manassra: I don't think they are supported. the accessors seem to be passed down from `XYPlot`.

PR #766, opened 2018-03-27
More stories

jckr:


jckr:
thanks for the heads up, will rebuild docs. 
using storybook adds one prerequisite to create docs, that the main react-vis project be built.

PR #768, opened 2018-03-28
Add ability to manually set label text alignment for label-series.

tumetus:
Added labelAnchorX and labelAnchorY props for label-series.

- **labelAnchorX**: Value of this prop will be used as textAnchor attribute for the <text> element. If it is not provided, the textAnchor will be calculated and set automatically depending on the label position (= as it has been set and calculated always until now)
- **labelAnchorY**: Value of this prop will be used as alignmentBaseline attribute for the <text> element. If it is not provided, the alignmentBaseline will be calculated and set automatically depending on the label position (= as it has been set and calculated always until now)

So basically these props enable user to manually set that for example label is always on the top and middle of the datapoint.

E.g. if we set labelAnchorX="middle" and labelAnchorY="text-after-edge" for the label series example, we get:

image

```javascript
import React from 'react';

import ShowcaseButton from '../showcase-components/showcase-button';
import {
  XYPlot,
  XAxis,
  YAxis,
  MarkSeries,
  LabelSeries
} from 'index';

function generateData() {
  return [
    {x: Math.random() * 3, y: Math.random() * 20, label: 'Wigglytuff', size: 30, style: {fontSize: 20}},
    {x: Math.random() * 3, y: Math.random() * 20, label: 'Psyduck', size: 10},
    {x: Math.random() * 3, y: Math.random() * 20, label: 'Geodude', size: 1},
    {x: Math.random() * 3, y: Math.random() * 20, label: 'red', size: 12, rotation: 45},
    {x: Math.random() * 3, y: Math.random() * 20, label: 'blue', size: 4}
  ];
}

export default class Example extends React.Component {
  state = {
    data: [
      {x: 3, y: 7, label: 'Wigglytuff', size: 30, style: {fontSize: 20}, rotation: 45},
      {x: 2, y: 4, label: 'Psyduck', size: 10},
      {x: 1, y: 20, label: 'Geodude', size: 1},
      {x: 4, y: 12, label: 'Ditto', size: 12, rotation: 45},
      {x: 1, y: 14, label: 'Snorlax', size: 4}
    ]
  }
  render() {
    const {data} = this.state;
    return (
      <div>
        <ShowcaseButton onClick={() => this.setState({data: generateData()})} buttonContent="UPDATE"/>
        <XYPlot
          yDomain={[-1, 22]}
          xDomain={[-1, 5]}
          width={300}
          height={300}>
          <XAxis />
          <YAxis />
          <MarkSeries
            className="mark-series-example"
            strokeWidth={2}
            sizeRange={[5, 15]}
            data={data}/>
          <LabelSeries
            animation
            allowOffsetToBeReversed
            data={data}
            labelAnchorX="middle"
            labelAnchorY="text-after-edge" />
        </XYPlot>
      </div>
    );
  }
}
```

### DIFF
--- a/src/plot/series/label-series.js
+++ b/src/plot/series/label-series.js
@@ -26,6 +26,12 @@ import {ANIMATED_SERIES_PROPS} from 'utils/series-utils';
 const predefinedClassName = 'rv-xy-plot__series rv-xy-plot__series--label';
 
 class LabelSeries extends AbstractSeries {
+  getTextAnchor = (labelAnchorX, leftOfMiddle) => {
+    return labelAnchorX ? labelAnchorX : (leftOfMiddle ? 'start' : 'end');
+  };
+  getAlignmentBaseline = (labelAnchorY, aboveMiddle) => {
+    return labelAnchorY ? labelAnchorY : (aboveMiddle ? 'text-before-edge' : 'text-after-edge');
+  };
   render() {
     const {
       animation,
@@ -39,7 +45,9 @@ class LabelSeries extends AbstractSeries {
       rotation,
       style,
       xRange,
-      yRange
+      yRange,
+      labelAnchorX,
+      labelAnchorY
     } = this.props;
     if (!data) {
       return null;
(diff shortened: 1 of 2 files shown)


src/plot/series/label-series.js
```diff
@@ -26,6 +26,12 @@ import {ANIMATED_SERIES_PROPS} from 'utils/series-utils';
 const predefinedClassName = 'rv-xy-plot__series rv-xy-plot__series--label';
 
 class LabelSeries extends AbstractSeries {
+  getTextAnchor = (labelAnchorX, leftOfMiddle) => {
```
mcnuttandrew: It'd be a little more in the style of the project to make these defined outside of LabelSeries (ie just above) in order to avoid this's.

src/plot/series/label-series.js
```diff
@@ -75,15 +83,19 @@ class LabelSeries extends AbstractSeries {
 
           const x = xVal + (allowOffsetToBeReversed && leftOfMiddle ? -1 : 1) * (xOffset || 0);
           const y = yVal + (allowOffsetToBeReversed && aboveMiddle ? -1 : 1) * (yOffset || 0);
+
+          const textAnchor = this.getTextAnchor(labelAnchorX, leftOfMiddle);
+          const alignmentBaseline = this.getAlignmentBaseline(labelAnchorY, aboveMiddle);
+
           const attrs = {
-            alignmentBaseline: aboveMiddle ? 'text-before-edge' : 'text-after-edge',
+            alignmentBaseline,
```
mcnuttandrew: you don't reuse textAnchor or alignmentBaseline, just call the functions when setting the attrs?

tumetus:
hi @mcnuttandrew 
Thanks for reviewing, good points! I made and pushed the changes you suggested.

PR #784, opened 2018-04-17
fix(docs): Fix source code link of zoomable chart

lzhuor:
- Fixed URL of the [zoomable chart](https://github.com/uber/react-vis/blob/master/showcase/examples/zoomable-chart/zoomable-chart-example.js)


mcnuttandrew:
Thanks!

Issue #616, opened 2017-09-19
Pass index through to `tickFormat` callback on AxisTicks component


jonsadka:
In the `AxisTicks` component, the `tickFormat` method [only gets passed a tick value](https://github.com/uber/react-vis/blob/master/src/plot/axis/axis-ticks.js#L179). There are instances where one would want to change the formatting of the tick based on the index of the tick (i.e. reposition the first or last tick, stagger even and odd ticks, etc.).

**Existing behavior (from [docs](https://github.com/uber/react-vis/blob/master/docs/axes.md#tickformat-optional)):**
```
function myFormatter(t) {
  return (
    <tspan x="0" dy="1em">{t}</tspan>
  );
}
```

**Desired behavior:**
```
function myFormatter(t, index) {
  return (
    <tspan x="0" dy={index === 0 ? `1em` : `-1em`}>{t}</tspan>
  );
}
```

jonsadka:
Enhancement added in https://github.com/uber/react-vis/pull/615

PR #791, opened 2018-04-26
Expose scale and tickTotal to custom tickFormat function at axis

chlumper:
In the `AxisTicks` component, the custom `tickFormat` method currently only gets passed a tick value and index. There are cases where one would want to change the formatting of the tick using the d3-scale tickFormat function with custom parameters. Which allows to format values using the various d3-format options on the scale (according to the actual value range) and make further custom adaptations to the tick label.  
e.g. use SI-prefix plus additional unit -> "1.2kWh"

**Existing behavior (from [docs](https://github.com/uber/react-vis/blob/master/docs/axes.md#tickformat-optional)):**
```javascript
function myFormatter(t, i) {
  return (<tspan>
    <tspan x="0" dy="1em">MY VALUE</tspan>
    <tspan x="0" dy="1em">{t}</tspan>
  </tspan>);
}
```
**Desired behavior:**
```javascript
function mySIPrefixFormatter(value, index, scale, tickTotal) {
  return `${scale.tickFormat(tickTotal, 's')(value)}Wh`;// -> e.g. 1.2kWh
}
```
image

**This PR provides:**
- Exposed scale and tickTotal for custom tickFormat function at axis
- Showcase example for customized axis tickFormat using scale
- Extend tickFormat function documentation 

Related to #616

### DIFF
--- a/docs/axes.md
+++ b/docs/axes.md
@@ -35,28 +35,28 @@ Which is produced via
 ## API Reference
 
 #### title (optional)
-Type: `string`  
+Type: `string`
 Shows the title for the axis.
 
 #### orientation (optional)
-Type: `'top'|'left'|'bottom'|'right'`  
+Type: `'top'|'left'|'bottom'|'right'`
 The position of the axis inside the chart.
 By default **it is already set** to `'bottom'` for `XAxis` and to `'left'` for `YAxis`. Similar to how the axis are oriented in d3-axis.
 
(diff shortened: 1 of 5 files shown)


chlumper:
Sorry I was quite busy the last days...
I'll add some tests within the next 2 days. Any recommendation which tests of the project I can use as a template?

PR #799, opened 2018-05-04
bumping ocular

jckr:


mcnuttandrew:
Looks good!

jckr:
The full context @mcnuttandrew is that we're trying to move all websites using a newer version of ocular. in parallel we are making a push on ocular to properly open source it as opposed to just have it around in a private repo.

PR #814, opened 2018-05-23
Styling LineMarkSeries

rhezab:
Instead of doing

style={{line: {stroke:"blue"}, mark: {stroke:"red"}}}

styling lines and marks separately should look like

  lineStyle={{stroke:"red"}}
  markStyle={{stroke:"blue"}}

### DIFF
--- a/docs/line-mark-series.md
+++ b/docs/line-mark-series.md
@@ -99,7 +99,8 @@ An object which holds CSS properties that will be applied to the SVG element(s) 
 ```jsx
 <LineMarkSeries
   data={data}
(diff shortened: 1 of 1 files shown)


mcnuttandrew:
Thanks for doing that! It looks like something dumb is broken about the build which is unrelated to this PR. Once we resolve the build i'll merge it in

mcnuttandrew:
I'm merging this in because it clearly doesn't break the build, something has happened in an upstream dependency that needs to be resolved that is unrelated to this PR.

PR #820, opened 2018-05-28
Correct doc hint.md; "orientation" --> "align"

rhezab:
The "orientation" prop should've been "align", and should be an Object type.


rhezab:
@mcnuttandrew

mcnuttandrew:
Very reasonable. I think the docs were right in spirit (ie they were consistent with the rest of the library), but wrong in practice. I've noted this in this @821 thread. I'll merge this in now and we'll make it correct once it comes true.

PR #829, opened 2018-06-07
Fix the broken link in series.md

### LINKED ISSUE #806: Dead link to bar series doc
While reading the documentation website, got a 404  https://uber.github.io/react-vis/website/dist/bar-series.md

whitedogg13:
This PR tries to fix #806 (Dead link to bar series doc)
 
Currently I only fix `series.md` (which maps to `documentation/api-reference/series`), but this issue  actually happens in all markdown generated pages. It causes developers hard to read documentations online.

The root cause is that we're using filename as link destination rather than url. For example, we have following content in `series.md`:

```markdown
```
The above `xy-plot.md` is a filename rather than a link. 

This patch works in my local build, and if this is acceptable I can help to further review the rest docs.

### DIFF
--- a/docs/series.md
+++ b/docs/series.md
@@ -2,17 +2,17 @@
 
 The library supports several types of series:
 
-* [ArcSeries](arc-series.md) for radial arcs such as might be found in pie charts.
-* [AreaSeries](area-series.md) for area charts;
-* [BarSeries](bar-series.md) for discrete bar charts, covers (covers HorizontalBarSeries and VerticalBarSeries);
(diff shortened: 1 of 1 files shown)


whitedogg13:
Hi, just wonder is there any problem with this PR? I see some similar ones have been merged (#830 #831 #832). Please tell me if anything wrong so I can further improve this, thanks!

mcnuttandrew:
@whitedogg13 there's no problem at all! I was on vacation or else I would have merged it sooner. thanks for doing this!!

Issue #736, opened 2018-03-05
React Error: "Element ref was specified as a string (series0) but no owner was set."


MattReimer:
I'm noticing an error when I use react-vis in an npm package that I then import into another product. 

In my project, the parent package is the "App" and the child package contains re-usable components including the [XYPlot from the Getting Started section of the documentation](https://github.com/uber/react-vis/blob/master/docs/getting-started/getting-started.md)

The plot renders fine when I view it in the child project but shows the following error when imported into the parent:

```Element ref was specified as a string (series0) but no owner was set. You may have multiple copies of React loaded. (details: https://fb.me/react-refs-must-have-owner).```

I'm really new react so I'm not 100% sure I haven't done something wrong but I thought others might have seen this already. 

### Update:

I realize this isn't much to go on but there are too many moving pieces for a really good repro description so I made a simple repo using `create-react-app` with two linked projects that "seems" to reproduce this error. 

<https://github.com/MattReimer/react-vis-repro-736>

#### Repo steps:

1. Clone the repo
2. yarn install in both child and parent
3. Link the child to the parent and boot up the parent:

```
cd child
yarn link
cd ../parent
yarn link child

yarn start
```

I'm really hoping this is something I'm doing wrong.

**Comment by jckr** [comment]
seems to be the same issue as #736. closing this one and investigating on #736.

**Comment by nareshbhatia** [comment]
I am running into the exact same issue:

```
Invariant Violation: Element ref was specified as a string (container) but no owner was set.
This could happen for one of the following reasons:
1. You may be adding a ref to a functional component
2. You may be adding a ref to a component that was not created inside a component's render method
3. You have multiple copies of React loaded
See https://fb.me/react-refs-must-have-owner for more information.
    at invariant (http://localhost:6006/static/preview.bundle.js:5450:15)
    at coerceRef (http://localhost:6006/static/preview.bundle.js:36799:25)
    at reconcileSingleElement (http://localhost:6006/static/preview.bundle.js:37504:23)
    at reconcileChildFibers (http://localhost:6006/static/preview.bundle.js:37559:35)
    at reconcileChildrenAtExpirationTime (http://localhost:6006/static/preview.bundle.js:37689:30)
    at finishClassComponent (http://localhost:6006/static/preview.bundle.js:37857:5)
    at updateClassComponent (http://localhost:6006/static/preview.bundle.js:37806:12)
    at beginWork (http://localhost:6006/static/preview.bundle.js:38427:16)
[...]

PR #840, opened 2018-06-26
Chore(ReactVis): Remove String Refs

benshope:
I just ran into https://github.com/uber/react-vis/issues/736 and this will probably fix it

### DIFF
--- a/showcase/examples/candlestick/candlestick.js
+++ b/showcase/examples/candlestick/candlestick.js
@@ -43,7 +43,7 @@ class CandlestickSeries extends AbstractSeries {
 
     return (
       <g className={`${predefinedClassName} ${className}`}
-         ref="container"
+         ref={ref => (this.container = ref)}
          transform={`translate(${marginLeft},${marginTop})`}>
         {data.map((d, i) => {
           const xTrans = xFunctor(d);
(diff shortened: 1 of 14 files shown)


radumiron:
Wow, I'm looking forward to see the build pass on this one!

jckr:
Thanks @benshope for taking this on!

benshope:
@radumiron yea I pushed this branch before checking the tests, I'll fix those!
@jckr 👍

benshope:
@jckr should I use normal methods instead of arrow functions for perf reasons?

mcnuttandrew:
> @jckr should I use normal methods instead of arrow functions for perf reasons?

Using arrow function vs explicit functions shouldn't change anything right? It's all passed through babel and transpiled into straight old fashioned js

mcnuttandrew:
Yea seems good! This feels like this might be a kinda significant change, so maybe do a minor bump

ericsoco:
@benshope the issue is not with arrow functions in `render`, it's any function declared in within `render`. The issue is that React compares current props with previous, sees that the function created last pass `!==` the fn created this pass, and therefore rerenders the component tree from there down.

Doesn't matter if it's an arrow func or pre-ES6 inline func, just that it's declared within `render`.

Ideally, callbacks-as-props are declared once, as a member of the React class or outside of the class def, and params are passed to it. This is not always possible, i.e. when context from within `render` needs to be passed in and can't be pulled e.g. off of `this.props`. In that case, you just accept the hopefully-minor perf hit.

benshope:
@ericsoco nice, yea!  I guess the exotic way of dispatching in dashbuilder avoided a significant amount of this

PR #844, opened 2018-06-27
Fix documentation links in docs/series.md and docs/xy-plot.md

sup:
Currently these links point to files in nonexistent directories. This pull request resolves those references so that they no longer 404.

### DIFF
--- a/docs/series.md
+++ b/docs/series.md
@@ -2,17 +2,17 @@
 
 The library supports several types of series:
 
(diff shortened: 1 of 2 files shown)


sup:
Noticed that these links might be used for the actual documentation site. If these are meant to be broken on Github, feel free to close. But it'd be nice if the links in the website worked as well.

PR #847, opened 2018-06-28
Adds additional tests to radar chart

NoobCIT:
These tests are a follow up to issue #796 and pull request #842 .

### DIFF
--- a/showcase/index.js
+++ b/showcase/index.js
@@ -129,6 +129,7 @@ import ArcSeriesExample from './radial-chart/arc-series-example';
 
 import BasicRadarChart from './radar-chart/basic-radar-chart';
 import AnimatedRadarChart from './radar-chart/animated-radar-chart';
+import FourQuadrantRadarChart from './radar-chart/four-quadrant-radar-chart';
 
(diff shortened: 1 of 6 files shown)


NoobCIT:
Hey thanks for looking at this!  I just wanted some clarification, so is this what you are referring to as seen in the commit file: ```showcase/radar-chart/four-quadrant-radar-chart.js``` ?

Otherwise, I'm not sure what you mean.

mcnuttandrew:
Yep good clarification:

- add an appropriate json object to this file https://github.com/uber/react-vis/blob/master/showcase/showcase-sections/radar-showcase.js (that way it'll show up in the showcase)
- to this file https://github.com/uber/react-vis/blob/master/showcase/showcase-links.js (that way it'll show up if referenced in the documentation)
- and to this file https://github.com/uber/react-vis/blob/49b08cf2842f87b41e13b400d8a413cc8c09aaa1/showcase/index.js that it'll get exported internally correctly.


Hmm that was more than I expecting to be necessary, this really should be automated so there is less maintaining a huge web of file, but that's for another PR

mcnuttandrew:
Awesome! Thanks for being so quick on this

PR #848, opened 2018-06-28
Chore(ReactVis): Remove Refs

benshope:
I was hoping that the latest version bump would solve #736 - and it does not.

This time I ran `yarn link` and made sure to get rid of remaining string `ref`s.


benshope:
@mcnuttandrew thanks!  and yea I was also surprised

PR #850, opened 2018-06-29
Added Typedefinitions for Typescript

Domino987:


mcnuttandrew:
There's a lot of copy and pasted content in there, is it possible to make this be programmatically generated?

Domino987:
The fastest way would probably be to write a small script. I will look into it.

Domino987:
I wrote a script to exchange the anys from the definitions to the type value.
If you take a look at it and check if it could be merged to the master branch.

mcnuttandrew:
I'm not seeing this script that you are referencing? Am I missing something?

Domino987:
I wrote it in python but the definition file is the new one so it is already executed. But I can upload it if that helps.

mcnuttandrew:
My interest in asking for a script to automatically assemble the type definitions is that we want to make sure that everything stays up to date. So it might be good to build node or bash script that can be part of the library and can be executed periodically to generate the definition files.

evgsil:
I've created such kind of script. It uses prop-types as source for automated type definitions generation. https://github.com/evgsil/react-vis-typings

mcnuttandrew:
@evgsil that looks super cool! i'd be delighted to see PR bringing that script into the library

medington:
Is there any reason that this PR should not be closed in favor of the work being done in #1003 ?

PR #858, opened 2018-07-05
Updated Highlight component

Domino987:
I improved the highlight component for additional Y zooming (which is one of your TODOs anyway) as a typescript file.
Needs my typescript definitions  from my other pull request xD

### DIFF
--- a/showcase/examples/zoomable-chart/highlight.js
+++ b/showcase/examples/zoomable-chart/highlight.js
@@ -10,53 +10,93 @@ export default class Highlight extends AbstractSeries {
     opacity: 0.3
   };
   state = {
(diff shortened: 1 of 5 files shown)


mcnuttandrew:
Interesting stuff! Would you mind rewriting the highlight component to be a regular javascript file? This is a javascript library, and while I totally support being friendly with typescript, we need to support our javascript users first.

Domino987:
I wrote it on javascript and added right click dragging

Domino987:
I also updated the zoom-able example to reflect the new highlight object

mcnuttandrew:
See #938

PR #865, opened 2018-07-10
Accidentally turned off all of the tests but one

mcnuttandrew:
I wish there was a lint rule for this ugh


mcnuttandrew:
> how about https://www.npmjs.com/package/eslint-plugin-no-only-tests?

I tried it out and it works great! I am very excited to have learned of this lint package.

Issue #861, opened 2018-07-06
Stateless function components cannot be given refs. Attempts to access this ref will fail


nabinked:
Following errors thrown while trying to render a Line Chart from the demo [here](https://raw.githubusercontent.com/uber/react-vis/master/showcase/plot/line-chart.js).

![Image](https://preview.ibb.co/mjEmCJ/Capture.png)

### Steps to reproduce
- Create a new react app using CRA
- Add react-vis to dependencies
- Create a Component with following content
```js

import React from 'react';

import {
  XYPlot,
  XAxis,
  YAxis,
  HorizontalGridLines,
  VerticalGridLines,
  LineSeries
} from 'react-vis';

export default class Example extends React.Component {
  render() {
    return (
      <XYPlot
        width={300}
        height={300}>
        <HorizontalGridLines />
        <VerticalGridLines />
        <XAxis title="X Axis" position="start"/>
        <YAxis title="Y Axis"/>
        <LineSeries
          className="first-series"
          data={[
            {x: 1, y: 3},
            {x: 2, y: 5},
            {x: 3, y: 15},
            {x: 4, y: 12}
          ]}/>
        <LineSeries
          className="second-series"
          data={null}/>
        <LineSeries
          className="third-series"
          curve={'curveMonotoneX'}
          style={{
            strokeDasharray: '2 2'
          }}
          data={[
            {x: 1, y: 10},
            {x: 2, y: 4},
            {x: 3, y: 2},
            {x: 4, y: 15}
          ]}
          strokeDasharray="7, 3"
          />
        <LineSeries
          className="fourth-series"
          data={[
            {x: 1, y: 7},
            {x: 2, y: 11},
            {x: 3, y: 9},
            {x: 4, y: 2}
          ]}/>
      </XYPlot>
    );
  }
}
```
- Import it to App.js
```js
import React, { Component } from 'react';
import LineChart from './LineChart';
import './App.css';

class App extends Component {
  render() {
    return (
      <LineChart />
    );
  }
}

export default App;
```
- Do a `yarn start` and see the console.

- package.json
```json
{
  "name": "my-app",
  "version": "0.1.0",
  "private": true,
  "dependencies": {
    "react": "^16.4.1",
    "react-dom": "^16.4.1",
    "react-scripts": "1.1.4",
    "react-vis": "^1.10.1"
  },
  "scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test --env=jsdom",
[...]

PR #867, opened 2018-07-10
Fix(ReactVis): Stop Setting Refs On Stateless Components

benshope:
00ffe61 tries to set the `ref` attribute on stateless components - which causes loads of console errors.  This fix only sets the `ref` attribute on components which were previously receiving it.

#861

### DIFF
--- a/tests/setup.js
+++ b/tests/setup.js
@@ -8,6 +8,9 @@ Object.keys(document.defaultView).forEach(function mapProperties(property) {
     global[property] = document.defaultView[property];
   }
 });
+console.error = (message) => {
+  throw new Error(message);
+};
 
 global.navigator = {
   userAgent: 'node.js'
--- a/tests/setup.js
+++ b/tests/setup.js
@@ -8,6 +8,7 @@ Object.keys(document.defaultView).forEach(function mapProperties(property) {
     global[property] = document.defaultView[property];
   }
 });
+// eslint-disable-next-line no-console
 console.error = (message) => {
   throw new Error(message);
 };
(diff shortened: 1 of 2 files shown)


benshope:
@mcnuttandrew and should I overwrite the previous buggy patch or push a new one?

mcnuttandrew:
I like your fix! Is there a way to write a test about this so it doesn't happen again? (if not i'm not incredibly concerned, I think we'll get a similar bunch of people yelling about it)

You should always cut new patches! Our versioning isn't super fancy or branded, so in my view it's better to have something that is actively versioned rather than try to fix buggy versions. Version 1.10.2 here we come!!

benshope:
@mcnuttandrew great, thanks - and sorry about breaking things!  Spying on `console.error` (or causing it to throw an error) during existing tests might cover this scenario and a number of other undesirable ones - let me try that - and possibly add the regression test to this PR

tests/setup.js
```diff
@@ -8,6 +8,9 @@ Object.keys(document.defaultView).forEach(function mapProperties(property) {
     global[property] = document.defaultView[property];
   }
 });
+console.error = (message) => {
```
benshope: this only works on React 16 for this specific bug, but it catches others

mcnuttandrew: It looks like you are failing lint? Also if this starts getting too rabbit hole-y, maybe push the console.error stuff into a seperate PR so this bug fix can go out?

benshope: yea agreed, I'll remove this and ship it

Issue #845, opened 2018-06-27
Real time interactive graph


saisubathra08:
I'm currently working on a project that requires me to use a real time interactive graph. I was wondering if I could use this library for that. 

Would anyone be able to give me some pointers as to how I should proceed to use this library for the same. 

Note: It should be real time and should be interactive (zoom, crosshair, etc)

Any help is greatly appreciated. 

Thanks.

jckr:
Hi, 
real time could mean several things so:
- if you mean real-time, live data, vs static data, react-vis is agnostic on that front. React-vis components will update (and if needed animate) when receiving new data through props. In other words, creating a static chart with data that won't be updated and a chart with data that will continuously refresh is the same as far as react-vis is concerned.
- if you mean interactive, then the various components have various methods to interact with them (see https://uber.github.io/react-vis/documentation/general-principles/interaction). The interactions described there are the easiest to implement with react-vis. 

There is no direct zoom method, but it can be done though such as in this example: https://uber.github.io/react-vis/examples/charts/zoomable-chart

abramjstamper:
Not to hijack this thread's topic - I have implemented these graphs in an application where I plan to use them to display real time graphs, and it works great for that. As new points are added it automatically animates the new point perfectly. However, my question is there a way to display the chart when the data prop is `[]`. In my implementation, it's blank until `data > 0`.

dgaitsgo:
@abramjstamper Could you explain more how you did that please? I'm trying to implement path transitions [like these](https://bost.ocks.org/mike/path/) but my react-vis animations are wobbling if I just set `animate` and am not sure how to interpolate the transform of the x-axis and not just between the old and new y values.

abramjstamper:
Well, animations and changing the dataset just worked out of the box for me. I set `animate={true}`, which by default will not wobble. I think (like any react component), react-vis is just doing a shallow comparison of the data to see if it changed, and if it did, it will re-render the component. The only addition of the animate prop is to animate between those changes of state.

[See this link for more info on animations](https://uber.github.io/react-vis/documentation/general-principles/animation)

dgaitsgo:
Thanks for the info! I'll see what I can do.

mcnuttandrew:
[...]

PR #882, opened 2018-07-18
No documentation on how to make empty charts

mcnuttandrew:
Addresses some points raised in #845. I built this feature a long time ago, but never documented it because it was mostly for internal use. In between now and then I had forgotten I made it, and had frequently wished for it to exist, this documentation will hopefully be a reminder this function is around.


showcase/axes/empty-chart.js
```diff
@@ -0,0 +1,52 @@
...
+
+import {
+  XYPlot,
+  XAxis,
+  YAxis,
+  VerticalGridLines,
+  HorizontalGridLines
+} from 'index';
+
+export default class EmptyChart extends React.Component {
```
benshope: could be stateless component?

PR #890, opened 2018-07-28
Chore(Docs): Update Docs Changes

benshope:
I just bumped the docs and these are some changes that were necessary along the way.  Also happy to break these changes up into separate PRs.
 - Adds a line for robustness to `publish-docs.sh`
 - Updates a broken path in  `website/src/mdRoutes.js`
 - Adds `.pretterrc` because large changes were happening in `website/src/mdRoutes.js` (adding `.prettierrc` only mitigated that problem slightly)
 - Updates out of date .lock files
 - Moves the ocular build after the storybook build because it seemed to be wiping out storybook

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,8 @@
+{
+  "singleQuote": true,
+  "trailingComma": "none",
+  "bracketSpacing": false,
+  "jsxBracketSameLine": false,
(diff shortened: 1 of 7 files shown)


benshope:
Thanks @mcnuttandrew!

Issue #896, opened 2018-08-09
How to add different color to plots on hovering mark series?


ksairamya:

mcnuttandrew:
There are lots of approaches! Here are two:
- You could stack two XYPlots on top of each other and set the cursor events to none on the top one and only provide the selected data (you should use the dontCheckIfEmpty prop on XYPlot if you go that route https://uber.github.io/react-vis/documentation/api-reference/xy-plot). This is a good approach if you have lots of data and need to avoid redraws.
- If your data is smaller (say less than like 500 points (not a hard line in anyway)) you could change the data, or you could provide a getColor function like getColor={d => d.id === MY_SELECTED_POINT ? 'red' : d.color} and on hover do a set state on the containing component with the points id (or index if you don't want to define an id).

ksairamya:
If you don't mind could you provide me a small example.

mcnuttandrew:
@ksairamya I've provided almost what I described in the second option in the above PR. I also make use of a voronoi for additional insight, though you could totally get by without using a voronoi and just putting your listener on the markseries directly.

mcnuttandrew:
Marking this as resolved, but feel free to re-open if you want

dkalamar:
Theres an example of how to handle this [here](https://github.com/uber/react-vis/blob/master/docs/interaction.md#mousing-over-near-scatterplot-marks)

PR #897, opened 2018-08-13
Add a mouse over scatterplot example

mcnuttandrew:
Addresses question #896 

Also add tests for previous voronoi example

### DIFF
--- a/showcase/misc/voronoi-scatterplot.js
+++ b/showcase/misc/voronoi-scatterplot.js
@@ -27,7 +27,6 @@ import {
   YAxis,
   HorizontalGridLines,
   VerticalGridLines,
-  LineSeries,
   MarkSeries,
   Voronoi
 } from 'index';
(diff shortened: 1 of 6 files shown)


showcase/misc/voronoi-scatterplot.js
```diff
@@ -0,0 +1,115 @@
...
+
+const nodes = lines.reduce((acc, d) => [...acc, ...d], [])
+  .map((d, id) => ({...d, id: `${id}`}));
+
+const margin = {top: 10, left: 40, bottom: 40, right: 10};
+const width = 300;
+const height = 300;
+
+const x = scaleLinear()
+  .domain([1, 4])
```
benshope: domain feels a bit like magic numbers, can I get a comment or can these be derived from the data?

showcase/misc/voronoi-scatterplot.js
```diff
@@ -0,0 +1,115 @@
...
+  .range([0, width]);
+const y = scaleLinear()
+  .domain([2, 15])
```
benshope: domain feels a bit like magic numbers, can I get a comment or can these be derived from the data?

mcnuttandrew:
See #911

PR #898, opened 2018-08-14
Update CODE_OF_CONDUCT.md

avin45h:
Added missing new line. Without this the text would appear with line above and all misconducts appeared as code of conduct :)


mcnuttandrew:
Good catch!

PR #916, opened 2018-08-20
sankey chart: fix labels' horizontal position

### LINKED ISSUE #906: sankey label position
I created a sankey diagram with the below code .  

```
import {Sankey} from 'react-vis'; 

const nodes = [{name: 'AA'}, {name: 'BBBB'}, {name: 'CC'}];
const links = [
  {source: 0, target: 1, value: 10},
  {source: 0, target: 2, value: 20},
  {source: 1, target: 2, value: 20}
];

<Sankey
  nodes={nodes}
  links={links}
  width={200}
  height={200}
/>
```
image

You can see  the position of the label 'BBBB' isn't right.  Please use react-vis@1.10.4.

Could that be fixed? Thank you.

zachguo:
fix #906

before:
screen shot 2018-08-20 at 2 14 58 pm

after:
screen shot 2018-08-20 at 2 15 27 pm

### DIFF
--- a/src/sankey/index.js
+++ b/src/sankey/index.js
@@ -132,6 +132,7 @@ class Sankey extends Component {
                 y: (node.y0 + node.y1) / 2 - marginTop,
                 label: node.name,
                 style: {
+                  textAnchor: node.x0 < width / 2 ? 'start' : 'end',
                   dy: '-.5em',
                   ...style.labels
                 }


mcnuttandrew:
This seems like it is more wrong than the previous version to me? Like now the labels are overlapping

zachguo:
@mcnuttandrew I guess `rotation` should be used when there are overlapping labels? Overlapping labels can happen whenever they are long enough. This PR is just trying to fix text horizontal alignment.
screen shot 2018-08-20 at 2 31 53 pm

mcnuttandrew:
Fair enough, in that line I'm cool with that change, I'm adding some more functionality to enable that type of per label behaviour #918

zachguo:
@mcnuttandrew That's cool. And thanks for your quick responses on these PRs!

Issue #821, opened 2018-05-28
API Audit Thread


mcnuttandrew:
There's been talk of an API audit for a while, I figured it would be good to get the ball rolling by crowdsourcing a list of points of the API that feel inconsistent. I'll start:

- hint uses orientation instead of align
- voronoi requires the user to create external scales

lzhuor:
- styles of `Legend` have to be controlled via CSS selectors instead of `style` props.

mcnuttandrew:
The style part of some series doesn't actually change the style, instead it changes the props to the constituents of the series. (I know label series does this, but there might be others) I think is a feature and not a bug and we should embrace this in some fashion.

mcnuttandrew:
See #908

Issue #355, opened 2017-04-05
Voronoi should infer default extent from XYPlot


mcnuttandrew:
To make a more automagical experience voronoi should extend off of abstractSeries to pick up it's default props, and use those to infer the the correct extent.

PR #917, opened 2018-08-20
Make voronoi component automatically infer scales and extent

mcnuttandrew:
At long last this PR addresses #355. It enables the voronoi component to automatically pick up the extent and scales for the current XYPlot. It also makes the API mostly conform to the rest of the react-vis style (#821). Finally, and most interestingly, I believe that this PR enables throwing voronois over non-linear scales. 

It is backwards compatible with the current usage, so nothing should break!

### DIFF
--- a/src/plot/voronoi.js
+++ b/src/plot/voronoi.js
@@ -1,11 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {voronoi} from 'd3-voronoi';
-import {scaleLinear} from 'd3-scale';
 
-import {
-  getAttributeFunctor
-} from 'utils/scales-utils';
+import {getAttributeFunctor} from 'utils/scales-utils';
 
 const NOOP = f => f;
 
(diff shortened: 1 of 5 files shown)


src/plot/voronoi.js
```diff
@@ -15,25 +17,39 @@ function getNodeIndex(evt) {
...
-  onClick,
-  onMouseUp,
-  onMouseDown,
-  onHover,
-  polygonStyle,
-  style,
-  x,
-  y
-}) {
+function getExtent(props) {
```
benshope: super nitpick but maybe destructure in the argument?

mcnuttandrew:
@benshope thanks for reviewing all these prs. You're a champ

benshope:
@mcnuttandrew np, thanks for making them!

benshope:
yes, absolutely!  I'll do that now

benshope:
@mcnuttandrew bumped to `1.10.7`, let me know if I should also update the docs

Issue #886, opened 2018-07-20
Some of the links on github pages are broken


truncs:
All the links similar to (website/dist/{doc}.md)
https://uber.github.io/react-vis/website/dist/arc-series.md

are broken on the website.

mcnuttandrew:
Closed due to valiant work of @gsas2

PR #921, opened 2018-08-21
Website: Fixes broken links. Docs: Fixes typo on links to interaction.md

gsas2:
Relates to: https://github.com/uber/react-vis/issues/886

This PR fixes most of the broken links on the website, an also a few typos inside area-series and linea-series docs.

I say "most" because there are some links that are not pointing to the right page (for example the ones for Axes). 
Ocular is having an issue with how it replace markdown links for actual <a> tags: It searches for the markdown filename used on the link, inside the array of routes it generates. When it finds a route that includes that text, is uses that URL as the href value of the anchor tag. 
As there is an Axes route for Example page (https://uber.github.io/react-vis/examples/showcases/axes), and another for Documentation page (https://uber.github.io/react-vis/documentation/api-reference/axes), it always stays with the first match (Example) so all the links to Axes are pointing to the Axes Example page.
The same issue applies to MarkSeries (because it finds LineMarkSeries first)

### DIFF
--- a/website/src/mdRoutes.js
+++ b/website/src/mdRoutes.js
@@ -62,6 +62,15 @@ import responsiveVis from '../../docs/examples/responsive-vis.md';
 import zoomableChart from '../../docs/examples/zoomable-chart.md';
 import gitHistory from '../../docs/examples/history-example.md';
 
+// Ocular searches for links with at least one '/' and replaces them
+// with the first route it finds that includes the same markdown name
+// (that is error prone, and that's why current Axes links will all
+// point to Example Axes page instead of Documentation Axes page)
+const fixMarkdownLinks = (markdown) => {
+  const markdownLinksRegex = /(?:\(([^()/\n]+\.md)\))/g;
+  return markdown.replace(markdownLinksRegex, '(/$1)');
+};
+
 export default [
   {
     name: 'Examples',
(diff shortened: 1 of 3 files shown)


website/src/mdRoutes.js
```diff
@@ -62,6 +62,15 @@ import responsiveVis from '../../docs/examples/responsive-vis.md';
...
+// Ocular searches for links with at least one '/' and replaces them
+// with the first route it finds that includes the same markdown name
+// (that is error prone, and that's why current Axes links will all
+// point to Example Axes page instead of Documentation Axes page)
+const fixMarkdownLinks = (markdown) => {
+  const markdownLinksRegex = /(?:\(([^()/\n]+\.md)\))/g;
+  return markdown.replace(markdownLinksRegex, '(/$1)');
+};
+
 export default [
```
mcnuttandrew: could you apply fixMarkdownLinks as a map across each of the pages and children? DRYer that way

gsas2: Sure! Thanks

gsas2: Do you mean to apply a .map on each children array? or iterating over all the structure? :thinking:

mcnuttandrew: I was imagining something like iterating over the whole structure, eg at the end add a .map(section => section.children.map(..))

gsas2: :+1: Ok

website/src/mdRoutes.js
```diff
@@ -335,3 +335,26 @@ export default [
...
+// Ocular searches for links with at least one '/' and replaces them
+// with the first route it finds that includes the same markdown name
+// (that is error prone, and that's why current Axes links will all
+// point to Example Axes page instead of Documentation Axes page)
+const fixMarkdownLinks = (markdown) => {
+  const markdownLinksRegex = /(?:\(([^()/\n]+\.md)\))/g;
+  return markdown.replace(markdownLinksRegex, '(/$1)');
+};
+
+mdRoutes.forEach(section => {
```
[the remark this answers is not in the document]
gsas2: Yes, I agree.

I started doing it with .map (as you suggested) but I was re-creating the same structure, which seemed to me unnecessary as we only need one.

That's why I ended up preferring this approach.

PR #926, opened 2018-08-21
Hexbin Series

mcnuttandrew:
hexbin series

This PR introduces a hexbin series. It is a little different than previous heatmap series that we provide in that the aggregation (though we could staple similar methods into the main heatmap series) is executed on the series level rather than being provided by the user. I think this is a valuable series to add as it might be difficult for users to do otherwise, it also sits well with the uh hexagonal branding that's in the air these days.

In the current state I have not yet written docs as I am not 100% sure about the API. I'd love to hear any feedback


benshope:
Just chatted with @nrabinowitz because he knows his hexagons and he's going to take a look

mcnuttandrew:
updated-hexes

Added docs and binning with sizing, built another example (that's pretty fun to play with tbh) and addressed @benshope 's comments. Still interested to hear comments obv

PR #938, opened 2018-08-27
Clean up of Updated Highlight component

mcnuttandrew:
This PR is a copy of #858 with some stuff cleaned up. I think the work done producing the highlight component in there is really valuable, so I'd figured i'd just finish it up (which basically meant linting and removing the typescript definitions (there is a separate PR out for those)). All credit to @Domino987, once this is in that highlight component can be extracted and moved into the main library i think it's pretty much ready.


Domino987:
Thank you for the clean up.

Issue #925, opened 2018-08-21
[Mark Series] Multiple Bubble selection by drawing a Marquee  (a bounding box)!


PalsRoy:
Hello, 
I'm currently working on a project that requires me to use a a bubble chart in which multiple bubble items can be selected at the same time. A user can perform selection by drawing a marquee (a bounding box) just like the oracle OJ charts feature. But since oracle charts are on the Jet framework, I couldn't seamlessly use it as a component in my React Project. 

Source link: http://www.oracle.com/webfolder/technetwork/jet/jetCookbook.html?component=bubbleChart&demo=selection

The most recommended Data Visualisation library being react-vis, I have started with the MarkSeries. 
I would really appreciate any pointers in this direction.

Note: As a part of my POC I am able to display and show tooltip for the same.

Any help is greatly appreciated!!

Thanks.

mcnuttandrew:
Hey @PalsRoy 

We don't currently have a 2d brushing component (as marquee/bounding box are sometimes called). There is an open pr #858 to add 2d functionality to 1d one found here http://uber.github.io/react-vis/examples/charts/zoomable-chart . You could probably copy the highlight component from that pr and use it to good effect. Sorry for the kinda hacky solution.

PalsRoy:
Hello @mcnuttandrew 

Thank you so much for responding and the great work you guys are doing with the react-vis. I did exactly that after sitting on it for a day. :-) 

Is there a way I could access the children or bubble data rendered within the highlighted zone. The idea is for a trading application to analyse, so that the highlighted orders could be pushed for trading.

`<XYPlot
         width={300}
         height={300}>
         <VerticalGridLines />
         <HorizontalGridLines />
         <XAxis />
         <YAxis />

         <Highlight onBrushEnd={(area) => {
           console.log(area);
          this.setState({
            lastDrawLocation: area
          });
          }} />

         <MarkSeries
           className="mark-series-example"
           animation
           strokeWidth={2}
           opacity="0.8"
           sizeRange={[5, 15]}
           data={[
             {x: 1, y: 10, size: 30},
             {x: 1.7, y: 12, size: 10},
             {x: 2, y: 5, size: 1},
             {x: 3, y: 15, size: 12},
             {x: 2.5, y: 7, size: 4}
           ]}/>
       </XYPlot>`

mcnuttandrew:
Thanks for your kind words!

[...]

Issue #809, opened 2018-05-16
Zoomable example has selection shifted on Firefox, making the right edge unzoomable


toddgardner:
Tested on firefox 60.0.1 (64-bit), mac os 10.12.6.

When Zooming on firefox the selection is actually shifted over several ticks to the left of where you actually clicked, making it impossible to zoom in on the right edge of the graph. I have trouble capturing a screen shot mid zoom showing the misplaced selection, but here's before:

<img width="926" alt="screen shot 2018-05-16 at 5 04 52 pm" src="https://user-images.githubusercontent.com/7430154/40144117-7c459a30-592b-11e8-8ef3-a87ce68b7968.png">

and after; this is clicking above the 80 marker and dragging off the right side of the chart:

<img width="915" alt="screen shot 2018-05-16 at 5 07 33 pm" src="https://user-images.githubusercontent.com/7430154/40144168-a9a13a02-592b-11e8-943e-557cd779ee45.png">

Note instead of starting at 80 it's shifted over a few and the end behavior, where bananas goes below apples, isn't in the zoom. The grey highlight shows over that section, significantly to the left of the mouse cursor when you start dragging.

The same example works fine with chrome on the same machine. This is affecting one of my projects so I'm digging into it.

mcnuttandrew:
If you get a solution I'd be hyped to see a PR

Domino987:
I improved the component to zoom x an y and fixed the offset at #858

mcnuttandrew:
See the new highlight component!

Issue #808, opened 2018-05-14
Zoom and click handler


nehaverma8:
I followed the example for zoom 
https://uber.github.io/react-vis/examples/charts/zoomable-chart 

In the same chart, I want to enable click ( for that I replaced LineSeries with LineMarkSeries in the above example) as I am creating a drill down chart  

However, the onValueClick event does not get fired at all. Would someone be able to help with it? 

Thanks!

mcnuttandrew:
The new highlight component should address this!

Issue #775, opened 2018-04-04
Y axis zoom


jonnydodad:
hi, I have successfully added zoom along the x-axis( using this https://uber.github.io/react-vis/examples/charts/zoomable-chart )but i want to zoom along the y-axis as well. any suggestions on how to do that? thank you

akindr:
Hey Jonny - this is certainly possible, but not with the current implementation. It would require more code to accomplish. Did you copy the `highlight.js` file into your codebase to use it, or did you import it some other way?

Regardless - there's some TODO's in the [file](https://github.com/uber/react-vis/blob/master/showcase/examples/zoomable-chart/highlight.js) that show some areas that require work to add y-axis zoom.

Ideally this would be part of the core code and not an example, but I haven't been able to migrate this yet.

thdhondt:
I have some old code laying around where I modified the existing example. 

Clicking on the x or y axis results in a zoom along the selected axis. Clicking inside the graph results in a zoom on both axis at the same time. The code is available here: https://codesandbox.io/s/3ql7o7lyr6.

Be careful for bugs, this was some of the first work I did in Javascript and react...

Domino987:
I extended the highlight example to zoom X and Y and fixed the offset. Take a look : #858

mcnuttandrew:
The new highlight component addresses this!

PR #939, opened 2018-08-28
Brushing at last

mcnuttandrew:
This is an progress PR that at long last adds brushing functionality to react-vis. Specifically it does this by evolving the highlight component from the showcase examples into a full fledged deal. 

drag-1
drag-2
drag-3

Still to do
- [x] docs
- [x] write less useless tests
- [x] click should trigger a brush and drag event that causes the highlighted region to close
- [x]  write another example showing 2d dragging

I would love any input on the API or anything else anyone might have. Big thanks to @ Domino987 whose recent work on the showcase pushed me to finish this saga up.


mcnuttandrew:
Addresses #202 #925 #809 #808 #775

akindr:
Thanks for taking this over @mcnuttandrew! Apologies for slacking 😛

src/plot/highlight.js
```diff
@@ -0,0 +1,240 @@
...
+};
+
+Highlight.propTypes = {
+  ...AbstractSeries.propTypes,
+  allow: PropTypes.arrayOf(PropTypes.oneOf(['x', 'y'])),
+  onBrushStart: PropTypes.func,
+  onDragStart: PropTypes.func,
+  onBrush: PropTypes.func,
+  onDrag: PropTypes.func,
+  onBrushEnd: PropTypes.func,
```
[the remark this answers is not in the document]
mcnuttandrew: I ended up just moving everything to be onBrush as i think that is a clearer term than draw

src/plot/highlight.js
```diff
@@ -0,0 +1,270 @@
...
+  };
+
+  _getDrawArea(xLoc, yLoc) {
+    const {startLocX, startLocY} = this.state;
+    const {allow, innerWidth, innerHeight, marginLeft, marginRight, marginBottom, marginTop} = this.props;
+    const plotHeight = innerHeight + marginTop + marginBottom;
+    const plotWidth = innerWidth + marginLeft + marginRight;
+
+    return {
+      bottom: allow.includes('y') ? Math.max(startLocY, yLoc) : plotHeight,
```
akre54: I don't like the api of using an array of strings. Can this use booleans for horizontal and vertical or x and y?

src/plot/highlight.js
```diff
@@ -0,0 +1,270 @@
...
+      top: yScale.invert(brushArea.top)
+    };
+  }
+
+  onParentMouseDown(e) {
+    const {onBrushStart, onDragStart, drag} = this.props;
+    const {dragArea} = this.state;
+    const {xLoc, yLoc} = getLocs(e.nativeEvent);
+
+    const startArea = (dragging, resetDrag) => {
```
akre54: d3 calls this brushing. Why are we introducing a new variable name and concept of dragging?

mcnuttandrew: Huh. I had thought of the action of creating a selection rectangle as the brushing step and the moving the box around as a dragging. These feel like independent behaviors to me, and browsing the d3 docs it appears that they are kinda lumped under a single term. Would an appropriate rotation of terms see

brushing -> drawing
dragging -> brushing

?

Or are you suggesting a larger change?

mcnuttandrew:
As i was preparing to wrap up this PR i wanted to make sure that the API made sense for my target application, brushing on parallel coordinates. In order to make it work I had to change a number of aspects about the architecture of the highlight component, most notably how the listeners are triggered. Rather than preparing a stacked PR which might have merge lag, I ended up just pulling the branches together into a single super PR

pc-brush

src/plot/highlight.js
```diff
@@ -0,0 +1,300 @@
...
+    );
+  }
+}
+
+Highlight.displayName = 'HighlightOverlay';
+Highlight.defaultProps = {
+  color: 'rgb(77, 182, 172)',
+  className: '',
+  enableX: false,
+  enableY: false,
```
akindr: minor nit - this is probably going to create a bunch of GitHub issues where people are like "I added Highlight but nothing happens??" because by default, the component doesn't do anything. My .02 would be to default these values to `true`, so someone can just drop this in and get highlighting in all directions.

mcnuttandrew: You are 100% correct, updating

mcnuttandrew:
Would some mind cutting a new version? I think for something big like this we should do a minor

benshope:
@mcnuttandrew yep!  I will bump a minor version

PR #956, opened 2018-09-10
Improve some of the examples

mcnuttandrew:
This PR adds an example of label series with ordinal data and improves the labeling of a recent canvas example.

### DIFF
--- a/showcase/plot/bar-chart.js
+++ b/showcase/plot/bar-chart.js
@@ -27,8 +27,26 @@ import {
   VerticalGridLines,
   HorizontalGridLines,
   VerticalBarSeries,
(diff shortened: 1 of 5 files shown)


showcase/plot/bar-chart.js
```diff
@@ -27,9 +27,27 @@ import {
...
+  {x: 'C', y: 15}
+];
+
+const blueData = [
+  {x: 'A', y: 12},
+  {x: 'B', y: 2},
+  {x: 'C', y: 11}
+];
+
+const labelData = greenData.map((d, idx) => ({
```
benshope: nit, could destructure in arg

PR #960, opened 2018-09-18
Run Prettier on All Files

benshope:
This runs the prettier on all the files.  

Some background, there were some React 16 console errors related to attribute names that were bugging me https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html.   I decided to see if there was a codemod from https://github.com/reactjs/react-codemod that would fix all console errors at once.  It looks like the codemods run prettier - so it is necessary to get consistency with prettier before trying the codemods.

@mcnuttandrew let me know if I got the prettier settings wrong or something because this seems like a lot of modified files


mcnuttandrew:
I gave a cursory scroll through all the prettier changes and they all appear to be in order. I don't really know much about using prettier, but it appears that the config you have now is maybe in disagreement with the react-vis lint configuration? It looks like almost all of the lint failures are about indentation stuff of lineoperator-linebreak. 

I zero percent care about the style of those elements, so you should feel empowered either to a) adjust the prettier config accordingly b) change the lint config. The one exception is to this is in the history showcase example, which seems to have a regex error of some sort? 

Aside: Do lint and prettier play nice together? If prettier is installed do we still use lint for static checking? Or does prettier do everything?

showcase/interaction/interaction-examples.js
```diff
@@ -20,23 +20,19 @@
...
-    y: Math.random() * 10
-  }));
+const scatterPlotData = [...Array(30).keys()].map(() => ({
+  x: Math.random() * 10,
+  y: Math.random() * 10
+}));
 
 const lineData = [...Array(3).keys()].map(() =>
-  [...Array(10).keys()].map((x) => ({x, y: Math.random() * 10})));
+  [...Array(10).keys()].map(x => ({x, y: Math.random() * 10}))
```
mcnuttandrew: okay i wasn't sold about prettier but now i am

showcase/plot/bar-chart.js
```diff
@@ -51,7 +43,7 @@ const labelData = greenData.map((d, idx) => ({
 export default class Example extends React.Component {
   state = {
     useCanvas: false
-  }
+  };
```
mcnuttandrew: i did not know that was legal syntax

benshope:
@mcnuttandrew thanks for stamping!  I'll do some manual testing before merging.  Always good to do some sanity checking for something like this

mcnuttandrew:
+10000000000 for sanity checking

benshope:
Checked it out this morning.  Should be good.  Merging 🤞

benshope:
Heads up that this made a small issue for me while `link`-ed.  I pushed a revert of the problematic bit

PR #962, opened 2018-09-19
Run error-boundaries Codemod

benshope:
This runs the `error-boundaries` codemod from https://github.com/reactjs/react-codemod


benshope:
file no longer in repo - deleting this PR

PR #965, opened 2018-09-19
Run pure-component Codemod

benshope:
This runs the `pure-component` codemod from https://github.com/reactjs/react-codemod


benshope:
will probably have to bump to the latest version of 15 for this to work

PR #966, opened 2018-09-19
Run rename-unsafe-lifecycles Codemod

benshope:
This runs the `rename-unsafe-lifecycles` codemod from https://github.com/reactjs/react-codemod


benshope:
looks like I may have to bump to the latest version of 15 for this one to work

PR #969, opened 2018-09-20
Showcase bundles don't need to be checked in

mcnuttandrew:


mcnuttandrew:
@benshope would you mind landing your codemod prs before i land this guy? i don't wanna create any zombie code

benshope:
Thanks for merging!  I was going to say that it's fine if this makes a conflict - conflicts against this will be easy to resolve

PR #979, opened 2018-09-22
Update doc routes, center all the docs examples

mcnuttandrew:


docs/hint.md
```diff
@@ -1,5 +1,7 @@
 ## Hint
```
mcnuttandrew: Bizzarely there's never been any of our many hint example on this page!

website/src/mdRoutes.js
```diff
@@ -1,65 +1,63 @@
-import presentation from '../../docs/presentation.md';
```
mcnuttandrew: i was trying to figure out where to put highlight and hexbin and I rage quit trying to mentally diff the whole thing

docs/hint.md
benshope: Seems good to add one

website/package.json
```diff
@@ -24,6 +24,6 @@
     "d3-force": "^1.0.6",
     "d3-random": "^1.1.0",
     "react-virtualized": "^9.18.5",
-    "react-vis": "^1.9.2"
+    "react-vis": "^1.11.2"
```
benshope: nice!  I wonder how this got out of date - perhaps bumping this should also be part of a script

PR #982, opened 2018-09-23
Fix markdown formatting

Martin-Nyaga:
The markdown text was incorrectly formatted, and caused text to appear inside a code block.


mcnuttandrew:
Nice catch!

Issue #959, opened 2018-09-18
RadarChart - onValueMouseEnter/onValueMouseLeave for Hints


emilyjeppson:
I have a `RadarChart` implemented and I would like to add Hints/tooltips. I was looking at the `Sunburst` example (https://github.com/uber/react-vis/blob/master/showcase/sunbursts/sunburst-with-tooltips.js), and I was hoping to do something similar. However, it looks like the `RadarChart` does not support `onValueMouseEnter` or `onValueMouseLeave` events.

Would it be possible to add them? And is this the best way to go about adding Hints to the RadarChart?

mcnuttandrew:
Hey @emilyjeppson 

I think that certainly can be done! If i'm understanding you right, you want to be able to attached listeners to the lines in the radar chart right? I think a simple version of that feature would be pretty easy to implement. Just add listeners props to the top of the radar chart and let them filter down to the line series that it has as children. I am somewhat short on time right now, but I would delightedly invite you to file a PR to this effect.

emilyjeppson:
@mcnuttandrew that sounds great! I'll dig into it and see if I can figure it out.

mcnuttandrew:
@benshope wanna do me a big favor and cut a patch?

benshope:
@mcnuttandrew yep no problem, patched to 1.11.3

Issue #771, opened 2018-04-01
How to add onClick event on the CustomSVGSeries item?


yashshah:
I'm using `CustomSVGSeries` to render my own svg and unable to add `onClick` on any of the SVG elements. How do you suggest to use onClick event on CustomSVGSeries?

Thanks!

jckr:
You're right, there's no click event listening going on on customSVGSeries. we can add in the future

mcnuttandrew:
Pull requests to address this are welcome!

PR #992, opened 2018-10-04
Radar chart tooltips

emilyjeppson:
This is a pull request for the issue I logged recently: https://github.com/uber/react-vis/issues/959

This has a few changes:

1. `RadarChart` - add the option to display axes on top of the polygons. I have charts with high opacity that cover the axes labels. I added a prop to give the option to render the axes on top (false by default). (The 1, 2, 3, 4 labels in the screenshot below illustrate this.)

2. `CustomSvgSeries` - ]add `onMouseOver`/`onMouseOut` events
  - Showcase (and tests) updated to show an example:

screen shot 2018-10-09 at 1 18 17 pm

3. `RadarChart` - Tooltips on the Polygon corners.
  - Added `onValueMouseOver` and `onValueMouseOut`
  - Circle rendered transparently with `MarkSeries` - has a `size` of 14 so that the mouse just needs to be close to the point to show the tooltip.
  - This screenshot is an example that has been added to the `showcase`. New tests have also been added.

screen shot 2018-10-09 at 1 19 48 pm

4. `RadarChart` - `onSeriesMouseOver`/`onSeriesMouseOut` events:

screen shot 2018-10-09 at 1 19 59 pm

### DIFF
--- a/showcase/radar-chart/radar-chart-with-tooltips.js
+++ b/showcase/radar-chart/radar-chart-with-tooltips.js
@@ -1,0 +1,193 @@
+// Copyright (c) 2016 - 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import React, {Component} from 'react';
+import RadarChart from 'radar-chart';
+import {Hint} from 'index';
+
+// The first 6 data elements here are to simulate a 'spider' type of radar chart -
+// similar to CircularGridLines, but straight edges instead.
+const DATA = [
+  {
+    name: 'Spider5',
+    mileage: 5,
+    price: 5,
+    safety: 5,
+    performance: 5,
+    interior: 5,
+    warranty: 5,
+    fill: '#f8f8f8',
+    stroke: '#cccccc'
+  },
+  {
+    name: 'Spider4',
+    mileage: 4,
+    price: 4,
+    safety: 4,
+    performance: 4,
+    interior: 4,
+    warranty: 4,
+    fill: 'white',
+    stroke: '#cccccc'
+  },
+  {
+    name: 'Spider3',
+    mileage: 3,
+    price: 3,
+    safety: 3,
+    performance: 3,
+    interior: 3,
+    warranty: 3,
+    fill: '#f8f8f8',
+    stroke: '#cccccc'
+  },
+  {
+    name: 'Spider2',
+    mileage: 2,
+    price: 2,
+    safety: 2,
+    performance: 2,
+    interior: 2,
+    warranty: 2,
+    fill: 'white',
+    stroke: '#cccccc'
+  },
+  {
+    name: 'Spider1',
+    mileage: 1,
+    price: 1,
+    safety: 1,
+    performance: 1,
+    interior: 1,
+    warranty: 1,
+    fill: '#f8f8f8',
+    stroke: '#cccccc'
+  },
+  {
+    name: 'Spider0',
+    mileage: 0.1,
+    price: 0.1,
+    safety: 0.1,
+    performance: 0.1,
+    interior: 0.1,
+    warranty: 0.1,
+    fill: '#f8f8f8',
+    stroke: '#cccccc'
+  },
+  {
+    name: 'Mercedes',
+    mileage: 3,
+    price: 4,
+    safety: 5,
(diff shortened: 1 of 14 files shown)


src/plot/series/custom-svg-series.js
```diff
@@ -146,13 +150,24 @@ class CustomSVGSeries extends AbstractSeries {
...
-        style
+        style,
+        propsSize: size
       });
       return (
         <g
           className="rv-xy-plot__series--custom-svg"
           key={`rv-xy-plot__series--custom-svg-${index}`}
           transform={`translate(${positionInPixels.x},${positionInPixels.y})`}
+          onMouseOver={() => {
```
mcnuttandrew: I think you can simplify this a little bit? Check out the [mark series code](https://github.com/uber/react-vis/blob/master/src/plot/series/mark-series.js) maybe?

src/radar-chart/index.js
```diff
@@ -53,7 +54,7 @@ function getAxes(props) {
     hideInnerMostValues
   } = props;
   return domains.map((domain, index) => {
-    const angle = (index / domains.length) * Math.PI * 2 + startingAngle;
```
mcnuttandrew: this is just a nit, but i tend to think the parens make this type of calculation more legible, would you mind putting them back?

showcase/radar-chart/radar-chart-with-tooltips.js
```diff
@@ -0,0 +1,190 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import React, {Component} from 'react';
+import RadarChart from 'radar-chart';
+import {Hint} from 'index';
+
+const DATA = [
```
mcnuttandrew: I think your example will stand out more if these series don't all have the same fils

src/radar-chart/index.js
```diff
@@ -28,6 +28,7 @@ import XYPlot from 'plot/xy-plot';
 import {DISCRETE_COLOR_RANGE} from 'theme';
 import {MarginPropType} from 'utils/chart-utils';
 import PolygonSeries from 'plot/series/polygon-series';
+import CustomSVGSeries from 'plot/series/custom-svg-series';
```
mcnuttandrew: I think you can do this more clearly (and probably just more easily) by using a MarkSeries to capture the points. I think it would maybe be also helpful to have a voronoi option

src/radar-chart/index.js
```diff
@@ -261,7 +353,10 @@ RadarChart.propTypes = {
     polygons: PropTypes.object
   }),
   tickFormat: PropTypes.func,
-  width: PropTypes.number.isRequired
+  width: PropTypes.number.isRequired,
+  renderAxesOverPolygons: PropTypes.bool,
+  onValueMouseOver: PropTypes.func,
+  onValueMouseOut: PropTypes.func
```
mcnuttandrew: Wow the thing I thought you were originally suggesting was an onSeriesMouseOver, this is great. While you are in here, will you add a onSeriesMouseOver? I think it should be pretty easy?

src/radar-chart/index.js
```diff
@@ -173,6 +174,74 @@ function getPolygons(props) {
...
+      return {
+        x: radius * Math.cos(angle),
+        y: radius * Math.sin(angle),
+        domain: name,
+        value: dataPoint,
+        dataName: row.name
+      };
+    });
+
+    return (
```
mcnuttandrew: so instead of mapping all of your points to customSVGSeries here you'd would just map them to an array and then use that as the input to a mark series or voronoi

mcnuttandrew:
Oh I was trying to find this pr (#771) yesterday, this resolves that

src/plot/series/custom-svg-series.js
emilyjeppson: Fixed to match `MarkSeries`

src/radar-chart/index.js
emilyjeppson: I completely agree. I accidentally had a very opinionated `JSPrettier` running that didn't like the extra parens. All 3 instances have been fixed 👍

showcase/radar-chart/radar-chart-with-tooltips.js
emilyjeppson: So the reason I did it this way is because I wanted to closely mimic my implementation, which uses the first 5 sets of data to draw polygon-shaped gridlines. (Similar to `CircularGridLines`, but with straight lines instead of round.) This makes it look more like a traditional "Spider" radar chart, which I thought could be useful for others trying to implement a radar chart? I did change them to alternating gray/white, but I could also completely remove them if you think that's better.

src/radar-chart/index.js
emilyjeppson: Awesome, I didn't know that was possible! Updated 👍

src/radar-chart/index.js
emilyjeppson: `onSeriesMouseOver` is added, with a showcase and tests.

src/radar-chart/index.js
emilyjeppson: I looked into the voronoi for a little bit, and I agree it would be cool to add here (but I didn't add it)

emilyjeppson:
Thanks @mcnuttandrew! I think this is ready for review again. I update the description to more closely align with the changes.

src/radar-chart/index.js
mcnuttandrew: Fair this PR has already gotten quite large

tests/components/custom-svg-series-tests.js
```diff
@@ -75,12 +75,23 @@ test('CustomSVGSeries: Showcase Example - CustomSVGRootLevelComponent', t => {
 });
 
 test('CustomSVGSeries: Showcase Example - CustomSVGAllTheMarks', t => {
```
mcnuttandrew: 🤩🤩🤩🤩🤩🤩🤩

showcase/radar-chart/radar-chart-with-tooltips.js
mcnuttandrew: Oh i see! In that case, i think this is probably fine for now if you leave a comment in the code about what this is the way it is. In the future it would be good to add a new component just for the radar chart to produce those spider grids (in fact would you mind creating an issue to that effect?).

mcnuttandrew:
Stamp! Before i merge it in will you resolve the merge conflict that has arisen?

mcnuttandrew:
MERGED WOOOOOOOOO

emilyjeppson:
WOOO!! Thanks @mcnuttandrew!

Issue #984, opened 2018-09-27
Better display of props in documentation


sharynneazhar:
The props in API references are hard to read.

<img width="292" alt="screen shot 2018-09-27 at 3 45 37 pm" src="https://user-images.githubusercontent.com/10108593/46173789-e50ea980-c26c-11e8-8449-c7d28fa4224d.png">

Can we update the docs so the props stand out more? E.g. 

<img width="388" alt="screen shot 2018-09-27 at 3 52 48 pm" src="https://user-images.githubusercontent.com/10108593/46173973-682fff80-c26d-11e8-9f82-26dc4a12566a.png">

PR #993, opened 2018-10-07
Make props in documentation easier to parse (#984)

Wattenberger:
I took a stab at making the props in the docs easier to parse - I like when the attributes are very visible, but totally understand if the style is too bold.

image

Since props in the docs are just generic headers (h3, h4, h5), I targeted h4 and h5s to prevent changing headers in other parts of the docs. There are a few prop declarations that use `h3`s, which are updated in the second commit :raised_hands:

### DIFF
--- a/docs/axes.md
+++ b/docs/axes.md
@@ -123,7 +123,7 @@ Height of the axis in pixels. **Already set by default**, but can be overridden 
 #### animation (optional)
 See the [XYPlot](xy-plot.md)'s `animation` section for more information.
 
-### style (optional)
+#### style (optional)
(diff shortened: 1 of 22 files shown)


Wattenberger:
Ah yes, good call! Updated to the underline and moved the blue highlight to code elements following a h4/h5. This also includes inline code elements in the text - I could go either way on the blue highlight in general. I think it's a little loud, but not necessarily too much, plus it gives the docs a bit of personality which I like.

image

I went through and added new lines around `Type:` and `Default:` statements - I think things are looking great!

Issue #860, opened 2018-07-05
Discrete Color Legend strokeDashes and strokeWidth


cvalmonte:
The discrete color legend component should include props to style the stroke (dashed, width, etc.) so that it will match the line series in my graphs.

mcnuttandrew:
I would accept a PR to address this issue!

olefrank:
`StrokeWidth` doesn't seem to work. 

```
const legendItems = [
    {
      title: 'Payments',
      color: '#40A695',
      strokeWidth: 3,
    },
    {
      title: 'Expected payments',
      color: '#999',
      strokeStyle: 'dashed',
      strokeWidth: 3,
    },
    {
      title: 'Costs',
      color: '#ED7C33',
      strokeWidth: 3,
    },
  ]

...

<DiscreteColorLegend items={this.legendItems} orientation="horizontal" />
```

![skaermbillede 2018-12-07 kl 13 58 34](https://user-images.githubusercontent.com/3809708/49649055-3e8f1380-fa28-11e8-83d2-ea7c88846282.png)

talc23:
Is there a solution?

kenfehling:
It looks like strokeWidth doesn't work because the parent div with class `rv-discrete-color-legend-item__color` has a preset height of 2, so the maximum stroke width that shows is 2px

talc23:
I've added this to my css file as a temp WA:
.rv-discrete-color-legend-item__color {
height: 10px;
}

The question is why on the showcase examples the strokewdith works and on installed npm packages it doesn't?
for example, I've played with this example on my local react-vis repo:
https://github.com/uber/react-vis/blob/master/showcase/plot/clustered-stacked-bar-chart.js

changed code to:
          <DiscreteColorLegend
            style={{position: 'absolute', left: '50px', top: '10px'}}
            orientation="horizontal"
            items={[
              {
                title: 'Apples',
                **strokeWidth:50**
              },
              {
                title: 'Oranges',
                color: '#79C7E3'
              }
            ]}
          />

And got this:

![image](https://user-images.githubusercontent.com/25937994/55617291-715dd400-579c-11e9-9483-48c596046eb6.png)

perryclee01:
Running into the same issue with stroke-width not working due to the inline height being set. You'll have to go into /react-vis/es/legends/discrete-color-legend-item.js and change the height in line 57 and then restart your webpage until there is a fix.

PR #994, opened 2018-10-08
add stroke properties to discrete color legend (#860)

Wattenberger:
Added the ability to set stroke properties (strokeDasharray, strokeStyle, strokeWidth) of Discrete Color Legend items to match Line Series components. At first I considered mimicking the `colors` attribute and letting users add a `strokeWidths` array, but it seemed to over-complicate the api, so I settled with grabbing those properties from an `item`'s object.

The legend items now render an svg path instead of a 2x14px div, which was necessary to match the dash-array property, but it also might help keep the legend consistent with the chart.

I edited one of the legend examples to show the implementation - happy to remove it if it's too heavy-handed.

image

Happy to hear any feedback, thanks!

### DIFF
--- a/docs/legends.md
+++ b/docs/legends.md
@@ -18,8 +18,8 @@ Currently following types of legends are supported:
 <!-- INJECT:"VerticalDiscreteColorLegendExampleWithLink" -->
 
 #### items (required)
-Type: `Array<string|{title: string, color: String, disabled: boolean}|react element>`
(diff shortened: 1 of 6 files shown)


Wattenberger:
Oh shoot! Tests, of course!

I updated the tests to match the svg styles/elements and added stroke-width and stroke-dasharray tests. Thanks for the speedy response!

mcnuttandrew:
Looks good! Thanks again

PR #996, opened 2018-10-08
Update hint to prevent crash on date x values

Domino987:
With this Change, the Hint does not crash, if a date is provided as x value

### DIFF
--- a/src/plot/hint.js
+++ b/src/plot/hint.js
@@ -22,6 +22,8 @@ import React, {PureComponent} from 'react';
 
 import PropTypes from 'prop-types';
 
+import {transformValueToString} from 'utils/data-utils';
+
 import {getAttributeFunctor} from 'utils/scales-utils';
 
 /*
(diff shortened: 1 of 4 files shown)


src/plot/hint.js
```diff
@@ -70,7 +70,7 @@ const ORIENTATION = {
  */
 function defaultFormat(value) {
   return Object.keys(value).map(function getProp(key) {
-    return {title: key, value: value[key]};
+    return { title: key, value: Object.prototype.toString.call(value[key]) === '[object Date]' ? value[key].toDateString() : value[key] };
```
mcnuttandrew: maybe break this out into a function and use typeof instead?

Domino987: I will look into it to change it. What ist your linting for the tests?

mcnuttandrew: I'm not sure I understand your question? We use eslint, you can run it locally by running 

```sh
npm run lint
```

Domino987: This throws 24237 erros:  Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style.
I thought that i have to change my linting to ignore those for example.

Domino987: Seems like Windows is the problem. I will work on it and try to resolve it.

mcnuttandrew:
You should make sure to squash your commits to address the CLA issue above

Domino987:
Ok I will. My linter is pretty strict (everything alphabetical ordered and such) so that makes testing difficult, but I will change it.

Domino987:
See #1077. Cleaned it up and reopened

PR #997, opened 2018-10-08
Updates Crosshair to prevent crash on date as x

Domino987:
Prevents crashing of the crosshair like in #996


mcnuttandrew:
This PR and the other one are really similar, maybe just merge them and make the date checking function a util?

Domino987:
Yes i will do that.

PR #1001, opened 2018-10-10
Update HorizontalDiscreteColor example to show off stroke style properties

Wattenberger:
I thought that adding another legend might crowd the examples, so I changed the horizontal legend example to use the new stroke style properties. Let me know if you think this is overloaded and I can break it out!

image


Wattenberger:
happy to! :raised_hands:

PR #1003, opened 2018-10-10
Typescript definition file generator

evgsil:
Tool consist of too scripts:

**generate-types.js** creates typescript definition file index.d.ts in root folder for whole library based on specified prop-types.

**generate-modules.js** creates *.d.ts files in **es** and **dist** folders. Each file just re-exports declaration from index.d.ts for individual component. This covers case with partial imports like `import XAxis from 'react-vis/es/plot/axis/x-axis';`


typegen/generate-modules.js
```diff
@@ -0,0 +1,34 @@
+const fs = require('fs');
+const path = require('path');
+
+function generateSubModules(mainJsFile, basePath) {
+  const pathMap = new Map();
+  const regex = /export\s+(\S+|\{[\s\w,]*\})\s+from\s+'(\.\/)?(\S+)'/g;
```
mcnuttandrew: Can you either write a test describing what this does or add some comments? Complex regex are tough to maintain without a little help

typegen/generate-modules.js
```diff
@@ -0,0 +1,34 @@
...
+    if (match) {
+      pathMap.set(match[1], match[3]);
+    }
+  } while (match);
+
+  Array.from(pathMap).forEach(([name, componentPath]) => {
+    let fullPath = path.join(basePath, `${componentPath}`);
+    fullPath = fs.existsSync(fullPath) && fs.statSync(fullPath).isDirectory()
+      ? path.join(fullPath, 'index.d.ts')
+      : fullPath + '.d.ts';
```
mcnuttandrew: How about template rather than concatenation

typegen/generate-modules.js
```diff
@@ -0,0 +1,34 @@
...
+  do {
+    match = regex.exec(mainJsFile);
+    if (match) {
+      pathMap.set(match[1], match[3]);
+    }
+  } while (match);
+
+  Array.from(pathMap).forEach(([name, componentPath]) => {
+    let fullPath = path.join(basePath, `${componentPath}`);
+    fullPath = fs.existsSync(fullPath) && fs.statSync(fullPath).isDirectory()
```
mcnuttandrew: can you make this conditional a named boolean so that future reads can understand what it is doing a little easier?

typegen/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "typegen",
+  "version": "1.0.0",
+  "description": "Generates TypeScript definition file",
+  "main": "generate-types.js",
+  "scripts": {
+    "generate-types": "node generate-types.js",
```
mcnuttandrew: maybe make this babel-node instead of node. I just tried running this locally (i'm on node v8) and it failed

typegen/header.d.ts
```diff
@@ -0,0 +1,270 @@
+import {
```
mcnuttandrew: Is there any way to verify that these are correct? I don't know typescript, so these look correct to me as far as I understand them, but is there an automated way to check them?

typegen/generate-types.js
```diff
@@ -0,0 +1,178 @@
...
+reactVis.AbstractSeries.propTypes = {
+  ...reactVis.AbstractSeries.propTypes,
+  color: colorType,
+  stroke: colorType,
+  fill: colorType,
+  opacity: PropTypes.number,
+}
+
+function splitPath(path) {
+  let [compName, ...propName] = path.split('.');
```
mcnuttandrew: typically the style of the library is to avoid lets as much as possible, either through variable renaming or just chains of stuff. I hadn't seen this usage of the ...operator before and I really like it, but just keep that let thing in mind

typegen/package.json
evgsil: What error did you get? Should work well in node v8.10 and above.
This tool loads built version of react-vis from the **dist** folder:
`const reactVis = require('../dist');` 
Maybe problem is here. Have you tried to run build from project root? It should run generate-types as final step of build

typegen/generate-modules.js
evgsil: I have retrieved generate-modules.js script. Now it uses runtime information of react-vis library instead of parsing index.js

typegen/header.d.ts
evgsil: The only automated way that came to my mind is just to try compile it with typescript compiler. Compiler will check definition file to some extent. However it only makes sens to compile final index.d.ts, because header.d.ts is just part of final file. It only includes some manual definitions that we can use later while iterating through prop-types

typegen/package.json
mcnuttandrew: Here's the error i got, I ran the generate-type script from package.json

```sh
 mcnutt$ npm run generate-types

> react-vis@1.11.2 generate-types /Users/mcnutt/react-vis
> cd typegen && npm install && npm run generate-types && npm run generate-modules

npm WARN typegen@1.0.0 No repository field.

audited 1 package in 2.894s
found 0 vulnerabilities


> typegen@1.0.0 generate-types /Users/mcnutt/react-vis/typegen
> node generate-types.js

/Users/mcnutt/react-vis/typegen/generate-types.js:11
  ...reactVis.AbstractSeries.propTypes,
  ^^^

SyntaxError: Unexpected token ...
    at createScript (vm.js:74:10)
    at Object.runInThisContext (vm.js:116:10)
    at Module._compile (module.js:533:28)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Function.Module.runMain (module.js:605:10)
    at startup (bootstrap_node.js:158:16)
    at bootstrap_node.js:575:3
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! typegen@1.0.0 generate-types: `node generate-types.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the typegen@1.0.0 generate-types script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/mcnutt/.npm/_logs/2018-10-18T03_55_21_272Z-debug.log
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! react-vis@1.11.2 generate-types: `cd typegen && npm install && npm run generate-types && npm run generate-modules`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the react-vis@1.11.2 generate-types script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/mcnutt/.npm/_logs/2018-10-18T03_55_21_362Z-debug.log
```

evgsil: I have refactored code in order to be compatible with node v6.14+

mcnuttandrew: That's really good, but I think it would be better to keep out scripting strategy consistent across the library. I think a better strategy would be to remove the package.json from /typegen and merge it with the main package.json (using prop-types-to-dts as a dev dep) and placing the generate-types command in scripts section of package. I tried this approach out locally and it appears to work well.

package.json
```diff
@@ -79,6 +82,7 @@
     "node-sass": "^4.9.3",
     "nyc": "^13.0.1",
     "prettier": "^1.14.2",
+    "prop-types-to-dts": "^0.3.2",
```
mcnuttandrew: I'm really nervous about introducing a relatively unknown package like this. Your package looks fine but I'm not a typescript expert, and I have a hard time vetting it beyond a superficial level. Given that the main body of the work is done in just one file, will you check that one (index.js) file into the typegen folder so we can keep an eye on it?

mcnuttandrew:
@jckr @benshope @btford (or anyone else) does anyone with typescript knowledge wanna comment on this PR? I feel a little out of my depth

typegen/header.d.ts
evgsil: I can include typescript compiler check of generated artifact like this:
```suggestion
"generate-types": "node ./typegen/generate-types.js && node ./typegen/generate-modules.js && tsc --noEmit --strict index.d.ts"
```
However this will require few additional devDependencies: 
"typescript" - compiler itself and "@types/react" - type definitions for React library

xogeny:
Excellent work!  @mcnuttandrew as to your question about whether these types are correct, I'm pretty sure you'll hear from TypeScript users if there are issues.  Since it looks like this is all based on prop-types, it should be consistent with your prop-type definitions and if issues are found (and the `generate-types` script doesn't have any bugs) then it might actually help you in finding errors in your non-TypeScript code.  Seems like a win-win.

xogeny:
Actually, using the type definitions along side some of the `react-vis` example code I pretty quickly found what I think are some mistakes in your `PropTypes` definitions.  Note, this could be because I'm misunderstanding the precise semantics of `PropTypes`, so don't take this as a definitive bug, it could easily be my own ignorance.  

The properties of `<LineSeries>` are defined [here](https://github.com/uber/react-vis/blob/master/src/plot/series/line-series.js#L114-L119) and inherited from `<AbstractSeries>` which is defined [here](https://github.com/uber/react-vis/blob/master/src/plot/series/abstract-series.js)

But consider [this example](https://github.com/uber/react-vis/blob/master/showcase/plot/line-chart.js).  In it, you use `<LineSeries>` with the properties `className`, `strokeDasharray` and `data={null}`.

Let's start with `className`.  I looked through the `PropTypes` documentation, I see nothing in there that says that `className` is automatically a property of all components.  So it seems to me using `className` on a `<LineSeries>` is an error.  You should have `className` defined in your `PropTypes`.  I'm not sure why the `PropTypes` validation isn't catching this error, but the TypeScript compiler definitely saw it. 😄

The same is true of `strokeDasharray`.  Yes, this is defined for some DOM elements, but there is nothing in your property definitions that implies that it should be applied.  Don't get me wrong, I know [you are using it](https://github.com/uber/react-vis/blob/master/src/plot/series/line-series.js#L77).  But the point is that there is an inconsistency between how the properties are defined (not including `strokeDasharray`) and how they are used (destructuring `strokeDasharray`) and the TypeScript compiler won't allow these to be inconsistent and so these kinds of things pop up immediately.

That being said, I think I've identified one defect in the `generate-types` scripts.  Specifically, it has to do with the `data` attribute which is defined [here](https://github.com/uber/react-vis/blob/master/src/plot/series/abstract-series.js#L42-L44).  With `PropTypes`, it seems that the default is that types are optional unless you chain `.isRequired` on the end and, furthermore, **the `prop-types` validator considers `null` as equivalent to `undefined`**.  But I note that the type definitions provided [in this repo](https://github.com/evgsil/react-vis-typings/blob/master/react-vis.d.ts) (which I'm assuming, perhaps incorrectly, were generated using `prop-types-to-dts`) correctly mark the `data` property as optional, but don't allow for the possibility of `null` as a value (only `undefined`).  So that is kind of a bug in `prop-types-to-dts` although I think it is actually cleaner to just include `data={undefined}` or just leave the `data` property off altogether (both are acceptable using the generated type definitions).

In the first two cases (where the property definitions are not consistent with usage), it seems like the PR is actually trying to compensate for that (*e.g.,* [here](https://github.com/uber/react-vis/pull/1003/files#diff-561d10475123f7a25eb2829695132fc4R10)).  Of course, a better long term solution would be to amend the property definitions so that the generator scripts don't need to tweak anything.

xogeny:
Another note is that the types defined [here](https://github.com/evgsil/react-vis-typings/blob/master/react-vis.d.ts) don't define `ChartLabel`.  Again, those definitions may not be representative of what gets generated by this PR.  But since the types themselves aren't included in the PR, I can't really comment on them without checkout out the PR and building everything.

evgsil:
> That being said, I think I've identified one defect in the `generate-types` scripts. Specifically, it has to do with the `data` attribute which is defined [here](https://github.com/uber/react-vis/blob/master/src/plot/series/abstract-series.js#L42-L44). With `PropTypes`, it seems that the default is that types are optional unless you chain `.isRequired` on the end and, furthermore, **the `prop-types` validator considers `null` as equivalent to `undefined`**. But I note that the type definitions provided [in this repo](https://github.com/evgsil/react-vis-typings/blob/master/react-vis.d.ts) (which I'm assuming, perhaps incorrectly, were generated using `prop-types-to-dts`) correctly mark the `data` property as optional, but don't allow for the possibility of `null` as a value (only `undefined`). So that is kind of a bug in `prop-types-to-dts` although I think it is actually cleaner to just include `data={undefined}` or just leave the `data` property off altogether (both are acceptable using the generated type definitions).

Unfortunately, at the moment, there are no possibility to make distinction [between null and undefined](https://github.com/facebook/react/issues/3163) with prop-types. I've looked through definitions created for several most popular react libraries ([react-router](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c2e4b83961326a32cd3dce1f26d7b20c1e816252/types/react-router-dom/index.d.ts), [react-virtualized](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9acca8ab3e8b198aa3332f3ff6f03c0a6e3ce1e9/types/react-virtualized/dist/es/Grid.d.ts), [material-ui](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Button/Button.d.ts)) and in most cases they just use optional ('?') param for non-required props. I can easily add logic that will comply with formal behavior of isRequired prop, and for non-required props generate something like this: `caption? : string | null;`. However in this case definition file IMHO will look a bit messy as we have more non-required than required props

> Another note is that the types defined [here](https://github.com/evgsil/react-vis-typings/blob/master/react-vis.d.ts) don't define `ChartLabel`. Again, those definitions may not be representative of what gets generated by this PR. But since the types themselves aren't included in the PR, I can't really comment on them without checkout out the PR and building everything.

I don't maintain that repository anymore, the most recent changes in this pull request. ChartLabel is new component. This tool will pickup new components like this and generate definitions automatically.

awjreynolds:
It does sound like auto-generation (specifically null vs undefined) is a problem.

Would it be better to adopt this into the library? https://github.com/evgsil/react-vis-typings/blob/master/react-vis.d.ts

We have already used this d.ts to highlight issues within our code.

paddotk:
Seems like this has been forgotten at some point. There is a passing PR so would someone please merge it?

benshope:
Hey @paddotk  sorry for being a little slow and thanks for contributing.  I'm a little busy today but I will move this forward this weekend

benshope:
Okay, so while it would be more ideal to have typescript or flow actually used in the project this feels like a net positive for users today and a step in a direction that will benefit the project long-term (having real static types that do not require this script).  I do not see a `react-vis` entry in `DefinitelyTyped` that we would be conflicting with so this doesn't appear to be a breaking change for any widely-used types.  I ran the script and it didn't seem to have any problems.  If @mcnuttandrew (or anyone else) doesn't object I will check this again tomorrow when I'm wide awake and ship it as a minor version bump.

paddotk:
@benshope Hey, thanks! Was just not sure whether it got off the radar over time. But good to hear, this will help us a great deal :)

mcnuttandrew:
Oh man this PR was in a different place than I remembered it being!! I approve!! Minor away!

evgsil:
I've added some minor fixes that i had found recently. However in order to improve and completely fix generated typing's some fixes should be applied to PropTypes of react-vis components itself. There are a number of missing and redundant PropTypes, however i think it will be a subject of another PR

benshope:
I read this over in detail and love the idea of improving react-vis's support of static types and having these scripts in `react-vis`.

I am a little worried that shipping them directly connected to `react-vis`'s versions would:
 - make typescript required knowledge for maintainers
 - make these scripts (with their regexes and `if` statements) required knowledge for maintainers
 - risk eventually delivering subtly-incorrect types to library consumers

My instinct is to:
 - add these scripts to `react-vis` as `npm run generate-types` but not called by  `npm build`
 - an addition to the ReadMe about using `react-vis` with typescript
 - a separate PR to https://github.com/DefinitelyTyped/DefinitelyTyped containing these generated types tied to a specific version of `react-vis`

@evgsil I'm sorry to propose more changes but are these concerns reasonable?  Is this solution something you would be okay with? Let me know your thoughts!

kevindelsh:
Is this still in the pipeline?

benshope:
@kevindelsh I think what I proposed was continuing to have @evgsil maintain these types either here or in his fork and to ship the types as part of `definitelytyped`.  I think that remains inclusive towards people who use typescript but doesn't require us to maintain these scripts (or depart from the types) when we make minor changes.  I'm open to feedback on this suggestion and I could make a PR to definitelytyped myself (or @evgsil could do it) if that sounds okay

fstylermiller:
I would really appreciate seeing this as a part of `DefinitelyTyped`!

rluiten:
I would also love to see this as either part of react-vis npm package or in definitely typed.
I cloned this repo and locally merged the source of this PR to generated a typedef.

1.  The default react-vis repo `npm run build` fails on windows 10, 
     This can be fixed by adding cross-env to dev dependencies and adjust the build rule in package.json
     to ( prefix `BABEL_ENV=` with `cross-env` )

     `"build": "npm run clean && babel src -d dist --copy-files && cross-env BABEL_ENV=es babel src -d es --copy-files && node-sass src/main.scss dist/style.css --output-style compressed && npm run build:browser && npm run generate-types",`

2.  The generated type def for XYPlot appears to be missing properties.
     The immediate ones I am using and were raised as type errors are `xDomain`,  `xType`, `yDomain`,
     I have included the generated types  of XYPlot here as a reference.
     The generated types are no surprise when I look at propTypes of XYPlot.
     So it appears propTypes of XYPlot may be incomplete?

```typescript
  export interface XYPlotProps {
    animation?: string | {
      stiffness?: number;
      nonAnimatedProps?: Array<string>;
      damping?: number;
    } | boolean;
    className?: string;
    dontCheckIfEmpty?: boolean;
    height: number;
    margin?: {
      left?: number;
      top?: number;
      right?: number;
      bottom?: number;
    } | number;
    onClick?: RVMouseEventHandler;
    onDoubleClick?: RVMouseEventHandler;
    onMouseDown?: RVMouseEventHandler;
    onMouseUp?: RVMouseEventHandler;
    onMouseEnter?: RVMouseEventHandler;
    onMouseLeave?: RVMouseEventHandler;
    onMouseMove?: RVMouseEventHandler;
    onTouchStart?: RVTouchEventHandler;
    onTouchMove?: RVTouchEventHandler;
    onTouchEnd?: RVTouchEventHandler;
    onTouchCancel?: RVTouchEventHandler;
    onWheel?: RVWheelEventHandler;
    stackBy?: 'x' | 'y' | 'radius' | 'angle' | 'color' | 'fill' | 'stroke' | 'opacity' | 'size';
    style?: CSSProperties;
    width: number;
  }
  export class XYPlot extends Component<XYPlotProps> {}
```

andrew-ironforge:
Could someone merge this PR if accepted, please?

benshope:
@andrew-ironforge @fstylermiller I have given this some thought and the only long-term sustainable way of supporting static types is going to be through `definitelytyped` (or `flow-typed` for flow) until we choose to use static types within the codebase.  A standalone set of types in this codebase would cause more headaches and bugs for users of the codebase than a set of stubs would.  This approach also has precedent - much of d3 relies on type-safety through 3rd-party definitions.  The work of anyone who makes a PR against `definitelytyped` to add these types would be greatly appreciated.  Please let me know if `definitelytyped` fails to meet the needs of your project

fstylermiller:
@benshope I'm working on creating a PR into DefinitelyTyped, but would like some support with flushing out the last bit.

I'm looking to further specify all `Function` types (required for the merge). I'm having trouble understanding what some of the inputs and outputs should be. These are the only two I had trouble defining.
```
LineSeries
  - curve

DecorativeAxis
  - tickValue
```

The others were either generated, or I've filled in.

rluiten:
@fstylermiller I have a few changes I made to the locally generated one I've been using.

My changes add inference to all in a series data render component, eg LineSeries.
This alllows the type of the data passed to LineSeries to be inferred as the data provided in the callbacks definable on LineSeries component, they helped me a fair bit.

I added a few missing missing fields but only the ones I specifically discovered were missing when I tried to use them, these are ones that the generation script didnt have because they were likely not in the declared proptypes of the code base.

Is your work visible, I can take a look next week see if anything I've done would fit and be worth adding.

fstylermiller:
@rluiten Cool, I've forked publicly and have been making changes. There were a few minor inconsistencies. I can add you to the fork if you'd like.

rluiten:
Just a link to your public fork is fine, you mean forked definitelytyped ?
or something else?

fstylermiller:
Yep! We should merge and work through conflicts. This is it: https://github.com/fstylermiller/DefinitelyTyped

rluiten:
@fstylermiller i created this on your repo, https://github.com/fstylermiller/DefinitelyTyped/wiki/Just-a-way-to-make-some-notes-for-Tyler because I couldnt create an issues to discuss stuff directly and not muddy this thread more than we need to.

pixelass:
Any news here? I still can't find type-definitions for this library.

mcnuttandrew:
there are type definitions now available at @types/react-vis

PR #1019, opened 2018-10-23
more flexibility when rendering tick labels with tickFormat

bodyflex:
right now the value rendered from `tickFormat` will always be wrapped in a `<text>` node which is quite limiting if you want to do more flexible stuff. this change checks if the value that's returned from `tickFormat` is a React element and renders it instead of the `<text>` in case it is one. the custom element also gets passed the width of the tick container and the total amount of ticks as props in case it wants to do some responsive stuff.

~as such, this would be a breaking change, but it could also be pretty easily modified to wrap those SVG elements with a `<text>` that can be descendants of `<text>` per spec.~ 

if you think this change is OK, i'll write some tests for it and then hopefully it could be merged :)

### DIFF
--- a/docs/axes.md
+++ b/docs/axes.md
@@ -87,7 +87,7 @@ function mySIPrefixFormatter(value, index, scale, tickTotal) {
 }
 ```
 
-**Note!** The return value will be wrapped with a `<text>` node if it's a string, `<tspan>`, or `<textPath>`. In all other cases the returned element will replace the `<text>` node and receive two additional props: `containerWidth` and `tickCount`. This way you can e.g. render a `<div>` to truncate long labels:
+**Note!** The return value will be wrapped with a `<text>` node if it's a string, `<tspan>`, or `<textPath>`. In all other cases the returned element will replace the `<text>` node. In case it's a custom React element it will also receive two additional props: `containerWidth` and `tickCount`. This way you can e.g. render a `<div>` to truncate long labels:
 
 ```javascript
 const MyLabel = props => (
@@ -107,7 +107,11 @@ const MyLabel = props => (
 function myFormatter(value) {
   return <MyLabel>{value}</MyLabel>;
 }
+
+<XAxis tickFormat={myFormatter} />
 ```
+
+<!-- INJECT:"CustomAxisTickElement" -->
 
 #### tickSize (optional)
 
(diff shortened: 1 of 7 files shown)


bodyflex:
i tried to find existing tests regarding `tickFormat` but i guess there are none yet. should i try to write some?

bodyflex:
@mcnuttandrew added some tests now!

docs/axes.md
```diff
@@ -87,6 +87,28 @@ function mySIPrefixFormatter(value, index, scale, tickTotal) {
...
+    >
+      {props.children}
+    </div>
+  </foreignObject>
+);
+
+function myFormatter(value) {
+  return <MyLabel>{value}</MyLabel>;
+}
+```
```
mcnuttandrew: Maybe also include an example of using this with the axes?

showcase/axes/custom-axis-tick-element.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  render() {
+    return (
+      <XYPlot
+        width={300}
+        height={300}
+        xType="linear"
+        xDomain={[0, 4]}
+        yType="linear"
```
mcnuttandrew: nit: unnecessary to list xType and yType

tests/components/axis-tick-format-tests.js
```diff
@@ -0,0 +1,47 @@
...
+  );
+  t.equal(
+    ticks.at(4).find('text').text(),
+    'Label',
+    'renders text inside text node'
+  );
+  t.end();
+});
+
+test('Axis Format: passes props to custom element', t => {
```
mcnuttandrew: nice test

docs/axes.md
```diff
@@ -87,6 +87,28 @@ function mySIPrefixFormatter(value, index, scale, tickTotal) {
...
+
```
mcnuttandrew: You can include your new showcase example in here! Just use the magic comment like:

```md
<!-- INJECT:"CustomAxesWithLink" -->
```
But with the appropriate chart name

bodyflex:
all review comments should be fixed now @mcnuttandrew

mcnuttandrew:
Sorry for the long delays! This looks good. stamp stamp

bodyflex:
would be awesome if we got a new release with this and other fixes @mcnuttandrew :)

mcnuttandrew:
@benshope mind cutting a patch?

benshope:
@mcnuttandrew happy to!  Bumped to `1.11.5`

mcnuttandrew:
Woot! Big thank! You continue to be wonderful

bodyflex:
amazing, thanks guys!

PR #1020, opened 2018-10-24
Added visual regression testing against Storybook

jayphelps:
Hey folks! Since you have a Storybook I figured you might want to have some visual regression testing against it, so Applitools was kind enough to sponsor me to help out and make some PRs.

If you're not familiar with automated visual regression testing there's a pretty good summary [in the Storybook docs](https://github.com/storybooks/storybook/blob/master/docs/src/pages/testing/automated-visual-testing/index.md) but the tl;dr is that it involves comparing the pixels visually like a human would, ideally ignoring differences that a human couldn't perceive e.g. 1 pixel changes, differences in how browsers render images, drop shadows, etc. to avoid false positives; but you can still do a strict 1:1 pixel match if that's your preference.

Applitools thankfully [provides an SDK](https://applitools.com/tutorials/storybook-react.html) that made it trivial to test Storybooks. So other than the things I included what you'll need to do is [create your own Applitools account](https://applitools.com/users/register) and [then setup your GitHub CI integration](https://applitools.com/docs/topics/integrations/github-integration.html) if you want it to run automatically on Pull Requests; optional, but probably a good idea. Happy to help with this as well, though someone with admin access to this repo will have to drive cause it requires that permission to setup.

While the default Applitools account has decent limits they also offer free unlimited accounts to open source projects, so if you want to unlock that you can either [contact them directly](https://applitools.com/contact/sales/open-source) or if you shoot me your account email I can ask them to hook you up.

***

You'll notice one thing I changed is that I added a random number generator that uses a stable seed value so that CI runs always produce the same random data because it sets `STORYBOOK_CHANCE_SEED=1`

image

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -9,18 +9,21 @@
     "build": "npm run clean && ocular build && npm run build-storybook",
     "build-storybook": "build-storybook -c .storybook -o ./dist/storybook",
     "lint": "ocular lint",
-    "storybook": "start-storybook -p 9001 -c .storybook"
+    "storybook": "start-storybook -p 9001 -c .storybook",
+    "test:visual": "STORYBOOK_CHANCE_SEED=1 npx eyes-storybook"
   },
   "keywords": [],
   "author": "",
   "license": "ISC",
   "devDependencies": {
+    "@applitools/eyes.storybook": "^2.1.5",
     "@storybook/addon-knobs": "^3.4.0",
     "@storybook/react": "^3.4.0",
     "ocular": "^0.4.8",
     "storybook-addon-jsx": "^5.3.0"
   },
   "dependencies": {
+    "chance": "^1.0.16",
     "d3-force": "^1.0.6",
     "d3-random": "^1.1.0",
     "react-virtualized": "^9.18.5",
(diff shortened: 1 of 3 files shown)


jayphelps:
Hey everyone! Any thoughts? 🕺

mcnuttandrew:
Hey @jayphelps 

Thanks for making this PR. While regression testing is a cool and important topic, I'm hesitant to introduce a new CI tool, especially a really new one like Applitools. I'd rather not leave you hanging, so I am going to reject this PR. I would be open to a future PR that adds regression testing (both for the main testing components and the showcase), but from a more commonly used source, eg puppeteer + a contemporary phantomcss.

jayphelps:
@mcnuttandrew Thanks! Very much appreciate the clarification. Applitools isn't new, though their automatic support for Storybook is. The goal was to add some visual regression tests that would have minimal maintenance burden but a large positive return. Since your Storybook is already there and kept up to date, it provides a great bang for your buck. Applitools also provides much smarter comparison than simple 1:1 pixel matching that PhantomCSS (no longer maintained) and similar tools to minimize false positives.

Thanks for your review and cheers! 🕺

PR #1024, opened 2018-10-25
Add HeatmapSeriesCanvas

nbardy:
Implemented HeatMapSeriesCanvas renderer.

NOTE:

I can across a bug when swapping over to this using where when the data is updated the CanvasWrapper gets the old props for the first call and then the next update it will receive the new props. Seems to be a canvas wrapper bug, but I wasn't able to track it down yet.

### DIFF
--- a/src/plot/series/heatmap-series-canvas.js
+++ b/src/plot/series/heatmap-series-canvas.js
@@ -1,0 +1,82 @@
+import {AbstractSeries, ScaleUtils} from 'react-vis';
+
+import {rgb} from 'd3-color';
+
+const getAttributeFunctor = ScaleUtils.getAttributeFunctor;
+const getScaleObjectFromProps = ScaleUtils.getScaleObjectFromProps;
+
+const DEFAULT_OPACITY = 1;
+
+function getScaleDistance(props, attr) {
+  const scaleObject = getScaleObjectFromProps(props, attr);
+  return scaleObject ? scaleObject.distance : 0;
+}
+
+export class HeatmapSeriesCanvas extends AbstractSeries {
+  static get requiresSVG() {
+    return false;
+  }
+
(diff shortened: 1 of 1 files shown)


mcnuttandrew:
Hey @nbardy 

This is an interesting idea! I like what you have so far. Would you mind adding a showcase example so that the differences can be checked? (see this file for example https://github.com/uber/react-vis/blob/master/showcase/plot/bar-chart.js) It can just be one of the old heatmap series examples but with a button to switch on and off from canvas mode.

I think we should address that canvas old props bug in another PR, mind filing an issue about it?

mcnuttandrew:
Oh also, if you rebase the build will probably fix it self

kobermeit:
I was looking to see if I could implement heatmap on top of a candlestick chart but it looks like we can't yet in the live code. Is this feature likely to get merged in?

I'm trying to make my own heatmap (like bookmap.com, tensorcharts.com) where the heatmap is displayed in a candlestick chart, but I'm having trouble with it so far. Any demos or help here?

mcnuttandrew:
@kobermeit I think you can probably do that with just the pure svg heatmap! That said if you are really dying to try out the canvas heatmap you can always copy the file used here copy the file and use it directly. Alternatively you can probably get away with just using the RectSeriesCanvas

kobermeit:
Thanks @mcnuttandrew , after experimenting with heatmap series in various chat libraries for a day, I decided it probably is best to do as you suggested - use an svg (or canvas) heatmap. What I'm struggling with now is getting the X Y coordinates of values in the candlestick chart so I can place the heatmap values in the correct spot (I'm trying to achieve something like tensorcharts, or bookmap.com). Is there a function I can use to get the coordinates by passing in the XY value?

Issue #1009, opened 2018-10-15
RadarChart - add straight "spider" lines (similar to CircularGridLines)


emilyjeppson:
Right now, it's possible to create a "spider" chart by using `RadarChart` and adding invalid data. See the "Radar Chart with Tooltips" showcase for an example:

<img width="263" alt="screen shot 2018-10-15 at 2 28 49 pm" src="https://user-images.githubusercontent.com/3814546/46976411-af4c2c00-d086-11e8-8908-3e298f9014ed.png">

In this example, the chart uses 6 dummy data entries to draw the 6 polygons before adding the real polygon. This is a little clunky.

It would be nice to have something like `CircularGridLines` (maybe `StraightCircularGridLines`?) to optionally add to the `RadarChart`. Consider the following example: a radar chart has 5 domains, each have a range of 0-5. The data you want to represent has a perfect 5 in all 5 domains. The `CircularGridLines` would make it look like the data is lacking - not a perfect 5/5/5 - due to the space between the polygon's straight lines and the rounded grid lines. Straight lines would more accurately depict the data.

mcnuttandrew:
@emilyjeppson feel free to take a crack at this if you like

emilyjeppson:
@mcnuttandrew Could I get your thoughts on the best way to implement this?

I figured there are 2 options: creating a new component (similar to `CircularGridLines`), or add a prop to `RadarChart` that draws the polygons. To me it seems best to create a new component, so that's what I've started trying. However, it'll be a bit more complicated than `CircularGridLines`. So I'm wondering if I'm going down the right path, and if so, if you have any recommendations on how to implement the component? (I'm currently looking at including the `polygon-series`, but I don't know if that's right)

emilyjeppson:
@mcnuttandrew no worries! I think I was over complicating it (I was getting `xFunctor` errors I didn't know what to do with). I have a potential solution for your review: https://github.com/uber/react-vis/pull/1028

PR #1028, opened 2018-11-01
Add RadarChartStraightGridLines - aka spider chart

emilyjeppson:
This is a potential solution for this: https://github.com/uber/react-vis/issues/1009

Basically, I wanted a solution that makes it easier to style a `RadarChart` as a "spider chart", similar to `CircularGridLines`, but with straight lines. I called it `RadarChartStraightGridLines` since this is specific to the `RadarChart`, but I don't really like that name, I'm definitely open to suggestions. Other notes:

- Although `line-series` could have been used, I used `polygon-series` so that the user can have more control over styling. Specifically, the polygon series allows a fill color for the polygons. 
- `radar-chart-straight-grid-lines` has it's own `XYPlot`. At first I tried it without this, but was getting errors about `xFunctor`. Maybe there's a better way to do that?
- Because the `XYPlot` is used, the `RadarChartStraightGridLines` needs to have a negative top margin to push it below the `RadarChart`. Additionally, a negative `z-index` is used to put it below the `RadarChart`.
- The `RadarChart` and `RadarChartStraightGridLines` need to have the same basic styling - height, width, starting angle, etc. Although it's possible for them to have different styling, that will make it the angles of the grid lines be different from the radar chart.
- The updated example looks like this:

screen shot 2018-11-01 at 11 40 29 am

- As an example, this is what it would look like when the domains have different ranges (this isn't committed to the examples):

screen shot 2018-11-01 at 11 10 29 am

Lastly, I haven't added this to the documentation, but I'd be glad to do so if you think this looks good.

### DIFF
--- a/showcase/radar-chart/radar-chart-with-tooltips.js
+++ b/showcase/radar-chart/radar-chart-with-tooltips.js
@@ -20,87 +20,31 @@
 
 import React, {Component} from 'react';
 import RadarChart from 'radar-chart';
+import RadarChartStraightGridLines from 'radar-chart/radar-chart-straight-grid-lines';
 import {Hint} from 'index';
 
-// The first 6 data elements here are to simulate a 'spider' type of radar chart -
-// similar to CircularGridLines, but straight edges instead.
 const DATA = [
-  {
-    name: 'Spider5',
-    mileage: 5,
-    price: 5,
-    safety: 5,
-    performance: 5,
-    interior: 5,
-    warranty: 5,
-    fill: '#f8f8f8',
-    stroke: '#cccccc'
-  },
-  {
-    name: 'Spider4',
-    mileage: 4,
-    price: 4,
-    safety: 4,
-    performance: 4,
-    interior: 4,
-    warranty: 4,
-    fill: 'white',
-    stroke: '#cccccc'
-  },
-  {
-    name: 'Spider3',
-    mileage: 3,
-    price: 3,
-    safety: 3,
-    performance: 3,
-    interior: 3,
-    warranty: 3,
-    fill: '#f8f8f8',
-    stroke: '#cccccc'
-  },
-  {
-    name: 'Spider2',
-    mileage: 2,
-    price: 2,
-    safety: 2,
-    performance: 2,
-    interior: 2,
-    warranty: 2,
-    fill: 'white',
-    stroke: '#cccccc'
-  },
-  {
-    name: 'Spider1',
-    mileage: 1,
-    price: 1,
-    safety: 1,
-    performance: 1,
-    interior: 1,
-    warranty: 1,
-    fill: '#f8f8f8',
-    stroke: '#cccccc'
-  },
-  {
-    name: 'Spider0',
-    mileage: 0.1,
-    price: 0.1,
-    safety: 0.1,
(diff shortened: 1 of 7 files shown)


src/plot/xy-plot.js
```diff
@@ -545,11 +545,13 @@ class XYPlot extends React.Component {
...
     }
     const components = this._getClonedChildComponents();
+    const xyPlotStyle = style ? style.XYPlot : {};
     return (
       <div
         style={{
           width: `${width}px`,
-          height: `${height}px`
+          height: `${height}px`,
+          ...xyPlotStyle
```
emilyjeppson: This was needed in order to pass in the negative `z-index`. The "empty" xy-plot puts the `...style` directly on the div. That could be done here, but I thought that could be dangerous, so I created a new style specifically for this div.

src/radar-chart/radar-chart-straight-grid-lines.js
```diff
@@ -0,0 +1,166 @@
...
+
+  return (
+    <XYPlot
+      height={height}
+      width={width}
+      margin={margin}
+      style={{
+        ...style,
+        XYPlot: {
+          zIndex: -1
```
emilyjeppson: Here's where that negative `z-index` is passed in.

showcase/radar-chart/radar-chart-with-tooltips.js
```diff
@@ -178,15 +123,28 @@ export default class RadarChartWithTooltips extends Component {
...
+          </Hint>
+        )}
+        <RadarChartStraightGridLines
+          numberOfGridlines={5}
+          numberOfDomains={6}
+          width={450}
+          height={350}
+          margin={{top: 30, right: 60, left: 60, bottom: 30}}
+          style={{
+            marginTop: '-350px',
```
emilyjeppson: This margin always needs to be a negative value of the height of the `RadarChart`. If approved, I would specifically call this out in the documentation.

showcase/radar-chart/radar-chart-with-tooltips.js
```diff
@@ -178,15 +121,28 @@ export default class RadarChartWithTooltips extends Component {
...
+        {hoveredCell && (
+          <Hint value={hoveredCell}>
+            <div style={tipStyle}>
+              {hoveredCell.domain}: {hoveredCell.value}
+            </div>
+          </Hint>
+        )}
+        <RadarChartStraightGridLines
+          numberOfGridlines={5}
+          numberOfDomains={6}
```
mcnuttandrew: numberOfDomains should probably be inferred by the component?

showcase/radar-chart/radar-chart-with-tooltips.js
```diff
@@ -178,15 +121,28 @@ export default class RadarChartWithTooltips extends Component {
...
-            </Hint>
-          )}
+        {hoveredCell && (
+          <Hint value={hoveredCell}>
+            <div style={tipStyle}>
+              {hoveredCell.domain}: {hoveredCell.value}
+            </div>
+          </Hint>
+        )}
+        <RadarChartStraightGridLines
```
mcnuttandrew: Maybe just RadarGridLines?

Issue #1033, opened 2018-11-08
Difference in rendering LabelSeries on firefox and Chrome


toddgardner:
There's a different rendering of Label Series on Firefox vs Chrome; I think this is because it uses alignment-baseline instead of dominant-baseline. I've included my motivating example below, but this is also reflected in the Label Series sample, will attach to the PR.

Given code like:

```
const score = 81;
const accuracyData = [
    {
      angle: score,
      color: "#ff8b52"
    },
    {
      angle: 100-score,
      color: "white"
    },
];

const labelData = [
    {
      x: 0,
      y: 0,
      xOffset: 13,
      yOffset: 13,
      label: score.toString(),
    }
];

return (
    <RadialChart
      innerRadius={25}
      radius={30}
      data={accuracyData}
      colorType="literal"
      width={70}
      height={70} >
      <LabelSeries data={labelData} />
    </RadialChart>
);
```

It renders differently on chrome and firefox:

Mac: 10.12.6
Chrome 69.0.3497.100:

![image](https://user-images.githubusercontent.com/7430154/48095749-82052080-e1e3-11e8-91ff-ebaa48b325a2.png)

Firefox 63.0.1:

![image](https://user-images.githubusercontent.com/7430154/48095759-86c9d480-e1e3-11e8-919e-c634986510ba.png)

Digging into this, it appears due to the `alignment-baseline` property used by LabelSeries; this is meant to be used relatively to a parent textual element but Chrome apparently accepts it without a parent, while Firefox does not. The usage of `dominant-baseline` is probably more correct and works on Firefox (though it appears you have to specify both if you want to change them).

Here's a workaround that renders the same on both firefox and chrome:

```
  const labelData = [
    {
      x: 0,
      y: 0,
      xOffset: 0,
      yOffset: 0,
      label: score.toString(),
      style: {
        dominantBaseline: "central",
        alignmentBaseline: "central",
        textAnchor: "middle"
      }
    }
  ];
```

Though changing the Label Series to use dominant is a more permanent fix (PR incoming)

agwells:
Thanks for the explanation! I encountered this same issue, and was very confused by it.

My workaround was to set `labelAnchorY="auto"` on the `LabelSeries` element, basically turning off this behavior in FF _and_ Chromium, and then fine-tune my labels' vertical positioning with the data `yOffset` field.

PR #1034, opened 2018-11-08
Use dominant baseline instead of alignment for consistent rendering

toddgardner:
This is a fix for https://github.com/uber/react-vis/issues/1033

AFAICT (and I'm by no means an SVG expert) alignment-baseline is only supposed to be meaningful relative to a textual parent like a tspan, though chrome accepts it; dominant baseline is the correct attribute to use in this case, as it renders irrespective of the parent.

Alignment Baseline: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/alignment-baseline
Dominant Baseline: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/dominant-baseline

Pre-firefox:

image

Pre-chrome:

image

Post-firefox:

image

Post-chrome:

image

### DIFF
--- a/docs/label-series.md
+++ b/docs/label-series.md
@@ -102,7 +102,7 @@ This attribute is used to align (start-, middle- or end-alignment) the label tex
 
 Type: `string`
 
-This attribute is used to align the label text vertically relative to the datapoint. (Sets the alignment-baseline attribute for the element https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/alignment-baseline)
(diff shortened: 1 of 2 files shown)


mcnuttandrew:
Wow, this is a fascinating catch. I guess i'm not surprised by chrome/FF incongruities, but jeez that's a subtle one. I super approve.

One caveat to this boundless approval: in your bug report you note that you need to set both alignment and dominant baselines in order to override. If users want to change the alignment using the style override (which i find myself doing sometimes) will the default values get in the way?

toddgardner:
I had to override both because the react-vis was specifying `alignment` and chrome prefers alignment to the overridden `dominant`.

Post-PR, if you wish to override, then overriding just `dominant` should be sufficient.

mcnuttandrew:
Cool, well that's good enough for me. Stamp stamp stamp

PR #1051, opened 2018-11-23
Update CustomSVGSeries to use getX / getY

SinZ163:
I ran into this problem when I refactored my graphs to use getX / getY rather than previously data.map to conform to x and y keys, and in doing so made my icons disappear as they had NaN as their x and y.

This has been successful in fixing them.


SinZ163:
Sorry for the delays, updated an example now

Issue #1049, opened 2018-11-22
Better error message on accidental import of wrong series


CAYdenberg:
I mistakenly imported BarSeries instead of VerticalBarSeries in my component file. Easy mistake to make.

Unfortunately the error message is 
`Cannot read property 'prototype' of undefined`

which occurs when react-vis is checking the children of XYPlot to see if they are a series or not. It would be cool if it could warn you you've done something terrible instead.

I'd be happy to contribute if you guys will take it.

mcnuttandrew:
Good idea! I would be delighted to see a PR to this effect.

Pardeep94:
May i know what is the problem with BarSeries?
I also use BarSeries and get same error.
when i am using VerticalBarSeries than it is working fine

CAYdenberg:
Yeah, the error is that there **is** no component called `BarSeries` in react-vis ... there is however a `VerticalBarSeries` and a `HorizontalBarSeries`.

It's a somewhat easy mistake to make because of the examples in the documentation.

Pardeep94:
Ok, Thank you

PR #1053, opened 2018-11-23
Throw explicit error when passing a child component that cannot be ch…

CAYdenberg:
…ecked by isSeriesChild

Not sure if you like the test of the full component as it posts a console warning during render.


CAYdenberg:
Closes #1049

mcnuttandrew:
Thanks for doing this! I'm not sure this will solve the bug you highlight in #1049 though. Would you mind writing a test to verify?

CAYdenberg:
This test is intended to mimic that exact case: https://github.com/uber/react-vis/pull/1053/files#diff-4f016ba23b1ae89c71289fa956c002caR252

If I import a component that doesn't exist (`import {BarPlot} from 'react-vis'} that component is `undefined` ... and that test fails if the code I added is not there.

PR #1059, opened 2018-12-03
Fix(Deps): Security Warning

benshope:
Lightly bumps packages with security warnings to version to remove this warning:

screen shot 2018-12-03 at 3 24 34 pm


benshope:
Nevermind, looked a bit deeper at something like the randomatic issue and it seems like these alerts don't need to be addressed at all

PR #1078, opened 2018-12-31
Add hovering example for heatmap series

mcnuttandrew:


showcase/plot/heatmap-chart.js
```diff
@@ -18,32 +18,42 @@
...
-          {x: 2, y: 15, color: 12},
-          {x: 3, y: 0, color: 9},
-          {x: 3, y: 5, color: 2},
-          {x: 3, y: 10, color: 6},
-          {x: 3, y: 15, color: 12}
-        ]}
-      />
-    </XYPlot>
-  );
+export default class HeatmapChart extends Component {
```
benshope: Can't wait until we're able to use hooks instead of making a class

showcase/plot/heatmap-chart.js
```diff
@@ -18,32 +18,42 @@
...
+            {x: 2, y: 5, color: 2},
+            {x: 2, y: 10, color: 1},
+            {x: 2, y: 15, color: 12},
+            {x: 3, y: 0, color: 9},
+            {x: 3, y: 5, color: 2},
+            {x: 3, y: 10, color: 6},
+            {x: 3, y: 15, color: 12}
+          ]}
+        />
+        {value && <Hint value={value} />}
```
benshope: `value !== false` if `value` is a number since it could be zero

showcase/plot/heatmap-chart.js
mcnuttandrew: 👍👍👍👍👍👍👍👍👍👍👍

PR #1094, opened 2019-01-23
docs: simplify usage a bit

Haroenv:
A Twitter thread showed this wording, and the consensus there was that this was an overly complicated way of phrasing. I simplified that text a bit, it seems just as clear to me (even after looking up what "to wit" means).

### DIFF
--- a/docs/xy-plot.md
+++ b/docs/xy-plot.md
@@ -92,7 +92,7 @@ Not all properties can be visualized in each series. Here's a short comparison o
 
 ### A note on ordering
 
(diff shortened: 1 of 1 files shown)


mcnuttandrew:
I tend to like turns of phrase like "to wit"  that allow for the author to declare that they will then be giving an example. That said, I'd rather have the docs be readable by everybody than fun to read by some, so stamp stamp stamp. Thanks!

Haroenv:
Perfect!

PR #1114, opened 2019-02-27
Avoid registering onWheel listener when not needed in XYPlot

lfbayer:
This pull request allows the onWheel handler to be undefined to avoid registering an active listener when not needed.

In the case that the XYPlot is used without an onWheel callback the code was registering a listener anyway. This would cause warnings in the chrome console about registering an active listener.

The warning will still be shown in the case that an onWheel callback is set, but this change prevents the listener from being registered when it is a no-op anyway.

### DIFF
--- a/src/plot/xy-plot.js
+++ b/src/plot/xy-plot.js
@@ -485,18 +485,6 @@ class XYPlot extends React.Component {
     });
   };
 
-  /**
-   * Trigger doule-click related callbacks if they are available.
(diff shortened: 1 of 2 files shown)


mcnuttandrew:
Seems fine! I feel like similar work should maybe be done to all the listeners, which have always seemed really clumsy to me. This is great, thanks for being thorough and writing tests.

lfbayer:
Thanks!

PR #1133, opened 2019-03-24
fix typo

foobarbecue:


mcnuttandrew:
Thanks! We should really run write-good and a spell checker as part of the docs build process

Issue #940, opened 2018-08-28
Windows build issues


WildHorse19:
Hey guys, i'm experiencing some minor problems building on windows.
1)"npm run build" produces:
` "'rm' is not recognized as an internal or external command"`
2)"npm run test" produces 
```
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
SyntaxError: missing ) after argument list
(i changed it to "babel-node --inspect ./tests/index.js")
```
3)"npm run lint" generates line-breaks errors (see screenshot) 
![lint-error](https://user-images.githubusercontent.com/41179841/44720399-612b9b00-aacf-11e8-9770-82c7a2c77f9e.jpg)
Adding "linebreak-style": ["error", "windows"] to "rules" object in both .eslintrc files fixing it but build will fail on travis with previously mentioned error (line-breaks).
4)Running "npm run build" inside showcase folder produces:
`'NODE_ENV' is not recognized as an internal or external command,`
(i changed it to "set NODE_ENV=production&&.\\node_modules\\.bin\\webpack")
5)I'm unable to build showcase on windows, for some reason paths resolution doesnt work, 
![before](https://user-images.githubusercontent.com/41179841/44720664-1e1df780-aad0-11e8-937f-3cbfda2f68f4.jpg)

i've tried to create alias inside webpack config which will point to outer index.js 
```
 resolve: {
    alias: {
      // 'index':path.join(__dirname,'..', 'src', 'index.js'),
      // 'theme':path.join(__dirname,'..', 'src', 'theme.js')
    }
  },
```
but now it generates different list of errors, 
![after](https://user-images.githubusercontent.com/41179841/44720668-2413d880-aad0-11e8-81fc-b75466f1931b.jpg)

I have been able to fix all problems except last one, maybe somebody knows some quick fix, otherwise i will have to create alias for each file in src. Thank you!

PR #1144, opened 2019-04-05
fix webpack windows build issue(#940)

talc23:
Fixing the build issue on windows, added only to the development webpack


mcnuttandrew:
I checked it out on my local, and it seems good! Thanks for doing this!

PR #1149, opened 2019-04-08
fix: Fix FlexiblePlot handling of container resize

WaldoJeffers:
## Tackled issue
A flexible plot is not properly resized when its parent container is resized

## Cause
The `_onResize` is called on the `componentWillReceiveProps` lifecycle event, which in addition to [being deprecated](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops), is the wrong event: computing the parent dimensions will yield incorrect results (the DOM hasn't been updated yet)

## Solution
Use [`componentDidUpdate`] instead, and make sure `onResize` only calls `setState` if something did change.

### DIFF
--- a/src/make-vis-flexible.js
+++ b/src/make-vis-flexible.js
@@ -122,17 +122,17 @@ function makeFlexible(Component, isWidthFlexible, isHeightFlexible) {
     _onResize = () => {
       const containerElement = getDOMNode(this[CONTAINER_REF]);
       const {offsetHeight, offsetWidth} = containerElement;
+      const {height, width} = this.state;
 
-      const newHeight =
-        this.state.height === offsetHeight ? {} : {height: offsetHeight};
-
-      const newWidth =
-        this.state.width === offsetWidth ? {} : {width: offsetWidth};
-
-      this.setState({
-        ...newHeight,
-        ...newWidth
-      });
+      if (height !== offsetHeight || width !== offsetWidth){
+        // Only call setState if one dimension has changed.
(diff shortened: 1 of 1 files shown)


ronocod:
Will this fix https://github.com/uber/react-vis/issues/1159?

abramjstamper:
I'm currently having an issue when using [Bulma's](https://bulma.io/) columns, which are based on flexbox.   It seems to be related to issue #1159. Anyway we could get this merged in when you get a chance? @mcnuttandrew

dcosson:
Any way to get this merged?

mcnuttandrew:
It's been a little bit, but I'll pass this if you provide a gif demonstrating before and after behaviour

zebralight:
> Will this fix #1159?

it does not. I just tried it.

zebralight:
> It's been a little bit, but I'll pass this if you provide a gif demonstrating before and after behaviour

If you can provide a code sample to reproduce the issue along with the gif, that would be great.

mcnuttandrew:
I had assumed that others asking for this be merged indicated that it did the job, sorry for being a lazy reviewer. Bummer to hear that it doesn't.

abramjstamper:
I had commented prematurely as well back in May when asking for this to be merged. Back in June, I pulled from git and built the source. I installed the version I built and still was experiencing issues. I looked into #1159 before since I've experienced it myself, but I'm not sure what was causing the re-scaling issue. I think there's a code sample in #1159 demonstrating the issue ([demo](https://codesandbox.io/s/wk6kpxl3jk)).

Issue #1004, opened 2018-10-11
Unclickable values in separated LineMarkSeries


dbertella:
I have different line mark series displaying on the same chart and I need to hover over the values and being able to select the point.

I have this problem when the value are close and I discover there is path covering my values.

setting the css belove solve the issue, I just want to know if it can create a problem somewhere else.
```
.rv-xy-plot__series {
   pointer-events: auto; /* this is all by default */
}
```

mcnuttandrew:
Using point-events manipulation is a totally reasonably approach to your problem. I tend also really like using voronois for complex mouse over interactions, it's a little less plug and play but it makes for way better user interactions. The TLDR is that you should assemble a voronoi of nodes that to mouse over by combining the two sets of marks of interest.

sasuomi:
@dbertella @mcnuttandrew FYI, check out my PR: https://github.com/uber/react-vis/pull/1152

If you want the fix without using the forked package, just add:

```jsx
lineStyle: {
  pointerEvents: 'none'
}
```

to each of your `LineMarkSeries` props.

Issue #1099, opened 2019-02-05
onValueMouseOver event in LineMarkSeries component cannot be triggered after adding Highlight component


renhax:
After adding Highlight component, there is a "rv-mouse-target" div created by Highlight component, all events from LineMarkSeries cannot be triggered. If I change its "pointer-events" css to None, LineMarkSeries back to working, but Highlight stops working

nehruji:
I have the same issue. When I try to use both either one is working at a time. Along with this onValueMouseOver has one more issue like when I have LineMarkSeries implemented, onValueMouseOver callback is not getting triggered for few LineMarkSeries. Don't know in what basis it is happening.

sasuomi:
@renhax did you try flipping the rendering order, i.e. render `<Highlight />` before all the `<LineMarkSeries />`s ?

@nehruji , check out my PR: https://github.com/uber/react-vis/pull/1152

If you want the fix without using the forked package, just add:

```jsx
lineStyle: {
  pointerEvents: 'none'
}
```

to each of your `LineMarkSeries` props.

radove:
This seems to happen with other situations too, if I order the <Highlight /> in front of <LineMarkSeries /> or <MarkSeries />, it blocks access to pointer events.  If I order <Highlight /> to draw behind LineMarkSeries / MarkSeries with pointer events, it is hard to get the ability to draw the Highlight.  Additionally, if I simply put pointerEvents: 'none' -- it is helpful, but if you were to draw your Highlight square, and run the mouse into an pointer event of a LineMarkSeries/MarkSeries, it causes the Highlight square to pre-maturely stop drawing.

PR #1152, opened 2019-04-10
Fix LineMarkSeries blocking pointer events on other LineMarkSeries

sasuomi:
Fixes https://github.com/uber/react-vis/issues/1004 (and probably https://github.com/uber/react-vis/issues/1099#issuecomment-463553067).

## Issue

Having multiple `LineMarkSeries` overlapping in a single plot may result in some of the marks not firing the `onValueMouseOver` callback.

This is because the `LineMarkSeries` draws an invisible area with the `<plot>` that is blocking pointer events on marks rendered behind it. I've set `lineStyle={{ fill: "#80808080" }}` in my codesandboxes to highlight this otherwise invisible area. 

### Before

https://codesandbox.io/s/43q7xyq3vw

before

### After

https://codesandbox.io/s/xp3w5kly1p

after


sasuomi:
Now that I thought about this more, shouldn't this bug also appear when using just `LineSeries`? I'm not sure how to fix the issue in that case. For ´LineMarkSeries`, fixing it like this seems like a reasonable approach for now but let me know if you have other ideas.

bmaupin:
This should fix #1157 as well

danchyy:
Will this Pull Request get merged anytime soon?

bmaupin:
> Will this Pull Request get merged anytime soon?

My guess is the answer is no. This repository appears to be unmaintained as per https://github.com/uber/react-vis/issues/1303

Thankfully the workaround is pretty straightforward: https://github.com/uber/react-vis/issues/1004#issuecomment-481666020

danchyy:
Thanks for reply @bmaupin . Could you help me out just a bit more please, should that CSS you mentioned in #1004 be assigned to `style` variable of LineMarkSeries? 

Like this?

```
<LineMarkSeries
   data={[...]}
   style={{
         lineStyle: {
              pointerEvents: 'none'
         }
   }}/>
```

Hivemind9000:
> Thanks for reply @bmaupin . Could you help me out just a bit more please, should that CSS you mentioned in #1004 be assigned to `style` variable of LineMarkSeries?
> 
> Like this?
> 
> ```
> <LineMarkSeries
>    data={[...]}
>    style={{
>          lineStyle: {
>               pointerEvents: 'none'
>          }
>    }}/>
> ```

You probably figured it out already, but for any future readers of this issue:

lineStyle is a property of LineMarkSeries (and probably LineSeries) though it doesn't seem to be in the documentation.

```
<LineMarkSeries
   data={[...]}
   lineStyle: {
       pointerEvents: 'none'
   }
 />
```

There's also a markStyle property you can set, if needed.

PR #1219, opened 2019-07-18
Fix typo which causes ChartLabel to misalign

richardlewisjones:


mcnuttandrew:
Doh, thanks for catching that

PR #1224, opened 2019-07-25
Add 'paddingDirections' property to the treemap

Bi0max:
We needed to have padding only on the top of the treemap, but it was not possible.
It is useful, when one wants the name of the parent to be seen above the children.
Currently, padding applies from all sides and makes all children components significantly smaller.

So, I decided to add a `paddingDirections` property, so that user can himself choose, where to apply padding.

Before (with padding=20):
image

After (with padding=20 and paddingDirections=['top']):
image

### DIFF
--- a/docs/treemap.md
+++ b/docs/treemap.md
@@ -98,7 +98,14 @@ Height of the component.
 
 Type: `number`
 
-The padding between cells the cells of the heatmap in pixels.
+The padding between the cells of the heatmap in pixels.
+
+#### paddingDirections (optional)
+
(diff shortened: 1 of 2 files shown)


Bi0max:
It seems, that build is failed not because of my changes to the code.

mcnuttandrew:
try rebasing off master i think the build should be fixed now

Bi0max:
Thank you for letting me know! build works now.

Bi0max:
Hi guys, do you plan to merge this pull request?

PR #1242, opened 2019-09-03
Exclude strings representing falsy values in generated classnames

### LINKED ISSUE #1186: Bug component Hint className property if not set outputs `undefined` in class v1.11.7
I believe in hint.js this is the line where the behavior occurs.

https://github.com/uber/react-vis/blob/master/src/plot/hint.js#L366
image

kkowalczuk:
With current approach of generating class names with just putting user-defined class names in the template literal, interpolated string will contain all values of custom signatures - even the ones that are not meant to be stringified (e.g. `undefined`).

Instead of re-inventing the wheel and implementing own version of helper that eliminates all falsy values from the interpolated string, I decided to include [classcat](https://github.com/jorgebucaran/classcat) library. It has no external dependencies, it's faster than alternatives like [classnames](https://github.com/JedWatson/classnames) (see [benchmark results](https://github.com/jorgebucaran/classcat#run-the-benchmarks)) and it only exposes one function that fully covers our needs (gzipped weights [~300B](https://bundlephobia.com/result?p=classcat@4.0.2)).

Fixes #1186


kkowalczuk:
@mcnuttandrew please check the CI environment config; few last consecutive builds have failed not neccesarily due to broken implementation ;)

kkowalczuk:
@mcnuttandrew:

> I am pretty strongly against adding a dependency

Got it, external dependency has been removed 👍 

> I believe that class-naming as it stand is not a bottle neck for react-vis (it appears to be more closely related to the process of putting stuff on screen).

Yeah, just to be crystal clear - my contribution does not address any performance bottlenecks. The class name interpolation (that affects "process of putting stuff on screen") though was done in a _wrong_ way, resulting in below HTML element in the DOM tree (notice the `class` attribute - inspect [showcase ](https://uber.github.io/react-vis/examples/showcases/plots) from official docs :) ):

```html
<g class="rv-xy-plot__series rv-xy-plot__series--mark undefined" transform="translate(10,10)">
  ...
</g>
```

which is completely off. Imagine that the final user has below declaration somewhere in his/her global CSS stylesheet (CSS class `.undefined` is very unlikely scenario, but not impossible):

```css
.undefined {
  display: none; // may be whatever changing the visual output
}
```

What happens? The user did not add `.undefined` to the XY plot intentionally, but it still got appended with no reason.

Anyway, PR updated. Looking forward for your feedback.

mcnuttandrew:
This is going in a good direction! 

> Yeah, just to be crystal clear - my contribution does not address any performance bottlenecks.

Sorry for the oververbose explanation re: perf bottle necks. Sometimes folks argue really hard for libraries bc of features like that and I wanted to be really clear why I was rejecting that library. I'm totally onboard with preventing undefineds from getting in the class list. 

I like your test for your utility! However, the test I was asking for would test that the behaviour documented in the #1186. So this would be a test that before your addition would fail because it would detect undefined as a class name on some test-rendered component (see any of the tests/components/ files for examples), and then after the change it would pass. This is to prevent future / forgetful maintainers from rebreaking this issue. Thanks again for doing this and being so responsive, I'll see about unbreaking the build

mcnuttandrew:
Build should be fixed now, try doing a rebase off master

kkowalczuk:
>the test I was asking for would test that the behaviour documented in the #1186.

Here you are ;)

Thanks for fixing the build, cheers!

mcnuttandrew:
Alright!! This is great. Thanks for doing and putting up with a picky review, you're da champ

Issue #1238, opened 2019-08-22
Highlight Y-Axis Accuracy is 


s-heap:
Using the highlight to add box dragging to your graph results in just inaccurate y-Axis area selection.

This can be seen in the documentation itself:
https://uber.github.io/react-vis/documentation/api-reference/brushing-and-dragging
![ReactVisBugPreZoom](https://user-images.githubusercontent.com/40580115/63506604-be351380-c4cd-11e9-9688-17824bf5b86c.png)

![ReactVisBugPostZoom](https://user-images.githubusercontent.com/40580115/63506624-c8571200-c4cd-11e9-8ea2-f109d6798e42.png)

As can be seen although I clearly selected a box between 0 and 20 on the y-axis, the recorded area is actually between -1.18 and 18.86. This difference has been causing me issues with higher fidelity data since I cannot accurately select the data I wish to view.

This lack of accuracy only occurs in the y-axis and in the x-axis the accuracy is far more acceptable.

s-heap:
I have created a pr for this issue, though It does not solve the problems I have found with the Voronoi component

PR #1243, opened 2019-09-04
Highlight component when dragging on X and Y not taking Y margin into account

s-heap:
Added "- marginTop" in the _convertAreaToCoordinates function when enableX and enableY are set to account for the vertical margin and prevent inaccuracies in y selection.

A pull request to fix the issue mentioned  in #1238 regarding the highlight component inaccuracies.


mcnuttandrew:
I think this was handled in another PR, try rebasing and see where you get

s-heap:
I've had a look but haven't been able to find a relevant pull request for this issue.
If you have any other information on how to find it that would be useful.

mcnuttandrew:
i 100% misremembered my bad, long week. the other pr resolved a similar error in a different place in the code

s-heap:
That's fair I understand. I've come across a few errors and issues that I think are linked to the system not taking margins properly into account but this is the only one I've found and nailed down as of yet.

Burbenog:
I there any chance this will be merged any time soon?

PR #1257, opened 2019-10-23
default to user supplied margins over defaults

exlexer:
Fix `RadialChart` margin prop to cascade to `ZYPlot` props.


mcnuttandrew:
👍👍👍

PR #1267, opened 2019-11-05
No need to define enabled legend item colour.

layik:
This is a minor contribution. Thinking in term of "themes" I think it is not necessary to define enabled items colour.


mcnuttandrew:
i think this would be a breaking change for anyone who relies upon this functionality currently. While the contribution is certainly appreciated, i don't think it is necessary at this time.

layik:
Care to say a little more than "think" that this would be breaking? I will see if I can send a PR that can expose the "title" colour then.

PR #1297, opened 2020-01-07
fix broken link in treemap docs

id0Sch:
first off, thanks for this awesome lib!

I was going through the Treemap docs and came across the broken link so I fixed it :) 
cheers! 🍻


mcnuttandrew:
👍👍👍

PR #1314, opened 2020-05-02
Cache Child Props

Xiot:
Cache the child props to prevent React from re-rendering unnessesarily.
The biggest culprits were the default accessors.  The defaults for each
attribute are now computed once, and re-used when required.


Xiot:
Using the 5th graph on the showcase (Line Chart Canvas) I was able to increase the fps while hovering over the graph from about 5 fps to 60 fps.

tests/utils/data-utils-tests.js
```diff
@@ -56,7 +56,7 @@ test('data-utils #addValueToArray', t => {
 test('data-utils #transformValueToString', t => {
   t.deepEqual(transformValueToString(0), 0,
     'Shouldn\'t transform the number value');
-  t.deepEqual(transformValueToString(new Date(0)), 'Thu Jan 01 1970',
+  t.deepEqual(transformValueToString(new Date(43200000)), 'Thu Jan 01 1970',
```
Xiot, line 59: This value is `12:00 PM` which should play nicer with timezone differences.

src/plot/xy-plot.js
```diff
@@ -187,6 +187,22 @@ class XYPlot extends React.Component {
     }
   };
 
+  _childPropCache = Object.create(null)
```
chrisirhc, line 190: I'm a bit hestiant to add a cache instead of relying on useMemo. How about we wait till the modernization to React 16.8 (#1316) and useMemo instead (after some refactoring of this prop cloning behavior perhaps to using context)?

Alternatively, if you'd like to try out what this looks like via the modernize branch. Feel free to go for it.

PR #1318, opened 2020-05-21
Fix relative imports in packages/react-vis/tests/ folder

a-peltier:
Fix relative imports in order to fix the `test` command for the react-vis package


Xiot:
looks great.  Can you quickly change the `engines` in the root `package.json` so that CI will run?
CI looks like it starts at Node 6, so `>=6.0.0` should work

packages/react-vis/package.json
```diff
@@ -32,12 +32,7 @@
     "full-test": "npm run lint && npm run cover",
     "cover": "nyc --reporter=text --reporter=html --reporter=lcov npm test",
     "prettier": "prettier --write $(git ls-files | grep '.js$')",
```
Xiot, line 34: extra comma at the end of this file

packages/react-vis/package.json
```diff
@@ -79,7 +74,6 @@
     "eslint-plugin-no-only-tests": "^2.0.1",
     "eslint-plugin-prettier": "^2.6.2",
     "eslint-plugin-react": "^7.11.1",
-    "husky": "^1.1.2",
```
Xiot, line 82: and we still need `huskey` listed as a dependency somewhere

a-peltier: yes I moved it to the root package.json file

a-peltier:
after some tweaking and updates with eslint: 

- I removed  `eslint-config-uber-es2015` `eslint-config-uber-jsx` because they don't seem to be maintained (baseweb don't use them so I figured it would be safe to remove too)
- I added `eslint:recommended` and `plugin:react/recommended`
- I deactivated the react/prop-types  rule because it was causing 120 errors like this :  `'FOO' is missing in props validation`

PR #1319, opened 2020-05-22
Get Storybook working.

Xiot:
Added the website to the list of packages, and got the storybook working.
Upgraded the required packages, and fix the storybooks to work with the new versions.

### DIFF
--- a/packages/website/.storybook/config.js
+++ b/packages/website/.storybook/config.js
@@ -1,4 +1,9 @@
-import {configure} from '@storybook/react';
+import {configure, addParameters, setAddon} from '@storybook/react';
+import JSXAddon, {jsxDecorator} from 'storybook-addon-jsx';
+import {SimpleChartWrapper, jsxOptions} from '../storybook/storybook-utils';
+
+setAddon(JSXAddon);
+addParameters({jsx: jsxOptions});
 
 function loadStories() {
   require('../storybook/index.js');
--- a/packages/website/package.json
+++ b/packages/website/package.json
@@ -1,5 +1,6 @@
 {
   "name": "react-vis-website",
+  "private": true,
   "version": "1.0.0",
   "description": "A composable charting library",
   "main": "index.js",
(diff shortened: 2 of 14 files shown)


Xiot:
I had to change how were were passing in the `jsxOptions` due to https://github.com/uber/react-vis/pull/1319.  There is a [PR](https://github.com/storybookjs/addon-jsx/pull/95) to fix it, but it hasn't been touched

packages/website/package.json
```diff
@@ -28,7 +28,7 @@
     "react-vis": "^1.11.2"
   },
   "peerDependencies": {
-    "react": "^16.0.0",
-    "react-dom": "^16.0.0"
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1"
```
chrisirhc, line 32: That's a pretty high version for peerDeps. Perhaps this should actually be deps or devDeps since this package isn't to be exported?

packages/website/.storybook/config.js
```diff
@@ -1,4 +1,9 @@
-import {configure} from '@storybook/react';
+import {configure, addParameters, setAddon} from '@storybook/react';
+import JSXAddon, {jsxDecorator} from 'storybook-addon-jsx';
+import {SimpleChartWrapper, jsxOptions} from '../storybook/storybook-utils';
+
+setAddon(JSXAddon);
+addParameters({jsx: jsxOptions});
```
chrisirhc, line 6: Nice. This is what removes the need for jsxOptiions?

Xiot: Ya.  There is a bug in `addWithJSX` (referenced above) that doesn't take the 3rd argument anymore.  There is an open PR to fix it from November, however it hasn't been touched.

Since all of the stories are using the exact same options, I figured that setting it globally was easier.  This way we can't forget to use it.

packages/website/package.json
Xiot: I can move it to `dependencies`

Xiot:
> Seems fine though CI seems to be failing due to node version.

Ya. the work that Alex is doing should resolve the build errors. Once he finishes up with the lint errors, I will rebase.

Xiot:
Interesting... the build is failing due to `fsevents` not being supported on `linux` but that package wasn't changed between the two commits.

PR #1322, opened 2020-05-23
Yarn workspace lint showcase

a-peltier:
I move all the eslint dependencies at the root in order to have the same config for every package. 

Since now the showcase parsed by eslint, I had to fix all the errors.


package.json
```diff
@@ -25,10 +25,18 @@
     "es6",
     "babel"
   ],
-  "scripts":{
+  "scripts": {
+    "lint": "eslint .",
```
Xiot, line 29: We should probably delegate the linting to a script on the package, but I agree that we should have a script at the root level, and that the config should be controlled at the root level.

At the minimum though, we should only run lint on the `packages/` folder and make sure to skip `node_modules` folders, and bundles.

```
eslint packages --ignore-pattern node_modules --ignore-pattern bundle.js
```

packages/showcase/axes/custom-axes.js
```diff
@@ -35,16 +35,17 @@ const WORDS = [
   'skateboard',
   'wow',
   'such',
+  // eslint-disable-next-line react/jsx-key
   <tspan>
-    <tspan x="0" dy="1em">
+    <tspan key="" mx="0" dy="1em">
```
Xiot, line 40: We should just be able to add a `key` to the `tspan` and get rid of the `eslint-disable`.
Also, why was `x` changed to `mx` ?

packages/showcase/color/mini-color-examples.js
```diff
@@ -312,7 +312,7 @@ function generateCharts(seriesData, props) {
       {[VerticalBarSeries, LineSeries, MarkSeries].map((Type, key) => (
         <XYPlot key={key} {...defaultXYPlotProps} {...props}>
           {seriesData.map((d, i) => (
-            <Type {...d} />
+            <Type key={i} {...d} />
```
Xiot, line 315: `key` _should_ be a string

package.json
a-peltier: You're right I will add an `.eslintignore` file to specify bundle.js

But node_modules are ignored by default, see https://eslint.org/docs/user-guide/configuring#eslintignore
> In addition to any patterns in a .eslintignore file, ESLint ignores files in /**/node_modules/* by default. It can still be added using !.

a-peltier: I have checked with `eslint . --debug` and I can confirm that node_modules folders were not parsed

packages/showcase/color/mini-color-examples.js
a-peltier: Fixed

packages/showcase/axes/custom-axes.js
a-peltier: Fixed, and `mx` was a typo, thanks

package.json
a-peltier: I added an .eslintignore file that allows the `lint` command to work even after building the `react-vis` package.

The showcase's build command fails on this branch so we should add the needed ignored files once the build is fixed

Xiot: nicely done

PR #1324, opened 2020-05-23
Use react@^16.8.3 as minimum

Xiot:
Update pre-commit script to run the remove-unpm-rfs script

### DIFF
--- a/packages/react-vis/package.json
+++ b/packages/react-vis/package.json
@@ -86,8 +86,8 @@
     "uglify-js": "^2.8.22"
   },
   "peerDependencies": {
-    "react": "^16.0.0",
-    "react-dom": "^16.0.0"
+    "react": "^16.8.3",
+    "react-dom": "^16.8.3"
   },
   "keywords": [
     "d3",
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -1622,7 +1622,6 @@ acorn-globals@^3.1.0:
 acorn-jsx@^3.0.0:
   version "3.0.1"
   resolved "https://registry.yarnpkg.com/acorn-jsx/-/acorn-jsx-3.0.1.tgz#afdf9488fb1ecefc8348f6fb22f464e32a58b36b"
-  integrity sha1-r9+UiPsezvyDSPb7IvRk4ypYs2s=
```
chrisirhc, line 1625: Are you using an old version of yarn? Weird that it's removing this line. Perhaps try removing the fine and regenerating?

Xiot: Just regenerated the lock file.
I added the volta config in all of the package.json files as well, so that should keep the proper yarn version.

packages/react-vis/package.json
```diff
@@ -86,8 +86,8 @@
     "uglify-js": "^2.8.22"
   },
   "peerDependencies": {
-    "react": "^16.0.0",
-    "react-dom": "^16.0.0"
+    "react": "^16.8.3",
```
a-peltier, line 89: Shouldn't you use 16.8.3 version in devDependencies as well ?

PR #1325, opened 2020-05-24
Upgrade to babel 7

Xiot:
Upgrade all packages to use babel 7.
Moved the plugins into the root babel.config.json file.
Use --root-mode upward to enable lookups into the root config file.


packages/react-vis/package.json
```diff
@@ -86,6 +103,7 @@
     "uglify-js": "^2.8.22"
   },
   "peerDependencies": {
+    "@babel/core": "^7.0.0",
```
chrisirhc, line 106: Hm why do we need this peer dependency? Aren't we compiling the bundle?

We may want to follow the practices on baseweb if we need a reference configuration.

babel.config.json
```diff
@@ -0,0 +1,115 @@
+{
+  "babelrcRoots": [
+    ".",
+    "./packages/*"
+  ],
```
chrisirhc, line 5: With this, do we still need the rootMode option? Reading the Babel docs it seems to indicate that it would search from the roots upwards.

packages/react-vis/package.json
a-peltier: I agree about using baseweb as a reference, the purpose of both projects are similar (components library) and they managed to keep it simple enough

babel.config.json
```diff
@@ -0,0 +1,115 @@
...
+          }
+        ],
+        "@babel/plugin-proposal-function-sent",
+        "@babel/plugin-proposal-export-namespace-from",
+        "@babel/plugin-proposal-numeric-separator",
+        "@babel/plugin-proposal-throw-expressions",
+        "@babel/plugin-proposal-export-default-from",
+        "@babel/plugin-proposal-logical-assignment-operators",
+        "@babel/plugin-proposal-optional-chaining",
+        [
```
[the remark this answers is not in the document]
a-peltier: Agreed, we could add only the features we currently use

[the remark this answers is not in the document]
Xiot: These will probably be mostly redundant when using typescript.
babel deprecated the yearly presents, and these plugins were added using the `babel-upgrade` script.  These are the latest versions of the plugins that we already had.

babel.config.json
Xiot: ya.. we still need the `rootMode` otherwise babel wont look up the tree for the `babel.config.js` file.

Here are the docs for configuring babel in a mono-repo
https://babeljs.io/docs/en/config-files#monorepos

packages/react-vis/package.json
Xiot: oops... I'll remove this.

Xiot:
The tests are currently failing because showcase directly references react-vis and it isn't being "compiled".  This will probably be fixed when we address the imports.

I was hoping to fix this after converting to babel-7, but it looks like I'll have to fix it first.

babel.config.json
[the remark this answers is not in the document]
chrisirhc: If we look at baseweb, looks like there's minimal configuration by using the react preset. Perhaps we can try what's there first?

Xiot:
My approach for tasks like this is to migrate first, and optimize later as it keeps diffs smaller and it more apparent on what the optimaztions are

babel.config.json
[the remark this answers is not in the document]
Xiot: My approach for tasks like this is to migrate first, and optimize later as it keeps diffs smaller and it more apparent on what the optimizations are.

Either way, this won't work until we fix the tests since the tests are referencing 'showcase' and it isn't being "compiled"

Xiot:
PR to fix the imports https://github.com/uber/react-vis/pull/1326

babel.config.json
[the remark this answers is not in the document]
Xiot: trimmed to
```
 "presets": [
        "@babel/preset-env",
        "@babel/preset-react"
      ],
      "plugins": [
        "@babel/plugin-syntax-dynamic-import",
        "@babel/plugin-proposal-class-properties",
        "@babel/plugin-proposal-export-default-from",
        "@babel/plugin-proposal-optional-chaining",
        "@babel/plugin-proposal-nullish-coalescing-operator"
      ]
```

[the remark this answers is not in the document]
Xiot: actually just removed `dynamic-import` as well.

packages/react-vis/package.json
```diff
@@ -80,6 +97,7 @@
     "react-addons-test-utils": ">=15.4.2",
     "react-dom": "^16.0.0",
     "react-test-renderer": "^16.13.1",
+    "react-vis-showcase": "^0.1.0",
```
chrisirhc, line 100: Hm why is this necessary? Won't it become a circular depedency?

packages/showcase/app.css
```diff
@@ -0,0 +1,12 @@
+.candlestick-example{width:100%}.candlestick-example .chart{width:100%}.candlestick-example .dashed-example-line{stroke-dasharray:2, 2}
```
chrisirhc, line 1: Missed out in gitignore?

packages/showcase/webpack.config.js
```diff
@@ -23,10 +25,7 @@ const config = isProd
       },
 
       resolve: {
-        alias: {
-          // 'index':path.join(__dirname,'..', 'src', 'index.js'),
-          // 'theme':path.join(__dirname,'..', 'src', 'theme.js')
-        }
+        modules: ['node_modules', path.join(__dirname,'..', 'react-vis', 'src')],
```
chrisirhc, line 28: urgh, we should avoid custom node_modules resolutions, it's bound to cause headaches late.
I thought yarn workspaces would allow us to avoid this.

packages/react-vis/package.json
Xiot: it is a ciruclar dependency.  But that dependency already existed since the tests directly reference the components in showcase.

We can fix this when we refactor the tests.
https://github.com/uber/react-vis/issues/1328

packages/showcase/webpack.config.js
Xiot: Completely agree.
`yarn workspaces`  will definately help us with this, however there is a bunch of cleanup that has to be done first.

Next step (separate PR) is to clean up the imports in the `react-vis`, then we should be able to remove this module resolution.

PR #1326, opened 2020-05-24
Reference react-vis properly from showcase.

Xiot:
Showcase was using 'index' when importing 'react-vis';
This causes issues when trying to migrate to babel 7.

### DIFF
--- a/packages/react-vis/package.json
+++ b/packages/react-vis/package.json
@@ -106,7 +106,8 @@
             "src"
           ],
           "alias": {
-            "react-vis/*": "src/*"
+            "react-vis/*": "src/*",
+            "react-vis": "src"
           }
         }
       ]
--- a/packages/react-vis/package.json
+++ b/packages/react-vis/package.json
@@ -104,7 +104,10 @@
         {
           "root": [
             "src"
-          ]
+          ],
+          "alias": {
+            "react-vis/*": "src/*"
+          }
         }
       ]
     ],
--- a/packages/react-vis/package.json
+++ b/packages/react-vis/package.json
@@ -104,11 +104,7 @@
         {
           "root": [
             "src"
-          ],
-          "alias": {
-            "react-vis/*": "src/*",
-            "react-vis": "src"
-          }
+          ]
         }
       ]
     ],
--- a/packages/react-vis/tests/components/xy-plot-tests.js
+++ b/packages/react-vis/tests/components/xy-plot-tests.js
@@ -24,6 +24,7 @@ import {mount, shallow} from 'enzyme';
 
 import AbstractSeries from 'plot/series/abstract-series';
 import VerticalBarSeries from 'plot/series/vertical-bar-series';
+import BarSeries from 'plot/series/bar-series';
 import LineSeries from 'plot/series/line-series';
 import XAxis from 'plot/axis/x-axis';
 import XYPlot from 'plot/xy-plot';
@@ -161,8 +162,8 @@ test('testing flexible charts', t => {
 test('Render two stacked bar series with a non-stacked line series chart', t => {
   const $ = mount(<MixedStackedChart />);
 
-  const renderedBarsWrapper = $.find('BarSeries');
-  const renderedLineWrapper = $.find('LineSeries');
+  const renderedBarsWrapper = $.find(BarSeries);
+  const renderedLineWrapper = $.find(LineSeries);
   t.deepEqual(
     renderedBarsWrapper.at(0).prop('data'),
     [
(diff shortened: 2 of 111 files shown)


packages/react-vis/package.json
```diff
@@ -104,7 +104,10 @@
         {
           "root": [
             "src"
-          ]
+          ],
+          "alias": {
+            "react-vis/*": "src/*"
+          }
```
chrisirhc, line 110: This seems incorrect? shouldn't this be in showcase?

packages/react-vis/tests/components/xy-plot-tests.js
```diff
@@ -162,9 +161,8 @@ test('testing flexible charts', t => {
 test('Render two stacked bar series with a non-stacked line series chart', t => {
   const $ = mount(<MixedStackedChart />);
 
-  const renderedBarsWrapper = $.find(BarSeries);
-  const renderedLineWrapper = $.find(LineSeries);
```
chrisirhc, line 166: Pretty sure this was correct before. Why do we need to change this?

packages/showcase/examples/responsive-vis/responsive-bar-chart.js
```diff
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 import React from 'react';
-import {AreaSeries, HorizontalBarSeries, XAxis, XYPlot, YAxis} from 'index';
+import {AreaSeries, HorizontalBarSeries, XAxis, XYPlot, YAxis}from 'react-vis';
```
chrisirhc, line 22: prettier?

packages/react-vis/tests/components/xy-plot-tests.js
Xiot: The `BarSeries` that the showcase is referencing, is different than the `BarSeries` in the test.

packages/react-vis/package.json
Xiot: Theres a whole bunch of indirection happening with the references between these projects.  I'm trying to narrow it down.

Of course, these tests all work on my machine :P

packages/react-vis/tests/components/xy-plot-tests.js
Xiot: fixed the imports and I was able to change this back so that its properly referencing the same instance of the components.

packages/showcase/legends/vertical-discrete-color.js
```diff
@@ -20,7 +20,7 @@
 
 import React from 'react';
 
-import DiscreteColorLegend from 'legends/discrete-color-legend';
+import DiscreteColorLegend from 'react-vis/legends/discrete-color-legend';
```
chrisirhc, line 23: We might want to check whether react-vis exports components in this way, since these showcase files can be used to demonstrate code and be easily-copy-pastable. But that can be a future task. This change seems to be in the right direction.

packages/showcase/package.json
```diff
@@ -31,5 +31,26 @@
...
+    ],
+    "plugins": [
+      [
+        "module-resolver",
+        {
+          "root": [
+            "."
+          ],
+          "alias": {
+            "react-vis/*": "../react-vis/src/*",
```
chrisirhc, line 49: Interesting that we need this statement. Makes sense.
Do we need to remove some duplication from the storybook webpack config?

Xiot: We need this so that the tests will pick up the same instances of the components as the tests.  
Eventually we will need the tests to use their own plots.
The way things are now, a simple change to the showcase can break the tests, which definitely isn't ideal.

packages/showcase/legends/vertical-discrete-color.js
Xiot: This is the way that react-vis exports items.
Although, we should eventually change these imports only use the root `react-vis`. 
This allows us to easily curate our public api.

In general, you _shouldn't_ really import more than one level into a library as it tends to be internal.  `baseui` just recently added lint warnings if you try to go more than 2 levels deep in your imports.

packages/showcase/examples/responsive-vis/responsive-bar-chart.js
Xiot: for some reason prettier doesn't seem to be working.  I can look into this later.

packages/showcase/legends/vertical-discrete-color.js
Xiot: actually... I need to test this..

PR #1329, opened 2020-05-24
Don't lint the dist/ and es/ folders

Xiot:
Move the exlude patterns to .eslint

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,15 +1,15 @@
 {
   "extends": [
     "eslint:recommended",
-    "plugin:prettier/recommended",
-    "plugin:react/recommended"
+    "plugin:react/recommended",
+    "prettier",
+    "prettier/react"
   ],
   "parser": "babel-eslint",
   "plugins": [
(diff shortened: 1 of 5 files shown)


.eslintrc
```diff
@@ -1,4 +1,23 @@
...
+  ],
+  "parser": "babel-eslint",
+  "plugins": [
+    "react",
+    "prettier",
+    "babel",
+    "no-only-tests"
+  ],
+  "ignorePatterns": [
+    "node_modules",
```
a-peltier, line 16: I guess you now removed the .eslintignore file, which is fine if you prefer to only use the .eslintrc

But adding node_modules in ignorePatterns is not needed since it is already automatically ignored by eslint, see https://eslint.org/docs/user-guide/configuring#eslintignore

Xiot: I like explicitly having it here since we have other ignore patterns.
Just a piece of mind, and I don't have to remember if it does it by default or not.

PR #1336, opened 2020-05-27
Migrate tests to jest runner

a-peltier:
- remove tape by jest
- migrate all tests to use jest API thanks to [jest-codemods](https://github.com/skovhus/jest-codemods) and some manual modifications
- remove nfc since jest can generate a coverage report without it

The downside compared to the previous setup is that with jest it's not as easy to add custom message on assertions [liek this one](https://github.com/uber/react-vis/pull/1336/files#diff-753830e44e0f494cf072b29f7afca188L111 ) 

If we want to keep the messages, I think we can try [jest-expect-message](https://github.com/mattphillips/jest-expect-message) but I didn't want to add a new dependency before discussing it

### DIFF
--- a/packages/react-vis/jest.config.js
+++ b/packages/react-vis/jest.config.js
@@ -1,0 +1,10 @@
+/*eslint-env node*/
+const path = require('path');
+
+module.exports = {
+  testMatch: ['**/?(*-)+(test|tests).[tj]s?(x)'],
+  transform: {
+    '^.+\\.js$': path.resolve(__dirname, './jestBabelTransform.js')
+  },
+  setupFilesAfterEnv: ['./jest.setup.js']
+};
--- a/packages/react-vis/jest.setup.js
+++ b/packages/react-vis/jest.setup.js
@@ -1,0 +1,18 @@
+/*eslint-env node*/
+import jsdom from 'jsdom';
+import Enzyme from 'enzyme';
+
+import Adapter from 'enzyme-adapter-react-16';
+Enzyme.configure({adapter: new Adapter()});
+
+global.document = jsdom.jsdom('<body></body>');
+global.window = document.defaultView;
+Object.keys(document.defaultView).forEach(function mapProperties(property) {
+  if (typeof global[property] === 'undefined') {
+    global[property] = document.defaultView[property];
+  }
+});
+
+global.navigator = {
+  userAgent: 'node.js'
+};
--- a/packages/react-vis/package.json
+++ b/packages/react-vis/package.json
@@ -27,9 +27,9 @@
     "build": "yarn run clean && babel --root-mode upward src -d dist --copy-files && BABEL_ENV=es babel --root-mode upward src -d es --copy-files && node-sass src/main.scss dist/style.css --output-style compressed && yarn run build:browser",
     "lint-styles": "stylelint src/styles/*.scss --syntax scss",
     "test:windows": "babel-node --inspect ./tests/index.js",
-    "test": "node --inspect ./node_modules/.bin/babel-node --root-mode upward ./tests/index.js --only='../showcase/**/*.js,src,tests'",
+    "test": "NODE_ENV=development jest",
     "full-test": "npm run lint && npm run cover",
-    "cover": "nyc --reporter=text --reporter=html --reporter=lcov npm test",
+    "cover": "NODE_ENV=development jest --coverage",
     "prettier": "prettier --write $(git ls-files | grep '.js$')"
   },
   "dependencies": {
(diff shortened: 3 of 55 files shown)


packages/react-vis/jest.config.js
```diff
@@ -0,0 +1,10 @@
+/*eslint-env node*/
+const path = require('path');
+
+module.exports = {
+  testMatch: ['**/?(*-)+(test|tests).[tj]s?(x)'],
```
a-peltier, line 5: @chrisirhc @Xiot  For now I put a regex to match our current test files, i.e  `foo-test.js` and `bar-tests.js` 

Many repo use the `.test.` convention, I don't have a strong opinion on this kind of naming, but I think this PR is the opportunity to rename all the test files if you want to

packages/react-vis/package.json
```diff
@@ -27,9 +27,9 @@
     "build": "yarn run clean && babel --root-mode upward src -d dist --copy-files && BABEL_ENV=es babel --root-mode upward src -d es --copy-files && node-sass src/main.scss dist/style.css --output-style compressed && yarn run build:browser",
     "lint-styles": "stylelint src/styles/*.scss --syntax scss",
     "test:windows": "babel-node --inspect ./tests/index.js",
-    "test": "node --inspect ./node_modules/.bin/babel-node --root-mode upward ./tests/index.js --only='../showcase/**/*.js,src,tests'",
+    "test": "NODE_ENV=development jest",
```
a-peltier, line 30: I had to add this because by default jest sets `NODE_END` to `test`, see [this issue](https://github.com/facebook/jest/issues/1654)

We can also add a `test` key to [babel.config.json](https://github.com/uber/react-vis/blob/master/babel.config.json) but since all the plugins are copy-pasted between each key I didn't want to add one more, it would make it more likely to forget a plugin somewhere when we update the file.

packages/react-vis/package.json
```diff
@@ -63,15 +63,16 @@
...
     "babel-plugin-module-resolver": "^4.0.0",
     "babelify": "^10.0.0",
     "browserify": "^14.3.0",
     "canvas-prebuilt": "^1.6.11",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.2",
+    "jest":"^25.5.4",
     "jsdom": "^9.9.1",
     "node-sass": "^4.9.3",
-    "nyc": "^13.0.1",
```
a-peltier, line 74: According to [nyc docs](https://www.npmjs.com/package/nyc): 

> If you use jest or tap, you do not need to install nyc. Those runners already have the IstanbulJS libraries to provide coverage for you.

packages/react-vis/tests/components/xy-plot-tests.js
```diff
@@ -297,21 +237,16 @@ test('Trigger all onParentMouse handlers on Series components', t => {
     .at(0)
     .simulate('touchmove');
 
-  t.end();
+  expect(onParentMouseHandler).toHaveBeenCalledTimes(14);
```
a-peltier, line 240: this assertion is not great but I wanted to migrate to jest by keeping the tests identical if possible and not bring any opinionated change

packages/react-vis/jest.config.js
Xiot: Do we want to keep tests the separate top level `tests` folder, or use the side-by-side `__tests__` folder.   Or just the files completely side by side.

I tend to to prefer the top level `test` folder as I find it keeps things neater.

packages/react-vis/package.json
Xiot: good call... I need to take another pass to clean up that babel config a little bit.

packages/react-vis/package.json
Xiot: 😄

packages/react-vis/jest.config.js
a-peltier: I'm used to have the files side by side at it's easier for me to spot files without tests and to navigate between a file and its test. 

But I also work on projects with a top folder and we have the coverage reports to track missing tests so it's not a big deal for me

Xiot:
We should probably add the `eslint-plugin-jest` plugin as well

packages/react-vis/jest.setup.js
```diff
@@ -0,0 +1,18 @@
+/*eslint-env node*/
```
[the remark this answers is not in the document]
a-peltier: If I need to use the `describe` blocks I am afraid I will have to do it manually. It is quite tedious and can take a while so I think we shouldn't block this PR for that.

As for using `it` instead of `test` imo it can improve readability when the tests are named properly. If we are to keep current names it would give something like: 

`it('Axis: Showcase Example - Even more Custom axes')`

which doesn't makes more sense than:

`test(Axis: Showcase Example - Even more Custom axes')` 

Even if we use the `describe` block, it would still read `it('Even more Custom axes')`

So I guess my take is the same for both questions: we would need to update a lot of stuff manually so it can be done in a second PR in order to merge this one quicker and get rid of the really verbose logs we have with current tests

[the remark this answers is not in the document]
Xiot: ya.. I didn't mean for it to block this review, just a general question.

Issue #1323, opened 2020-05-23
Discuss Minimum supported versions of Node


Xiot:
Currently our CI tests run on Node 8 and 10.
The Node 8 tests were failing because the latest version of `mkdirp` (used by `storybook-addon-jsx` -> `js-beautify`) requires a minum of Node 10.

I've temporarily put in a `resolution` for `js-beatify` that keeps it pinned at a version that uses the old `mkdirp`.

If we remove the dependency on Node 8 we can remove this
https://github.com/uber/react-vis/blob/fa32e967a63924f19ff69643283c2e25e4a06fc1/package.json#L84

In terms of node versions, since the current LTS is `@12` I think we can safely remove `@8` from our CI, and probably add `12`

a-peltier:
Agreed, I removed `@8` in `https://github.com/uber/react-vis/pull/1339`. 

Once it is merged I can add `@12` and remove the `resolution` field

PR #1339, opened 2020-05-28
Add eslint-plugin-jest and rename all test files as *.test.js

### LINKED ISSUE #1337: Add eslint-plugin-jest to eslint config
- [ ] add eslint-plugin-jest
- [ ] enforce recommended rules (such as wrapping all `test` calls in a `describe`

One more thing I would consider is to be in line with jest's default `testMatch` and rename all our test files from `foo-test.js` to `foo.test.js`

a-peltier:
Fixes #1337
See #1323

Reviewing commit by commit should make it easier to read since I renamed many files in the last commit


Xiot:
```
error semver@7.3.2: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"
```
I say we just get rid of Node 8 in our CI

a-peltier:
> ```
> error semver@7.3.2: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"
> ```
> 
> I say we just get rid of Node 8 in our CI

agreed :)

PR #1344, opened 2020-05-31
Fix spread

Xiot:
The spread operation was the last item, so it would actually override what it previously set.


mcnuttandrew:
😬doh, good catch

Issue #1335, opened 2020-05-26
Replace react lifecycle methods with hooks


a-peltier:
See [issue 1320](https://github.com/uber/react-vis/issues/1320#issuecomment-633614883)

`packages/react-vis`: 
- [ ] CanvasWrapper 
- [ ] Animation
- [ ] ArcSeries
- [ ] Treemap
- [ ] make-vis-flexible component factory 

`packages/showcase` (easy to fix, and not as high priority imo since it doesn't cause bugs in the lib): 
- [ ] ClockExample
- [ ] ResponsiveScatterPlot
- [ ] ForceDirectedGraph

`packages/website` (not a priority imo, but good to know): 
- [ ] Hero

a-peltier:
the showcase and website packages can be updated already, but before updating any components from `packages/react-vis` we should make decisions about how we want to measure performance, see #1320 and #1346

PR #1345, opened 2020-05-31
Update problematic showcase examples to use hooks

mcnuttandrew:
This converts the showcase examples listed in #1335 to use hooks. This is slightly in conflict with #1343 which addresses those same examples, although it does so via replacing componentWillRecieve with componentWillUpdate.

### DIFF
--- a/packages/showcase/examples/responsive-vis/responsive-vis-example.js
+++ b/packages/showcase/examples/responsive-vis/responsive-vis-example.js
@@ -18,11 +18,15 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import React from 'react';
+import React, {useState} from 'react';
 
 import {createData, getPPP} from './responsive-vis-utils';
-import ResponsiveScatterplot from './responsive-scatterplot';
-import ResponsiveBarChart from './responsive-bar-chart';
+import ResponsiveScatterplot, {
+  getFeatures as getScatterplotFeatures
+} from './responsive-scatterplot';
+import ResponsiveBarChart, {
+  getFeatures as getBarFeatures
+} from './responsive-bar-chart';
 
 import 'styles/examples.scss';
 import './responsive-vis.scss';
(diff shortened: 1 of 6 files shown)


Xiot:
I would go with using hooks, but the main thing about #1343 was the usage of `componentWillReceiveProps` in the components themselves.  Both PRs are still relevant, but yes, there will be some conflict.  I'm fine with rebasing my PR to incorporate these changes.

packages/showcase/examples/responsive-vis/responsive-vis-example.js
```diff
@@ -18,117 +18,108 @@
...
-export default class ResponsiveVisDemo extends React.Component {
-  state = {
-    dataSize: 1,
-    visSize: 400,
-    data: createData(5, true),
-    chartType: 'barChart'
-  };
+export default function ResponsiveVisDemo() {
+  const [chartType, setChartType] = useState('barChart');
+  const [data, setData] = useState(createData(5, true));
```
Xiot, line 39: `useState(() => createData(5, true))` is a slightly better way to do this, otherwise `createData` will be called on every render, even though the result will only be used the first time.

packages/showcase/examples/responsive-vis/responsive-vis-example.js
```diff
@@ -18,117 +18,108 @@
...
-      </div>
-    );
-  }
+          {`Data Size: ${~~Math.pow(10, dataSize)}`}
+          <input
+            onChange={e => {
+              setDataSize(e.target.value);
+              setData(
+                createData(
+                  ~~Math.pow(10, e.target.value),
```
Xiot, line 95: I've never seen `~~` being used before

packages/showcase/sunbursts/clock-example.js
```diff
@@ -30,53 +30,41 @@ function getSeconds() {
...
-          colorRange={EXTENDED_DISCRETE_COLOR_RANGE}
-        />
-      </XYPlot>
-    );
-  }
+export default function ClockExample() {
+  const [time, setTime] = useState(getSeconds());
+  useEffect(() => {
+    setInterval(() => setTime(getSeconds()), 100);
+  }, []);
```
Xiot, line 37: We should clean up the `setInterval` using the cleanup method of `useEffect`

```
useEffect(() => {
  const handle = setInterval(() ...)
  return () => clearInterval(handle)
}, []);
```

The "downside" to using hooks is that it become very easy to cause re-renders and memory leaks

packages/showcase/examples/responsive-vis/responsive-vis-example.js
mcnuttandrew: This example is derived from http://nrabinowitz.github.io/rdv/ and I was just trying to copy as much of his code as I could. I'm not a huge fan of ascii art operators like that, but felt appropriate in keeping with the style of the copied code. shurggie

packages/showcase/examples/responsive-vis/responsive-vis-example.js
mcnuttandrew: ah nice, i didn't know about that

packages/showcase/examples/responsive-vis/responsive-vis-example.js
Xiot: I'm fine with it... I just have never seen it before.
I had completely forgotten about the `~arr.indexOf()` to check if the value is included.
No need to change.

PR #1348, opened 2020-06-06
Fix imports

Xiot:
Change the import paths so they no longer look like node_module imports.
Project relative can still be done, but using the `~/` prefix.


Xiot:
Checked that `react-vis` builds cleanly, and that `showcase` and `storybook` both spin up.

Xiot:
I'll remove the `codemods` directory before I merge, but I left them in there for now so you can see how I changed everything.

a-peltier:
Nice I will take a look at the codemods!

Issue #1347, opened 2020-06-04
Remove invisible datapoints without derpy animation


KaappoRaivio:
I have a graph displaying real-time data. Over time the number of data points grows impractical to manage, so I have tried to "shave off" the invisible ones. Unfortunately, the animation isn't working right.

Desired outcome:
![desired](https://user-images.githubusercontent.com/33051516/83755898-e9f94280-a676-11ea-9795-d91a0ceda05a.gif)

Actual outcome ("notice how the library thinks that the new data is unrelated to the old one, resulting in the points moving vertically, not horizontally):
![derpy](https://user-images.githubusercontent.com/33051516/83755980-0ac19800-a677-11ea-8897-5188798578a6.gif)

This is the code that removes the old data:
```
const truncate = (data, length) => {
    if (data.length > length) {
        while (data.length >= length) {
            data.shift();
        }
    }

    return data;
}
```

Is there a way to get rid of the derpiness?

Xiot:
@KaappoRaivio Do you have a sharable reproduction of this?
I understand why this would happen, (which is nicely described in this [blog](https://bost.ocks.org/mike/path/)), however in my own tests I am seeing a nice horizontal scrolling.

![smooth-scrolling](https://user-images.githubusercontent.com/1887318/83948086-f1405d80-a7e8-11ea-8d99-4efbf3c6a76a.gif)

Here is the code that I used to generate that gif.

```
import React, {useState, useRef, useEffect} from 'react';
import {XYPlot, LineSeries, YAxis, XAxis, Borders} from 'react-vis';

import {AutoSizer} from 'react-virtualized';

const POINT_COUNT = 100;
const FREQUENCY = 200;

export const ScrollingDataChart = () => {
  const {current: iterator} = useRef(points());

  const [data, setData] = useState(() => {
    let data = [];
    for (let i = 0; i < POINT_COUNT; i++) {
      data.push(iterator.next().value);
    }
    return data;
  });

  useEffect(() => {
    const handle = setInterval(() => {
      setData(data => {
        const {value: point} = iterator.next();
        data.push(point);
        while (data.length > POINT_COUNT) {
          data.shift();
        }
        return [...data];
      });
    }, FREQUENCY);
    return () => clearInterval(handle);
  }, []);

  const xDomain = [data[1].x, data[data.length - 2].x];

  return (
    <AutoSizer>
      {({width, height}) => (
        <XYPlot
          height={height}
          width={width}
[...]

PR #1352, opened 2020-06-07
Allow using Storybook for local development.

Xiot:
Storybook is an easy to use UI to enable testing out features.
This allows storybook to pick up changes from react-vis/src so that it
can be used while developing components.


Xiot:
I realize that its baking in some of the external config, but it only does it for development.

Previously, the storybook, used `react-vis` from the `node_modules`, so if you wanted to make a change and see it in storybook, you had to 
1. make the change
2. stop storybook
3. `yarn build` on `react-vis`
4.  start up the storybook.

With this, the development cycle becomes.
1. start storybook
2. make a change to the component.
3.  repeat step 2.

This greatly enhances the flow.

I really like developing with storybook because of the ability to add `knobs`. You end up with a fairly fluid experience.

When I was digging into https://github.com/uber/react-vis/issues/1347 I actually set up a storybook as it allowed me to easily visualize and test what was going on.  Unfortunately I had to follow the painful original flow of stopping, changing, building, starting

Issue #1328, opened 2020-05-24
Refactor Tests


Xiot:
Currently the tests directly reference the components in 'showcase'.
Tests should be using their own fixtures to test.

Move tests to use `jest`

a-peltier:
I can look into migrating the tests to jest

a-peltier:
Tests with jest display many warning/errors coming from the react components, I am going to work on fixing them to make it easier to read tests and also because many of these messages also display in the browser console

For instance: 
* `Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package)`
* ` Warning: The tag <g> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.`
* `Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.`

Xiot:
The warning for `<g>` is interesting, as it is a valid `svg` tag.
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/g

a-peltier:
It is valid but the warning appears when it is not wrapped by an `svg` tag. I'm currently working on fixing these tests

Xiot:
ah.. that was my next thought.  Thanks for cleaning these up.

PR #1353, opened 2020-06-07
Clean test warnings

a-peltier:
Fix warnings as discussed in #1328 

The only remaining warnings come from react-motion which still uses the deprecated lifecycle methods but the MR to fix these is still pending, see https://github.com/chenglou/react-motion/issues/569


packages/react-vis/tests/components/line-series-canvas.test.js
```diff
@@ -11,8 +11,9 @@ describe('LineSeriesCanvas', () => {
       <XYPlot width={300} height={300}>
         {Array(k)
           .fill(0)
-          .map(() => (
+          .map((_, index) => (
             <LineSeriesCanvas
+              key={index}
```
Xiot, line 16: technically `key` should be a string.

packages/showcase/showcase-components/showcase-button.js
```diff
@@ -30,7 +30,7 @@ function ShowcaseButton(props) {
   );
 }
 
-ShowcaseButton.PropTypes = {
+ShowcaseButton.ropTypes = {
```
Xiot, line 33: `propTypes`

a-peltier: nice catch.. thanks

packages/react-vis/tests/components/line-series-canvas.test.js
a-peltier: done

packages/showcase/showcase-components/showcase-button.js
a-peltier: done

PR #1358, opened 2020-06-17
Migrate Treemap and makeFlexible to functional components

a-peltier:
Relates to  #1335

### DIFF
--- a/packages/react-vis/src/treemap/index.js
+++ b/packages/react-vis/src/treemap/index.js
@@ -91,30 +91,18 @@ function _getScaleFns(props) {
   };
 }
 
-class Treemap extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      scales: _getScaleFns(props),
-      ...getInnerDimensions(props, props.margin)
-    };
-  }
-
-  UNSAFE_componentWillReceiveProps(props) {
-    this.setState({
-      scales: _getScaleFns(props),
-      ...getInnerDimensions(props, props.margin)
-    });
-  }
+function Treemap(props) {
+  const scales = _getScaleFns(props);
+  const innerDimensions = getInnerDimensions(props, props.margin);
 
   /**
    * Create the list of nodes to render.
    * @returns {Array} Array of nodes.
    * @private
    */
-  _getNodesToRender() {
-    const {innerWidth, innerHeight} = this.state;
-    const {data, mode, padding, sortFunction, getSize} = this.props;
+  function _getNodesToRender() {
+    const {innerWidth, innerHeight} = innerDimensions;
+    const {data, mode, padding, sortFunction, getSize} = props;
     if (!data) {
       return [];
     }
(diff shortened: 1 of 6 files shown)


packages/react-vis/src/treemap/index.js
```diff
@@ -91,30 +91,18 @@ function _getScaleFns(props) {
...
 
   /**
    * Create the list of nodes to render.
    * @returns {Array} Array of nodes.
    * @private
    */
-  _getNodesToRender() {
-    const {innerWidth, innerHeight} = this.state;
-    const {data, mode, padding, sortFunction, getSize} = this.props;
+  function _getNodesToRender() {
```
Xiot, line 103: Since this function is only used in one place, this should probably be changed into 
```
const nodes = useMemo(() => { ... })
```

packages/react-vis/src/treemap/index.js
```diff
@@ -91,30 +91,18 @@ function _getScaleFns(props) {
...
-
-  UNSAFE_componentWillReceiveProps(props) {
-    this.setState({
-      scales: _getScaleFns(props),
-      ...getInnerDimensions(props, props.margin)
-    });
-  }
+function Treemap(props) {
+  const scales = _getScaleFns(props);
+  const innerDimensions = getInnerDimensions(props, props.margin);
```
Xiot, line 96: May want to memoize these as well, but it may not matter too much since I believe that they are fairly small methods.

a-peltier: I think memoization makes sense your previous comment as the computation looks a bit complex, but I am more reticent to use memoization on `getInnerDimensions`

My rule of thumb for knowing whether to use memoization or not is "do not use it first, and if you have perf issues, try to fix it with memoization". Otherwise you can in some cases do more harm than good since memoization does have its tradeoffs:

see  [this link](https://blog.logrocket.com/rethinking-hooks-memoization/) and also [that one](https://kentcdodds.com/blog/usememo-and-usecallback)

packages/react-vis/src/treemap/index.js
a-peltier: So I added the `eslint-plugin-react-hooks` because we were missing it, and if I try to wrap the `nodes` computation with `useMemo` I need to have `props` as a dependency: 

> React Hook useMemo has missing dependencies: 'innerDimensions' and 'props'. Either include them or remove the dependency array 

In that case, it would make more sense to me to wrap the component with memo i.e `export default memo(Treemap);` instead of using `useMemo` with props as dependency. wydt?

Xiot: `memo` should work as well since it the funtion depends on the entire `props` object.

packages/react-vis/src/treemap/index.js
Xiot: True, memoization does have its trade offs, but I've also seen the "death by a thousand cuts" way too many times before.  Usually its the culmination of a whole bunch of little performance hits that don't look very big on their own.
Its more of a concern in a bigger app, rather than a library, but once you get to that point its hard to get out of.

I'm not too worried about`getInnerDimensions` since it is simple and is passed as primatives, however `_getScaleFns` creates new functions every execution.  Although this would be better fixed at a higher level (as per my other PR).

PR #1362, opened 2020-07-13
Add onTouchStart event listener to support touch actions

adammartiska:
If we want zoom to work with touch devides we need to also register onTouchStart event that sets brushing to true in order to not return immediately in onBrushEnd method (where is both brushing and dragging false)


adammartiska:
Code is not working properly

Issue #1248, opened 2019-09-16
Is this library maintained?


andylacko:
don't see much progress in solving issues or replying to issues, what is happening to this library please?

m7mdBalboul:
same question here

nullberri:
Uber did just lay off a ton of engineers.

andylacko:
@avindra 😄 that is what we have been using before... anyone idea about really good react based library for graphs? but not just for simple lines, but combined charts and whole functionality

uber team? the update of react to 17  is coming and we have still incompatible library 🤷‍♂️

PS: everytime I take uber I am thinking about this, maybe I'll start complaining to driver about react-vis

nullberri:
@andylacko I'd still stick with react-vis even post 17.x, theres a number of open pull requests which will likely keep the library running. I have tried all the other big name react graphing libraries and they all have their fatal flaws. This one is still the best of the bunch even if its abandoned. 

Just fork and cherry pick the pull requests.

pixelass:
No answer is an answer too.

Looks like building something custom is the only option if you need a maintained up-to-date lib that offers SSR. (which is really sad considering the number of libraries out there)

😭

chrisirhc:
Hi folks, we're deprecating this library as we don't have an active maintainer.
If you'd like to take take ownership, please feel free to fork or discuss over here.

mcnuttandrew:
Closing as this is now more covered in #1303

jufemaiz:
Deprecated notice was reverted. https://github.com/uber/react-vis/commits/master/DEPRECATED.md

Perhaps it needs to be added back in.

mcnuttandrew:
Yea i think that's a fine idea. I can plan on doing so sometime in the future, or I'd happily review a PR to that effect

jufemaiz:
https://github.com/uber/react-vis/pull/1444 🎉

PR #1444, opened 2022-03-18
Reapply reverted commit regarding deprecation

jufemaiz:
As per https://github.com/uber/react-vis/issues/1248#issuecomment-1071062630


mcnuttandrew:
Thanks for doing that! Unfortunately it looks like the doc structure got shift around some since that original commit, so this PR creates a ton of files. I've filed another PR (#1445) which repeats the original deprecating. Sorry about that!

PR #1469, opened 2023-01-07
remove redundant files in dist/ after build

kenns29:


spencercharestpaypal:
@kenns29 it appears this PR has broken imports in `1.11.11`. It looks to me like either the build file `dist/dist.min.js` should be renamed to `dist/index.js` or that `package.json` `main` field should be updated to `dist/dist.min.js`. `main` is currently pointing to `dist/index.js` which no longer exists. Is there anyway you could resolve this for `1.11.12`? Let me know if you'd like me to file an issue instead.

Issue #1356, opened 2020-06-15
Support React 17


Xiot:
React 17 adds support for async rendering which will need some testing and tweaking to properly support.

It explicitly removes support for various lifecycle methods on class components:
```
componentWillReceiveProps
componentWillMount
componentWillUpdate
```
and moves to use the new ones
```
getDerivedStateFromProps
getSnapshotBeforeUpdate
```

These new methods, and the deprecation warnings were added back in 16.3.

Do do have a couple places that I believe are still using the old methods, although there was a pass to remove them.

wojtekmaj:
Looks like it does not remove these methods after all! Postponed to React 18+.

daveisfera:
Ya, so `|| ^17.0.0` needs to be added to the dependency ( see https://github.com/uber/react-vis/blob/cce5980cd29416bdae592b34f2e52a088176493c/packages/react-vis/package.json#L89 )

karamvirsingh98:
Hi, I can't seem to install react-vis in a create-react-app project with React 17. Here's the npm error log, hope that helps! 

npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: react-vis@0.1.0
npm ERR! Found: react@17.0.2
npm ERR! node_modules/react
npm ERR!   react@"^17.0.2" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"15.3.0 - 16.x" from react-vis@1.11.7
npm ERR! node_modules/react-vis
npm ERR!   react-vis@"*" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.

bmaupin:
Now that npm has overrides, I wonder if that could be used as a workaround in this case since this library is unmaintained?

https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides

e.g.

```json
  "overrides": {
    "react": "^17.0.0"
```

wojtekmaj:
Yes it could. I'm happily using react-vis with React 17 in production on a large scale commercial application with zero complaints. Hell, preparing to upgrade to React 18 as soon as it gets released, with RC being evaluated right now, and everything still works just fine.

bmaupin:
[...]

PR #1473, opened 2023-02-04
Upgrade to React 17

Sergio-Mira:
Attempt to upgrade to React 17 https://github.com/uber/react-vis/issues/1356

### DIFF
--- a/packages/react-vis/package.json
+++ b/packages/react-vis/package.json
@@ -76,9 +76,9 @@
     "jsdom": "^9.9.1",
     "node-sass": "^4.9.3",
     "prettier": "^1.14.2",
-    "react": "^16.0.0",
+    "react": "^17.0.2",
     "react-addons-test-utils": ">=15.4.2",
-    "react-dom": "^16.0.0",
+    "react-dom": "^17.0.2",
     "react-test-renderer": "^16.13.1",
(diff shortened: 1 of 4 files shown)


jkytomak:
This looks reallly straigthforward PR.
@kenns29 any chance mergning this?
@Sergio-Mira , did I understand right that this makes react-vis even work with React 18? Referring to this comment of yours:
https://github.com/jaegertracing/jaeger-ui/pull/1173#discussion_r1096591635

Sergio-Mira:
I believe that the current version works with React 18 already, as those warnings are still warnings in those React 18.X versions. You can see that `jaeger-ui` is using React 18 here and using the current version of `react-vis`: https://github.com/jaegertracing/jaeger-ui/blob/main/yarn.lock#L9025

jkytomak:
> I believe that the current version works with React 18 already, as those warnings are still warnings in those React 18.X versions. You can see that `jaeger-ui` is using React 18 here and using the current version of `react-vis`: https://github.com/jaegertracing/jaeger-ui/blob/main/yarn.lock#L9025

Great, thanks! We got earlier plenty of type errors when trying to upgrade. Looks like manually resolving right types for 17 solved it also for 18. Had the problem first when trying to upgrade newest 17 types.

